PR #38, opened 2026-04-10
feat: trie data structure in Web Worker for autocomplete

### LINKED ISSUE #31: feat: trie data structure in Web Worker
## Parent PRD

#29

## What to build

Implement the trie (prefix tree) data structure inside a Web Worker (`frontend/src/workers/trie.worker.ts`). All trie construction and querying runs off the main thread to avoid blocking the UI.

- Implement trie node structure and insert/search methods
- On receiving a `{ type: 'build', words }` message: insert all words into the trie, then send `{ type: 'ready' }` back to confirm completion
- On receiving a `{ type: 'query', prefix }` message: traverse the trie and return up to 5 matching words sorted alphabetically
- Worker is a singleton — created once at module level, never terminated

## Acceptance criteria

- [ ] `frontend/src/workers/trie.worker.ts` exists
- [ ] Trie correctly inserts words and returns all words matching a given prefix
- [ ] Sends `{ type: 'ready' }` message after trie is fully built
- [ ] Returns up to 5 alphabetically sorted suggestions for a prefix query
- [ ] Worker is instantiated as a singleton (module-level, not per component)
- [ ] Uses Vite worker syntax: `new Worker(new URL('./trie.worker.ts', import.meta.url))`

## Blocked by

- Blocked by #30 (needs the endpoint to fetch words from)

## User stories addressed

- User story 1 (core engine for autocomplete suggestions)

tomerp20:
## Summary
- Implements a prefix trie data structure that runs inside a Web Worker, keeping 100K+ word insertions off the main thread
- Responds to `{ type: 'build', words }` by building the trie and posting `{ type: 'ready' }` when complete
- Responds to `{ type: 'query', prefix }` by returning up to 5 alphabetically sorted matching words as `{ type: 'results', suggestions }`

## Changes
- `frontend/src/workers/trie.worker.ts` — trie worker implementation (TrieNode structure, insert, DFS prefix search, message handler)
- `frontend/src/workers/trieWorker.ts` — singleton Worker export using Vite's `new Worker(new URL(...))` syntax

## Test plan
- [ ] Import `trieWorker` from `trieWorker.ts` in a component and send a `build` message with a list of Hebrew words; verify `ready` message is received
- [ ] Send a `query` message with a known prefix; verify up to 5 alphabetically sorted matches are returned
- [ ] Send a `query` with a prefix that has no matches; verify an empty `suggestions` array is returned
- [ ] Verify `trieWorker` is the same instance across multiple imports (singleton)

Closes #31

### DIFF
--- a/frontend/src/workers/trie.worker.ts
+++ b/frontend/src/workers/trie.worker.ts
@@ -32,12 +32,10 @@ function getWordsWithPrefix(prefix: string, limit: number): string[] {
   const results: string[] = [];
 
   function dfs(current: TrieNode, path: string): void {
-    if (results.length >= limit) return;
     if (current.isEnd) {
       results.push(path);
     }
     for (const [char, child] of current.children) {
-      if (results.length >= limit) return;
       dfs(child, path + char);
     }
   }
@@ -53,6 +51,7 @@ self.onmessage = (event: MessageEvent) => {
     | { type: 'query'; prefix: string };
 
   if (data.type === 'build') {
+    if (!Array.isArray(data.words)) return;
     for (const word of data.words) {
       insert(word);
     }
--- a/frontend/src/workers/trie.worker.ts
+++ b/frontend/src/workers/trie.worker.ts
@@ -1,0 +1,66 @@
+interface TrieNode {
+  children: Map<string, TrieNode>;
+  isEnd: boolean;
+}
+
+function createNode(): TrieNode {
+  return { children: new Map(), isEnd: false };
+}
+
+const root: TrieNode = createNode();
+
+function insert(word: string): void {
+  let node = root;
+  for (const char of word) {
+    if (!node.children.has(char)) {
+      node.children.set(char, createNode());
+    }
+    node = node.children.get(char)!;
+  }
+  node.isEnd = true;
+}
+
+function getWordsWithPrefix(prefix: string, limit: number): string[] {
+  let node = root;
+  for (const char of prefix) {
+    if (!node.children.has(char)) {
+      return [];
+    }
+    node = node.children.get(char)!;
+  }
+
+  const results: string[] = [];
+
+  function dfs(current: TrieNode, path: string): void {
+    if (results.length >= limit) return;
+    if (current.isEnd) {
+      results.push(path);
+    }
+    for (const [char, child] of current.children) {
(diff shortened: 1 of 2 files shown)


tomerp20 (commented):
## Code Review — feature/autocomplete-trie-worker

### Issues Found

**1. Semantic bug: DFS returns first `N` traversal-order words, not first `N` alphabetically**

In `trie.worker.ts`, `getWordsWithPrefix` cuts off collection at `limit` during DFS (traversal/insertion order), then sorts and slices. Because the slice is applied after sorting a list that is already at most `limit` elements, the final result is the first 5 words found during DFS — sorted — rather than the alphabetically first 5 words among all matches.

Example: if the trie contains `["מנהל", "מנגינה", "מנוחה", "מנוף", "מנורה", "מנות"]` (all sharing prefix `מנ`), and they were inserted in that order, DFS returns those 5 and never considers `מנות` as a 6th candidate — even though it may sort before some of the 5 already collected.

Fix: remove the early-exit limit from DFS and collect all matching words, then sort and slice. For 100K words, worst case (empty prefix) the full word list is traversed — but queries will always have at least 1–2 characters, keeping the result set manageable.

**2. No input validation in `build` message handler**

The `build` handler in `trie.worker.ts` casts `event.data` with `as` but never verifies that `data.words` is an array of strings before passing each element to `insert`. If a caller posts `{ type: 'build', words: null }` or passes objects instead of strings, `insert` will iterate over `null` (TypeError) or call `for...of` on objects (TypeError/silent corruption).

Fix: add a guard — `if (!Array.isArray(data.words)) return;` — before the build loop.

**3. Redundant `results.slice(0, limit)` return**

Because DFS already caps `results` at exactly `limit` entries, `results.slice(0, limit)` always returns the full array unchanged. This is harmless but misleading — it implies the array may contain more than `limit` elements when it cannot. Once the DFS early-exit is removed (to fix issue #1), the slice becomes meaningful and necessary.

**4. Minor: over-documented singleton file**

`trieWorker.ts` has 3 lines of block comments for a 2-line module. The codebase prefers minimal comments; the variable name and file name already communicate the intent.

---

**Summary:** Issues #1 (semantic correctness of alphabetical ordering) and #2 (input validation) should be addressed before merge. Issues #3 and #4 are minor cleanup.

PR #39, opened 2026-04-10
feat: add useAutocomplete hook with lazy init and debounced trie queries

tomerp20:
## Summary
- Adds `useAutocomplete` hook at `frontend/src/hooks/useAutocomplete.ts`
- Lazily fetches `GET /api/autocomplete/words` and builds the trie worker on first focus (via `triggerInit()`)
- Debounces prefix queries to the trie worker by 150ms using `useRef` + `setTimeout` — no external library

## Changes
- New file: `frontend/src/hooks/useAutocomplete.ts`
- Uses the singleton `trieWorker` from `frontend/src/workers/trieWorker.ts`
- `isLoading` is `true` from `triggerInit()` until the worker posts `{ type: 'ready' }`
- On fetch failure, resets `initializedRef` so the next focus attempt can retry
- Returns `{ suggestions, isLoading, triggerInit, query }` — generic, not coupled to any specific component

## Test plan
- [ ] Call `triggerInit()` on first input focus — verify fetch to `/api/autocomplete/words` fires only once
- [ ] `isLoading` becomes `true` on first `triggerInit()` and `false` after worker signals ready
- [ ] Type rapidly into an autocomplete input — verify the trie worker only receives a query after 150ms of inactivity
- [ ] Simulate a network failure for `/api/autocomplete/words` — verify `isLoading` returns to `false` and a retry works on next focus
- [ ] Confirm `triggerInit()` is a no-op on all subsequent calls after the first

### DIFF
--- a/frontend/src/hooks/useAutocomplete.ts
+++ b/frontend/src/hooks/useAutocomplete.ts
@@ -1,0 +1,69 @@
+import { useState, useRef, useCallback, useEffect } from 'react';
+import trieWorker from '@/workers/trieWorker';
+
+interface UseAutocompleteResult {
+  suggestions: string[];
+  isLoading: boolean;
+  triggerInit: () => void;
+  query: (prefix: string) => void;
+}
+
+export function useAutocomplete(): UseAutocompleteResult {
+  const [suggestions, setSuggestions] = useState<string[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+
+  const initializedRef = useRef(false);
+  const debounceTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+
+  useEffect(() => {
+    const handleMessage = (event: MessageEvent) => {
+      const data = event.data as
+        | { type: 'ready' }
+        | { type: 'results'; suggestions: string[] };
+
+      if (data.type === 'ready') {
+        setIsLoading(false);
+      } else if (data.type === 'results') {
+        setSuggestions(data.suggestions);
+      }
+    };
+
+    trieWorker.addEventListener('message', handleMessage);
+    return () => {
+      trieWorker.removeEventListener('message', handleMessage);
+    };
+  }, []);
+
+  const triggerInit = useCallback(() => {
+    if (initializedRef.current) return;
+    initializedRef.current = true;
+    setIsLoading(true);
+
+    fetch('/api/autocomplete/words')
+      .then((res) => {
+        if (!res.ok) {
+          throw new Error(`Failed to fetch words: ${res.status}`);
+        }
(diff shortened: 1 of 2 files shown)


tomerp20 (commented):
## Code Review — useAutocomplete Hook

Overall the hook is clean and follows the spec. Found **4 issues** ranging from a correctness bug to best-practice concerns.

---

### Issue 1 (Bug) — Debounce timer not cancelled on unmount

`debounceTimerRef.current` may hold a live timer when the component unmounts. The `useEffect` cleanup only removes the message listener; the pending `setTimeout` is never cleared. This means `trieWorker.postMessage({ type: 'query', ... })` can fire after unmount, and `setIsLoading(false)` in the catch block of `triggerInit` can be called on an unmounted component (a React warning in dev mode).

**File:** `frontend/src/hooks/useAutocomplete.ts` — the `useEffect` cleanup at line 32–34 must also call `clearTimeout(debounceTimerRef.current)`.

---

### Issue 2 (Bug) — Singleton worker sends results to all mounted instances

`trieWorker` is a module-level singleton. Every mounted `useAutocomplete` instance adds its own `message` listener. When the worker posts `{ type: 'results', suggestions }`, **every** instance receives it and updates its own `suggestions` state — even the ones that didn't issue the query. If two components use the hook simultaneously they'll both update with whichever query was last fired.

**Fix:** Each hook instance should only process results that it triggered. One lightweight approach is to give each query a monotonically-increasing sequence number (via a `useRef`) and have the result handler discard results from stale requests.

---

### Issue 3 (Best practice) — `fetch` bypasses the shared axios instance / base URL

All other API calls go through the axios `api` instance in `services/api.ts`, which respects `VITE_API_URL` and attaches the auth bearer token. This hook calls the native `fetch` with a hardcoded `/api/autocomplete/words`. While the backend marks this endpoint as unauthenticated today, bypassing the shared client means the hook will break if `VITE_API_URL` is ever set to an absolute URL (e.g. `https://api.example.com`) in a non-localhost environment — `/api/...` would resolve against the frontend origin instead.

**Fix:** Import and use the axios `api` instance (or at least compose the URL from `import.meta.env.VITE_API_URL`).

---

### Issue 4 (Code quality) — No guard against querying before trie is ready

`query()` posts to the worker even while `isLoading` is true (i.e. the trie hasn't been built yet). The worker will receive a `query` message before it processes `build`, and will respond with an empty array. Callers won't notice, but it produces a spurious empty-suggestions flash.

**Fix:** Add an `isReadyRef` that is set to `true` inside the `handleMessage` for `'ready'`, and return early in `query()` if `!isReadyRef.current`.

---

*Reviewed by Claude Code on 2026-04-09*

PR #40, opened 2026-04-10
feat: AutocompleteDropdown generic UI component

### LINKED ISSUE #33: feat: AutocompleteDropdown generic UI component
## Parent PRD

#29

## What to build

Implement the generic `AutocompleteDropdown` component (`frontend/src/components/AutocompleteDropdown.tsx`). This component is not tied to SearchBar — it can be dropped into any input field in the future.

- Accepts `suggestions: string[]`, `isLoading: boolean`, and selection/dismiss callbacks as props
- While `isLoading` is `true`: show a small loading spinner (user knows system is initializing, not broken)
- Once loaded: show up to 5 suggestions
- Each suggestion item has `dir="auto"` for correct Hebrew/English RTL rendering
- Dismisses on: suggestion selected, Escape key, click outside, input cleared
- On selection: calls a callback that replaces the input value and triggers search immediately

## Acceptance criteria

- [ ] `frontend/src/components/AutocompleteDropdown.tsx` exists
- [ ] Shows loading spinner when `isLoading` is `true`
- [ ] Shows up to 5 suggestion items when loaded
- [ ] Each suggestion item has `dir="auto"`
- [ ] Dismisses on selection, Escape, outside click, and input clear
- [ ] Selection callback replaces input value and triggers search
- [ ] Component has no dependency on SearchBar — generic and reusable

## Blocked by

- Blocked by #32 (depends on the hook's interface to define props correctly)

## User stories addressed

- User story 2, 3, 4

tomerp20:
## Summary
- Adds a generic `AutocompleteDropdown` component at `frontend/src/components/AutocompleteDropdown.tsx`
- Accepts `suggestions`, `isLoading`, `onSelect`, and `onDismiss` props — no coupling to SearchBar or useAutocomplete

## Changes
- Shows a loading spinner (Lucide `Loader2`) when `isLoading` is `true`
- Shows up to 5 suggestion items once loaded
- Each suggestion item has `dir="auto"` for correct Hebrew (RTL) / English (LTR) rendering
- Dismisses on: suggestion selected, Escape key, click outside (150 ms blur delay so click registers first)
- `onMouseDown` uses `preventDefault` to prevent the blur from stealing the click event
- Styled consistently with the existing app — `rounded-lg`, `border-gray-200`, `shadow-lg`, `primary-*` hover colours

## Test plan
- [ ] Render `<AutocompleteDropdown suggestions={["React", "Node"]} isLoading={false} onSelect={...} onDismiss={...} />` — two items appear
- [ ] Set `isLoading={true}` — spinner appears instead of items
- [ ] Click a suggestion — `onSelect` is called with the word; dropdown closes
- [ ] Press Escape — `onDismiss` is called
- [ ] Click outside the dropdown — `onDismiss` is called after ~150 ms
- [ ] Pass a Hebrew suggestion — it renders RTL correctly due to `dir="auto"`
- [ ] Pass more than 5 suggestions — only 5 are displayed

Closes #33

### DIFF
--- a/frontend/src/components/AutocompleteDropdown.tsx
+++ b/frontend/src/components/AutocompleteDropdown.tsx
@@ -14,7 +14,7 @@ export default function AutocompleteDropdown({
   onSelect,
   onDismiss,
 }: AutocompleteDropdownProps) {
-  const containerRef = useRef<HTMLUListElement>(null);
+  const containerRef = useRef<HTMLDivElement>(null);
 
   // Dismiss on Escape key
   useEffect(() => {
@@ -27,6 +27,17 @@ export default function AutocompleteDropdown({
     return () => document.removeEventListener('keydown', handleKeyDown);
   }, [onDismiss]);
 
+  // Dismiss on outside click via document mousedown listener
+  useEffect(() => {
+    const handleMouseDown = (e: MouseEvent) => {
+      if (containerRef.current && !containerRef.current.contains(e.target as Node)) {
+        onDismiss();
+      }
+    };
+    document.addEventListener('mousedown', handleMouseDown);
+    return () => document.removeEventListener('mousedown', handleMouseDown);
+  }, [onDismiss]);
+
   const visibleSuggestions = suggestions.slice(0, 5);
 
   if (!isLoading && visibleSuggestions.length === 0) {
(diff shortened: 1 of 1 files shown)


tomerp20 (commented):
## Code Review — AutocompleteDropdown

### Issues Found

**1. `onBlur` on `<ul>` never fires — outside-click dismissal is broken**

`<ul>` elements are not focusable by default, so the `onBlur` handler attached to the list will never fire. The `containerRef` and the blur-based dismiss logic are effectively dead code. Outside-click detection should use a `document` `mousedown` listener instead, which is the standard React pattern for this use case.

`frontend/src/components/AutocompleteDropdown.tsx`, lines 40–45

**2. `focus:outline-none` removes keyboard accessibility without an adequate replacement**

Removing the default focus outline is an accessibility issue. The `focus:bg-primary-50` background change is too subtle to serve as a visible focus indicator. The rest of the app (e.g. SearchBar) uses `focus:ring-2 focus:ring-primary-500/20` — this component should follow the same pattern.

`frontend/src/components/AutocompleteDropdown.tsx`, line 52

**3. `containerRef` is unused in practice**

Because `<ul>` never receives blur events, `containerRef` is never meaningfully used. Once the blur-based approach is replaced with a `document mousedown` listener, `containerRef` becomes the anchor for the `contains` check instead — so it should be kept but used correctly.

`frontend/src/components/AutocompleteDropdown.tsx`, line 17

**4. Minor style: `onClick` wraps a single call in an unnecessary arrow function**

`onClick={() => { onSelect(word); }}` should be `onClick={() => onSelect(word)}` — consistent with the codebase style.

`frontend/src/components/AutocompleteDropdown.tsx`, line 60

---

### Summary

The main functional bug is item 1 — the outside-click dismiss will silently never work. Items 2–4 are quality/accessibility issues. Will fix all and push a follow-up commit.

PR #41, opened 2026-04-10
feat: wire autocomplete into SearchBar

tomerp20:
## Summary
- Integrates `useAutocomplete` hook and `AutocompleteDropdown` component into `SearchBar`
- Lazy trie initialization triggers on input focus; dropdown appears when suggestions/loading state is active
- Selecting a suggestion immediately fires the search without waiting for the 300ms debounce

## Changes
- Import `useAutocomplete` and `AutocompleteDropdown` in `SearchBar.tsx`
- Track `inputFocused` state to control dropdown visibility
- Add `onFocus` handler: sets focused state and calls `triggerInit()`
- Add `onBlur` handler: clears focused state (hides dropdown)
- `useEffect` calls `query(searchText)` on every text change for live suggestions
- `handleSelect`: sets input value, clears focused state, cancels debounce, fires `onFiltersChange` immediately
- `handleDismiss`: clears focused state (hides dropdown)
- All existing behavior preserved: debounced search, stage/priority/workMode filters, clear button

## Test plan
- [ ] Focus the search input — trie initialisation starts (loading spinner appears briefly)
- [ ] Type Hebrew or Latin characters — matching suggestions appear in the dropdown
- [ ] Click a suggestion — input is filled and search fires immediately (cards filter without pressing Enter)
- [ ] Press Escape while dropdown is open — dropdown closes
- [ ] Click outside the search area — dropdown closes
- [ ] Use stage / priority / work-mode filters — still works as before
- [ ] Clear button still resets search text
- [ ] Typing normally without selecting still debounces at 300ms

### DIFF
--- a/frontend/src/components/Search/SearchBar.tsx
+++ b/frontend/src/components/Search/SearchBar.tsx
@@ -1,6 +1,8 @@
-import { useState, useEffect, useRef } from 'react';
+import { useState, useEffect, useRef, useCallback } from 'react';
 import type { CardFilters, Stage } from '@/types';
 import { Search, X, SlidersHorizontal } from 'lucide-react';
+import { useAutocomplete } from '@/hooks/useAutocomplete';
+import AutocompleteDropdown from '@/components/AutocompleteDropdown';
 
 interface SearchBarProps {
   filters: CardFilters;
@@ -11,7 +13,9 @@ interface SearchBarProps {
 export default function SearchBar({ filters, onFiltersChange, stages }: SearchBarProps) {
   const [searchText, setSearchText] = useState(filters.search || '');
   const [showFilters, setShowFilters] = useState(false);
+  const [inputFocused, setInputFocused] = useState(false);
   const debounceRef = useRef<ReturnType<typeof setTimeout>>(undefined);
+  const { suggestions, isLoading, triggerInit, query } = useAutocomplete();
 
   useEffect(() => {
     debounceRef.current = setTimeout(() => {
@@ -22,6 +26,27 @@ export default function SearchBar({ filters, onFiltersChange, stages }: SearchBa
     }, 300);
     return () => clearTimeout(debounceRef.current);
   }, [searchText]);
+
+  useEffect(() => {
+    query(searchText);
+  }, [searchText, query]);
+
+  const handleSelect = useCallback(
+    (word: string) => {
+      setSearchText(word);
+      setInputFocused(false);
+      // Cancel the pending debounce and fire immediately
+      clearTimeout(debounceRef.current);
+      onFiltersChange({ ...filters, search: word || undefined });
+    },
+    [filters, onFiltersChange],
+  );
+
+  const handleDismiss = useCallback(() => {
+    setInputFocused(false);
+  }, []);
+
+  const showDropdown = inputFocused && (isLoading || suggestions.length > 0);
 
   const hasActiveFilters = filters.stageId || filters.priority || filters.workMode;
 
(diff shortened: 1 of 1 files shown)


tomerp20 (commented):
## Code Review — feature/autocomplete-searchbar

Overall the integration is clean and the happy path works correctly. Three issues found:

---

### Issue 1 (Bug): `clearFilters` does not close the dropdown

**File:** `frontend/src/components/Search/SearchBar.tsx`

When the user clicks **"Clear all"**, `clearFilters` calls `setSearchText('')` and resets filters, but it never sets `inputFocused(false)`. If the input is currently focused and the trie is already initialised, clicking "Clear all" leaves the dropdown open with stale suggestions (the previous query result remains in state until a new query resolves). The dropdown should be dismissed alongside the text reset.

**Fix:** call `setInputFocused(false)` inside `clearFilters`.

---

### Issue 2 (Minor): `query` useEffect fires on mount with empty string

**File:** `frontend/src/components/Search/SearchBar.tsx`

```ts
useEffect(() => {
  query(searchText);
}, [searchText, query]);
```

This fires immediately on mount with `searchText = ''`. The hook guards with `isReadyRef.current`, so it's a safe no-op, but it still schedules a 150ms debounce timer that is immediately cancelled and reset whenever the trie isn't ready. The intent is to query on *changes* to `searchText`, not on mount. Using a `useRef` mount-guard or a conditional `if (searchText)` avoids the spurious timer.

**Fix:** add `if (!searchText) return;` at the top of the effect (empty prefix produces no meaningful suggestions and the trie init path is already handled by `triggerInit`).

---

### Issue 3 (Minor): `onBlur` dismisses dropdown before keyboard-initiated suggestion focus

**File:** `frontend/src/components/Search/SearchBar.tsx`

```tsx
onBlur={() => setInputFocused(false)}
```

If a user presses Tab from the search input to reach a suggestion button, the input `blur` fires first, setting `inputFocused = false` and un-mounting the dropdown before the button ever receives focus. The current spec only mentions Escape/click-outside dismiss, and `AutocompleteDropdown` has no keyboard arrow-key navigation, but this will silently break any future keyboard accessibility work.

A safe fix is to replace the `onBlur` approach with a `relatedTarget` check: only dismiss if focus moves to an element outside the dropdown container. Alternatively, add a `onFocusCapture` wrapper. For now, adding a comment noting this limitation is the minimum; for full a11y correctness the check should use `relatedTarget`.

**Recommendation for now:** at minimum, document the limitation. If Tab-to-suggestion is a real use case, use `relatedTarget` to guard the dismiss.

---

**Summary:** Issue 1 is a real UX bug that should be fixed. Issues 2 and 3 are minor. I will fix Issues 1 and 2 and document Issue 3.

tomerp20 (commented):
## Code Review — feature/autocomplete-searchbar

Overall the integration is clean and the happy path works correctly. Three issues found:

### Issue 1 (Bug): clearFilters does not close the dropdown

frontend/src/components/Search/SearchBar.tsx

When the user clicks "Clear all", clearFilters calls setSearchText('') and resets filters, but it never sets inputFocused(false). If the input is currently focused and the trie is already initialised, clicking "Clear all" leaves the dropdown open with stale suggestions (the previous query result remains in state until a new query resolves).

Fix: call setInputFocused(false) inside clearFilters.

### Issue 2 (Minor): query useEffect fires on mount with empty string

frontend/src/components/Search/SearchBar.tsx

The effect fires immediately on mount with searchText = ''. The hook guards with isReadyRef.current so it is a safe no-op, but it schedules a spurious 150ms debounce timer. Empty prefix produces no useful suggestions.

Fix: add an early return when searchText is empty.

### Issue 3 (Minor/A11y): onBlur dismisses dropdown before keyboard-initiated suggestion focus

If a user presses Tab from the search input to a suggestion button, the input blur fires first and un-mounts the dropdown before the button receives focus. This silently breaks any future keyboard navigation. A relatedTarget check would fix this properly; for now a comment documenting the limitation is the minimum.

tomerp20 (commented):
Code Review: feature/autocomplete-searchbar

Issue 1 (Bug): clearFilters does not close the dropdown. When the user clicks Clear all, setInputFocused(false) is never called. If the input was focused and the trie is ready, the dropdown stays open with stale suggestions after clearing.

Issue 2 (Minor): The query useEffect fires on mount with an empty searchText string, scheduling a spurious debounce timer. Adding an early return when searchText is empty avoids this.

Issue 3 (Minor/A11y): onBlur unconditionally hides the dropdown. If a user presses Tab from the input to a suggestion button, blur fires first and the dropdown unmounts before the button gets focus, breaking keyboard accessibility. A relatedTarget guard would fix this; at minimum a comment should document the limitation.

tomerp20 (commented):
Code review for feature/autocomplete-searchbar: Issue 1 Bug clearFilters does not dismiss dropdown. Issue 2 Minor query effect fires on mount with empty string. Issue 3 Minor onBlur breaks Tab-key navigation.

PR #54, opened 2026-04-15
fix: preserve Tab-key navigation in autocomplete dropdown

### LINKED ISSUE #42: bug: Tab-key navigation closes autocomplete dropdown prematurely
## Description

When the autocomplete dropdown is open and the user presses Tab to navigate to a suggestion, the `onBlur` event on the search input fires unconditionally and closes the dropdown before the suggestion can be selected.

This works correctly with mouse clicks but breaks keyboard accessibility.

## Root cause

In `SearchBar.tsx`, the `onBlur` handler sets `inputFocused(false)` without checking `relatedTarget`. When Tab is pressed, focus moves to the suggestion buttons inside `AutocompleteDropdown` — but `onBlur` fires first and hides the dropdown, making the suggestion buttons unreachable via keyboard.

## Fix direction

Guard the `onBlur` handler with `relatedTarget`:

```ts
onBlur={(e) => {
  if (!e.currentTarget.contains(e.relatedTarget as Node)) {
    setInputFocused(false)
  }
}}
```

This keeps the dropdown open when focus moves to a child element (a suggestion button), and only closes it when focus truly leaves the component.

## Affected file

`frontend/src/components/Search/SearchBar.tsx`

## Parent PRD

#29

tomerp20:
## Summary
- Fixes Tab-key navigation in the autocomplete dropdown being closed prematurely (issue #42)
- Moves the `onBlur` handler from the `input` to the wrapper `div` and guards it with `e.currentTarget.contains(e.relatedTarget)` so the dropdown stays open when focus moves to a suggestion button

## Changes
- `SearchBar.tsx`: Removed `onBlur={() => setInputFocused(false)}` from `<input>` and replaced with a bubbling `onBlur` on the wrapper `div` that checks `relatedTarget` before closing
- `SearchBar.tsx`: Removed stale comment describing the known limitation (now fixed)

## Test plan
- [ ] Type in the search bar to open autocomplete dropdown
- [ ] Press Tab — dropdown stays open and focus moves to first suggestion
- [ ] Press Tab again to cycle through suggestions
- [ ] Press Enter on a focused suggestion — it is selected
- [ ] Click outside the search bar — dropdown closes normally
- [ ] Press Escape — dropdown closes normally

Closes #42

### DIFF
--- a/frontend/src/components/Search/SearchBar.tsx
+++ b/frontend/src/components/Search/SearchBar.tsx
@@ -61,7 +61,14 @@ export default function SearchBar({ filters, onFiltersChange, stages }: SearchBa
     <div className="px-4 py-3 space-y-2">
       <div className="flex items-center gap-2">
         {/* Search input */}
-        <div className="relative flex-1 max-w-md">
+        <div
+          className="relative flex-1 max-w-md"
+          onBlur={(e) => {
+            if (!e.currentTarget.contains(e.relatedTarget as Node)) {
+              setInputFocused(false);
+            }
+          }}
+        >
           <Search size={16} className="absolute left-3 top-1/2 -translate-y-1/2 text-gray-400" />
           <input
             type="text"
@@ -71,11 +78,6 @@ export default function SearchBar({ filters, onFiltersChange, stages }: SearchBa
               setInputFocused(true);
               triggerInit();
             }}
-            // Note: onBlur unconditionally hides the dropdown. Tab-key navigation
-            // from the input to a suggestion button will not work because blur fires
-            // before the button receives focus. A relatedTarget check is needed for
-            // full keyboard accessibility if that use case is added in the future.
-            onBlur={() => setInputFocused(false)}
             placeholder="Search companies, roles, tech stack..."
             className="w-full rounded-lg border border-gray-300 bg-white py-2 pl-9 pr-8 text-sm transition focus:border-primary-500 focus:outline-none focus:ring-2 focus:ring-primary-500/20"
           />
--- a/frontend/src/components/Search/SearchBar.tsx
+++ b/frontend/src/components/Search/SearchBar.tsx
@@ -64,7 +64,7 @@ export default function SearchBar({ filters, onFiltersChange, stages }: SearchBa
         <div
           className="relative flex-1 max-w-md"
           onBlur={(e) => {
-            if (!e.currentTarget.contains(e.relatedTarget as Node)) {
+            if (!e.currentTarget.contains(e.relatedTarget as Node | null)) {
               setInputFocused(false);
             }
           }}


tomerp20 (commented):
## Code Review — PR #54: fix: preserve Tab-key navigation in autocomplete dropdown

### Overview
This PR fixes a real accessibility bug (issue #42): the unconditional `onBlur` on the `<input>` was closing the autocomplete dropdown before Tab-focused suggestion buttons could receive focus. The approach — moving `onBlur` to the wrapper `<div>` and guarding with `e.currentTarget.contains(e.relatedTarget)` — is the correct React pattern for this problem. The stale TODO comment is also cleaned up.

---

### Issues Found

**1. Redundant dismiss mechanism in `AutocompleteDropdown` — now superseded (minor)**

`AutocompleteDropdown` already has a `document.mousedown` listener that calls `onDismiss()` for outside clicks. With the wrapper-div `onBlur` now correctly handling all focus-leave cases (keyboard and mouse), the `mousedown` listener in `AutocompleteDropdown` is partially redundant for the mouse-click-outside case. This is not a regression — both handlers cover it safely — but it creates two code paths for the same dismissal. Not a blocker, just worth noting for future cleanup.

**2. `e.relatedTarget as Node` cast is safe but can be tightened (low severity)**

```tsx
if (!e.currentTarget.contains(e.relatedTarget as Node)) {
```

`e.relatedTarget` can be `null` (e.g. focus moving to a non-focusable element or leaving the page). `Node.contains(null)` returns `false` in all modern browsers, so the cast is safe in practice. However, the `as Node` cast suppresses TypeScript's null-safety checks. A more defensive, zero-risk alternative:

```tsx
if (!e.currentTarget.contains(e.relatedTarget as Node | null)) {
```

or simply use optional chaining via a helper to make the null intent explicit. This is a low-severity nit, not a functional issue.

**3. `showDropdown` gating still relies on `inputFocused` (observation, not a regression)**

The dropdown is only shown while `inputFocused === true`. With the fix, `inputFocused` stays `true` when tabbing into the dropdown, which is correct. However, if a suggestion button's `onClick` fires and calls `handleSelect` → `setInputFocused(false)`, the dropdown closes immediately — which is the intended "committed selection" path. This is fine; just confirming no regression here.

---

### Code Quality
- Correct use of React's synthetic bubbling `onBlur` on the container div — idiomatic and well-understood pattern.
- Removal of the stale comment is clean; the comment described the very limitation now fixed.
- Change is minimal and focused — no unrelated modifications.

### Security
No security concerns.

### Performance
No performance concerns — no extra renders, no new state, no new effects.

### Verdict
The fix is **correct and safe**. The `relatedTarget` cast nit (point 2) is the only thing worth addressing before merge.

PR #65, opened 2026-04-19
feat: add Knex migration for PostgreSQL card notify trigger

### LINKED ISSUE #60: feat: create Knex migration for PostgreSQL card notify trigger
## Parent

## What
Create Knex migration `backend/migrations/009_card_notify_trigger.ts` that installs a PL/pgSQL trigger function on the `cards` table.

## Details

The migration creates:
1. A function `notify_card_event()` that fires `pg_notify('card_events', payload)` with:
```json
{ "event": "card.created|card.updated|card.moved|card.deleted", "card_id": "...", "user_id": "..." }
```

Event type logic:
- `INSERT` → `card.created`
- `DELETE` → `card.deleted`
- `UPDATE` where `OLD.stage_id IS DISTINCT FROM NEW.stage_id` → `card.moved`
- `UPDATE` otherwise → `card.updated`

2. A trigger `card_notify_trigger` — `AFTER INSERT OR UPDATE OR DELETE ON cards FOR EACH ROW`

The `down()` function drops the trigger then the function.

## Verification
- Run `npm run migrate` — no errors
- Use `psql` to insert/update/delete a card row and confirm the trigger fires (can temporarily add a `LISTEN card_events` in psql session)

## Files
- **Create** `backend/migrations/009_card_notify_trigger.ts`

tomerp20:
## Summary
- Adds `backend/migrations/009_card_notify_trigger.ts`, a Knex migration that installs a PL/pgSQL trigger on the `cards` table
- The `notify_card_event()` function fires `pg_notify('card_events', payload)` on every INSERT, UPDATE, and DELETE, emitting `card.created`, `card.updated`, `card.moved`, or `card.deleted` events
- The `down()` function idempotently drops both the trigger and the function

## Changes
- New migration file `009_card_notify_trigger.ts` with `up()` and `down()` implementations
- `up()`: creates the PL/pgSQL function with `CREATE OR REPLACE FUNCTION`, then attaches `card_notify_trigger` via `CREATE TRIGGER AFTER INSERT OR UPDATE OR DELETE ON cards FOR EACH ROW`
- `down()`: drops trigger and function with `IF EXISTS` guards

## Test plan
- [ ] Run `npm run migrate` in the backend and verify no errors
- [ ] Connect to the database and confirm `notify_card_event` function exists (`\df notify_card_event`)
- [ ] Confirm `card_notify_trigger` exists on the `cards` table (`\d cards`)
- [ ] `LISTEN card_events;` in psql, then INSERT/UPDATE/DELETE a card row and verify the correct JSON payload arrives on the channel
- [ ] Run `npm run migrate:rollback` and verify the trigger and function are removed cleanly

Closes #60

### DIFF
--- a/backend/migrations/009_card_notify_trigger.ts
+++ b/backend/migrations/009_card_notify_trigger.ts
@@ -28,6 +28,8 @@ export async function up(knex: Knex): Promise<void> {
         END IF;
       END IF;
 
+      -- pg_notify enforces an 8000-byte payload limit; this payload is
+      -- intentionally minimal (event, card_id, user_id) to stay well under it.
       payload := json_build_object(
         'event',   event_type,
         'card_id', card_id,
@@ -41,6 +43,9 @@ export async function up(knex: Knex): Promise<void> {
     $$ LANGUAGE plpgsql;
   `);
 
+  // Drop before re-creating so up() is idempotent if re-run after a partial failure.
+  await knex.raw(`DROP TRIGGER IF EXISTS card_notify_trigger ON cards;`);
+
   await knex.raw(`
     CREATE TRIGGER card_notify_trigger
     AFTER INSERT OR UPDATE OR DELETE ON cards
@@ -50,5 +55,6 @@ export async function up(knex: Knex): Promise<void> {
 
 export async function down(knex: Knex): Promise<void> {
   await knex.raw(`DROP TRIGGER IF EXISTS card_notify_trigger ON cards;`);
-  await knex.raw(`DROP FUNCTION IF EXISTS notify_card_event();`);
+  // CASCADE ensures rollback succeeds even if other objects reference the function.
+  await knex.raw(`DROP FUNCTION IF EXISTS notify_card_event() CASCADE;`);
 }
--- a/backend/migrations/009_card_notify_trigger.ts
+++ b/backend/migrations/009_card_notify_trigger.ts
@@ -1,0 +1,54 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`
+    CREATE OR REPLACE FUNCTION notify_card_event()
+    RETURNS trigger AS $$
+    DECLARE
+      payload TEXT;
+      event_type TEXT;
+      card_id TEXT;
+      user_id TEXT;
+    BEGIN
+      IF TG_OP = 'DELETE' THEN
+        card_id    := OLD.id;
+        user_id    := OLD.user_id;
+        event_type := 'card.deleted';
+      ELSIF TG_OP = 'INSERT' THEN
+        card_id    := NEW.id;
+        user_id    := NEW.user_id;
+        event_type := 'card.created';
+      ELSIF TG_OP = 'UPDATE' THEN
+        card_id    := NEW.id;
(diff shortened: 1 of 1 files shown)


tomerp20 (commented):
## Code Review — PR #65: `009_card_notify_trigger.ts`

### Overview
This migration adds a PL/pgSQL trigger function (`notify_card_event`) and a row-level AFTER trigger (`card_notify_trigger`) on the `cards` table. On every INSERT, UPDATE, or DELETE the trigger fires `pg_notify('card_events', payload)` with a JSON payload containing `event`, `card_id`, and `user_id`. The `down()` tears both down idempotently. The logic is correct and the structure matches the project's conventions well.

---

### Issues Found

**1. `pg_notify` payload is silently truncated at 8000 bytes (correctness)**

`pg_notify` enforces an 8000-byte hard limit on the channel payload. The current payload only carries three small fields so it is safe today, but if the shape ever grows (e.g. including `stage_id` before/after) it will silently truncate. Add a guard or a comment documenting the constraint:

```sql
-- pg_notify payload limit is 8000 bytes; this payload is intentionally minimal.
PERFORM pg_notify('card_events', payload);
```

**2. `CREATE TRIGGER` is not idempotent — re-running `up()` after a failed migration will error (reliability)**

`CREATE OR REPLACE FUNCTION` is safe to re-run, but `CREATE TRIGGER` is not. If the migration is applied partially (function created, then the DB connection drops before the trigger is created) and then re-run, the `CREATE TRIGGER` step will fail with `trigger "card_notify_trigger" for relation "cards" already exists`. Use `DROP TRIGGER IF EXISTS` before the `CREATE TRIGGER` to make the step idempotent:

```sql
DROP TRIGGER IF EXISTS card_notify_trigger ON cards;
CREATE TRIGGER card_notify_trigger
AFTER INSERT OR UPDATE OR DELETE ON cards
FOR EACH ROW EXECUTE FUNCTION notify_card_event();
```

**3. Missing `import type` — minor style inconsistency with some migrations**

Migration `008_create_todo_items.ts` uses `import { Knex } from 'knex'`. This PR follows the same form, which is fine and consistent. No action needed — calling this out only to confirm it was checked.

**4. `down()` drops the function without `CASCADE` — may fail if other objects depend on it (safety)**

`DROP FUNCTION IF EXISTS notify_card_event()` will fail with a dependency error if any other trigger or view references the function at rollback time. Using `CASCADE` makes rollback reliable:

```sql
DROP FUNCTION IF EXISTS notify_card_event() CASCADE;
```

---

### Positive Notes
- Correct use of `IS DISTINCT FROM` for nullable-safe `stage_id` comparison — this handles the `NULL` case that `!=` would miss.
- Payload deliberately omits large fields, keeping it under the `pg_notify` 8000-byte limit.
- `IF TG_OP = 'DELETE' THEN RETURN OLD; ELSE RETURN NEW; END IF;` is the correct return pattern for a mixed-op AFTER trigger.
- `down()` drops trigger before function, which is the correct ordering.
- File is numbered correctly as `009` in sequence.

---

### Summary
Three real issues: the non-idempotent `CREATE TRIGGER` (most important), missing `CASCADE` on `DROP FUNCTION`, and a missing comment about the `pg_notify` payload limit. All are straightforward to fix.

PR #66, opened 2026-04-19
feat: pgSubscriber service for SSE card-event delivery

### LINKED ISSUE #61: feat: implement pgSubscriber service for LISTEN/NOTIFY
## Parent

## What
Create `backend/src/services/pgSubscriber.ts` — a dedicated `pg.Client` that listens on the `card_events` channel and maintains a registry of active SSE connections per user.

### pg.Client (dedicated, outside Knex pool)
- Uses `process.env.DATABASE_URL` to connect
- Calls `LISTEN card_events` on connect
- Registers `client.on('notification', handler)` to process incoming events
- On `client.on('error')`: logs error and schedules reconnect after 5s

### SSE Client Registry
```typescript
const sseClients: Map<string, Set<Response>> = new Map();
```
- `registerClient(userId: string, res: Response)` — adds response to the user's set
- `removeClient(userId: string, res: Response)` — removes response, cleans up empty sets

### Notification Handler
- Parses `msg.payload` as JSON
- Calls `notifyUser(userId, data)` which writes `data: <json>\n\n` to all responses in that user's set

### Exports
```typescript
export const pgSubscriber = { connect, registerClient, removeClient };
```

## Verification
- Can be tested standalone by calling `connect()` and manually running `NOTIFY card_events, '{"event":"card.created","card_id":"test","user_id":"abc"}'` in psql — confirm the Node handler fires

## Files
- **Create** `backend/src/services/pgSubscriber.ts`

tomerp20:
## Summary
- Adds `backend/src/services/pgSubscriber.ts` — a dedicated `pg.Client` that LISTENs on the `card_events` PostgreSQL channel fired by the trigger introduced in migration 009
- Maintains a per-user `Map<string, Set<Response>>` of active SSE connections; exposes `registerClient` / `removeClient` for the upcoming SSE route (Task 3) to wire up
- Auto-reconnects with a 5-second delay on client error or failed initial connection

## Changes
- `backend/src/services/pgSubscriber.ts` — new service implementing `connect`, `registerClient`, `removeClient`, and internal `notifyUser`
- `backend/package.json` / `package-lock.json` — added `@types/pg` dev dependency so TypeScript compiles without errors

## Test plan
- [x] `cd backend && npx tsc --noEmit` passes with zero errors
- [ ] Manual end-to-end test: wire up SSE route (Task 3), open an SSE connection as a user, mutate a card, verify the event arrives over the stream

Closes #61

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -29,6 +29,7 @@
         "@types/jest": "^29.5.12",
         "@types/jsonwebtoken": "^9.0.7",
         "@types/node": "^22.5.0",
+        "@types/pg": "^8.20.0",
         "@types/supertest": "^6.0.2",
         "@types/uuid": "^10.0.0",
         "jest": "^29.7.0",
@@ -1308,6 +1309,18 @@
       "license": "MIT",
       "dependencies": {
         "undici-types": "~6.21.0"
+      }
+    },
+    "node_modules/@types/pg": {
+      "version": "8.20.0",
+      "resolved": "https://registry.npmjs.org/@types/pg/-/pg-8.20.0.tgz",
+      "integrity": "sha512-bEPFOaMAHTEP1EzpvHTbmwR8UsFyHSKsRisLIHVMXnpNefSbGA1bD6CVy+qKjGSqmZqNqBDV2azOBo8TgkcVow==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@types/node": "*",
+        "pg-protocol": "*",
+        "pg-types": "^2.2.0"
       }
     },
     "node_modules/@types/qs": {
--- a/backend/package.json
+++ b/backend/package.json
@@ -38,6 +38,7 @@
     "@types/jest": "^29.5.12",
     "@types/jsonwebtoken": "^9.0.7",
     "@types/node": "^22.5.0",
+    "@types/pg": "^8.20.0",
     "@types/supertest": "^6.0.2",
     "@types/uuid": "^10.0.0",
     "jest": "^29.7.0",
--- a/backend/src/services/pgSubscriber.ts
+++ b/backend/src/services/pgSubscriber.ts
@@ -2,6 +2,10 @@ import { Client } from 'pg';
 import type { Response } from 'express';
 import logger from '../config/logger';
 
+const DEFAULT_CONNECTION_STRING = 'postgresql://jobflow:jobflow@localhost:5432/jobflow';
+const BASE_RETRY_DELAY_MS = 5_000;
+const MAX_RETRY_DELAY_MS = 60_000;
+
 const sseClients = new Map<string, Set<Response>>();
 
 function registerClient(userId: string, res: Response): void {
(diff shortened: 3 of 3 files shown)


tomerp20 (commented):
## Code Review — pgSubscriber Service

### Overview
This PR introduces a dedicated `pg.Client` that subscribes to the `card_events` PostgreSQL NOTIFY channel and forwards events to active SSE connections. The architecture is correct — using a long-lived non-pooled client is the right approach for LISTEN/NOTIFY.

---

### Issues Found

**1. Missing fallback for `DATABASE_URL` (consistency + correctness)**

`pgSubscriber.ts` line 32:
```ts
const client = new Client({ connectionString: process.env.DATABASE_URL });
```
The rest of the app (`database.ts`) always falls back to the local dev connection string:
```ts
connectionString: process.env.DATABASE_URL || 'postgresql://jobflow:jobflow@localhost:5432/jobflow'
```
Without the fallback, `DATABASE_URL` being undefined in dev silently passes `undefined` to `pg.Client`, which will fall back to libpq environment variables or fail unexpectedly. The fallback should match the one used in `database.ts`.

---

**2. `notifyUser` silently drops writes to stale/closed connections**

`pgSubscriber.ts` lines 27-28:
```ts
for (const res of clients) {
  res.write(`data: ${JSON.stringify(data)}\n\n`);
}
```
`res.write()` will throw or return `false` if the underlying socket has already closed but `removeClient` was not called (e.g., due to an abrupt client disconnect where the `close` event fired but cleanup was missed). The write should be guarded — if write returns `false` or throws, the dead entry should be pruned from the Set.

---

**3. Client leaked on reconnect — old listeners accumulate**

`pgSubscriber.ts` lines 57-60:
```ts
client.on('error', (err) => {
  logger.error('pgSubscriber client error', { error: err.message });
  setTimeout(() => connect(), 5000);
});
```
After an error, `connect()` creates a brand-new `Client` and schedules reconnect, but `client.end()` is never called on the old one. The old client object and its OS socket are leaked. Even though the connection may already be broken, calling `client.end()` (ignoring the error) ensures the handle is cleaned up so Node's event loop isn't kept alive unnecessarily.

---

**4. Unbounded reconnect loop — no backoff or cap**

Both reconnect paths (initial connection failure and client error) always retry after exactly 5000 ms with no backoff and no cap on the number of retries. If the database is permanently unreachable (e.g., bad `DATABASE_URL`) this silently spins forever, filling logs. A simple exponential backoff with a max delay (e.g., cap at 60 s) would make production failures much easier to diagnose.

---

### Minor / Non-blocking

- **`notifyUser` type**: `data: object` is overly broad — `Record<string, unknown>` is the idiomatic TypeScript type for a JSON-shaped value and makes intent clearer, though this is style only.
- **No `connect()` call at startup**: `pgSubscriber.connect()` must be called from `server.ts` (similar to how `db.raw('SELECT 1')` is called there). This is expected to be wired up in Task 3, but it is worth noting so it isn't missed.

---

### Summary

| # | Severity | Finding |
|---|----------|---------|
| 1 | Medium | Missing `DATABASE_URL` fallback — will pass `undefined` to `pg.Client` in dev |
| 2 | Medium | Dead SSE connections not pruned inside `notifyUser` — `res.write` can throw on a closed socket |
| 3 | Low | Old `pg.Client` not ended on reconnect — socket handle leaked |
| 4 | Low | No reconnect backoff — infinite 5 s loop if DB is permanently down |

PR #67, opened 2026-04-19
feat: SSE events endpoint and pgSubscriber wiring

### LINKED ISSUE #62: feat: add SSE endpoint and wire pgSubscriber into app startup
## Parent

## Depends on
- Task 2 (pgSubscriber service) must be complete first

## What
Create the `GET /api/events` SSE route, mount it in the router, and initialize the `pgSubscriber` in `server.ts` at app startup.

### SSE Route (`backend/src/routes/events.ts`)
`GET /api/events?token=<jwt>`

- Auth: read JWT from `req.query.token`, verify with `jsonwebtoken` (same secret as existing middleware). Return 401 if missing/invalid.
- Set response headers:
  ```
  Content-Type: text/event-stream
  Cache-Control: no-cache
  Connection: keep-alive
  X-Accel-Buffering: no
  ```
- Send initial `data: {"event":"connected"}\n\n`
- Call `pgSubscriber.registerClient(userId, res)`
- Start heartbeat: `setInterval(() => res.write(': heartbeat\n\n'), 30_000)`
- On `req.on('close')`: clear interval, call `pgSubscriber.removeClient(userId, res)`

### Route registration (`backend/src/routes/index.ts`)
```typescript
import eventsRouter from './events';
router.use('/events', eventsRouter);
```

### App startup (`backend/src/server.ts`)
```typescript
import { pgSubscriber } from './services/pgSubscriber';
await pgSubscriber.connect(); // before app.listen()
```

## Verification
- Start the backend
- Run: `curl -N "http://localhost:3001/api/events?token=<valid_jwt>"`
- Confirm: SSE stream opens, `connected` event received, heartbeat comment appears every 30s
- Confirm: 401 returned without a valid token

## Files
- **Create** `backend/src/routes/events.ts`
- **Modify** `backend/src/routes/index.ts`
- **Modify** `backend/src/server.ts`

tomerp20:
## Summary
- Adds `GET /api/events` SSE endpoint that authenticates via `?token=<jwt>` query param (browser `EventSource` cannot set headers) and streams real-time card-event notifications
- Adds `backend/src/services/pgSubscriber.ts` — dedicated `pg.Client` that LISTENs on `card_events` and maintains a `Map<userId, Set<Response>>` of active SSE connections with exponential-backoff reconnect
- Mounts `/events` in the API router and calls `pgSubscriber.connect()` at server startup after the DB health check

## Changes
- `backend/src/routes/events.ts` (new): SSE handler — token auth, SSE headers, `flushHeaders()`, connected event, heartbeat every 30s, cleanup on `req.close`
- `backend/src/services/pgSubscriber.ts` (new): pg LISTEN service with `connect()`, `registerClient()`, `removeClient()`, and auto-reconnect
- `backend/src/routes/index.ts`: add `router.use('/events', eventsRouter)`
- `backend/src/server.ts`: `await pgSubscriber.connect()` after DB health check

## Test plan
- [ ] `cd backend && npx tsc --noEmit` — passes with no errors
- [ ] Start backend, run `curl -N "http://localhost:3001/api/events?token=<valid_jwt>"` — see `data: {"event":"connected"}` immediately, then `: heartbeat` every 30s
- [ ] Omit or send bad token — verify 401 responses before any SSE headers are set
- [ ] Mutate a card via PATCH — verify the SSE stream receives the notification payload

Closes #62

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -29,6 +29,7 @@
         "@types/jest": "^29.5.12",
         "@types/jsonwebtoken": "^9.0.7",
         "@types/node": "^22.5.0",
+        "@types/pg": "^8.20.0",
         "@types/supertest": "^6.0.2",
         "@types/uuid": "^10.0.0",
         "jest": "^29.7.0",
@@ -1308,6 +1309,18 @@
       "license": "MIT",
       "dependencies": {
         "undici-types": "~6.21.0"
+      }
+    },
+    "node_modules/@types/pg": {
+      "version": "8.20.0",
+      "resolved": "https://registry.npmjs.org/@types/pg/-/pg-8.20.0.tgz",
+      "integrity": "sha512-bEPFOaMAHTEP1EzpvHTbmwR8UsFyHSKsRisLIHVMXnpNefSbGA1bD6CVy+qKjGSqmZqNqBDV2azOBo8TgkcVow==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@types/node": "*",
+        "pg-protocol": "*",
+        "pg-types": "^2.2.0"
       }
     },
     "node_modules/@types/qs": {
--- a/backend/package.json
+++ b/backend/package.json
@@ -38,6 +38,7 @@
     "@types/jest": "^29.5.12",
     "@types/jsonwebtoken": "^9.0.7",
     "@types/node": "^22.5.0",
+    "@types/pg": "^8.20.0",
     "@types/supertest": "^6.0.2",
     "@types/uuid": "^10.0.0",
     "jest": "^29.7.0",
--- a/backend/src/middleware/auth.ts
+++ b/backend/src/middleware/auth.ts
@@ -16,7 +16,7 @@ declare global {
   }
 }
 
-interface JwtPayload {
+export interface JwtPayload {
   userId: string;
   email: string;
   iat: number;
--- a/backend/src/routes/events.ts
+++ b/backend/src/routes/events.ts
@@ -1,0 +1,80 @@
+import { Router, Request, Response } from 'express';
+import jwt from 'jsonwebtoken';
+import db from '../config/database';
+import logger from '../config/logger';
+import { pgSubscriber } from '../services/pgSubscriber';
+
+const router = Router();
+
+interface JwtPayload {
+  userId: string;
+  email: string;
+  iat: number;
+  exp: number;
+}
+
+/**
+ * GET /api/events
+ *
+ * SSE endpoint for real-time card-event notifications. The browser
+ * EventSource API cannot set custom headers, so the JWT is passed as a
+ * query parameter: ?token=<jwt>
+ */
+router.get('/', async (req: Request, res: Response): Promise<void> => {
+  const token = req.query.token as string;
+
+  if (!token) {
+    res.status(401).json({ error: 'Missing token' });
+    return;
+  }
+
+  const secret = process.env.JWT_SECRET;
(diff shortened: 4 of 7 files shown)


tomerp20 (commented):
## Code Review — PR #67: SSE Events Endpoint + pgSubscriber Wiring

### Overview
This PR wires together the real-time card-event pipeline: a dedicated `pg.Client` that LISTENs on `card_events`, a `GET /api/events` SSE endpoint, and startup initialization in `server.ts`. The structure is sound and the code is readable. A few correctness and security issues need addressing.

---

### Issues Found

#### 1. Security — JWT token exposed in server access logs (events.ts, line 24)
Passing the JWT as a query parameter (`?token=<jwt>`) causes the full token to appear in server access logs, nginx logs, and browser history. This is a known security risk.

**Recommendation:** Add a comment documenting this risk and ensure any access log middleware (morgan, etc.) is configured to redact the `token` query param on this route.

#### 2. Correctness — unhandled rejection from DB call in events.ts (lines 47–55)
The `db('users').select(...).first()` call is inside an `async` handler but there is no `try/catch` around it. The rest of the codebase (e.g. `cards.ts`) wraps every async handler body in `try/catch { next(err) }`. This route should do the same for the auth portion, passing a `NextFunction` parameter and catching errors before SSE headers are flushed.

#### 3. Correctness — hardcoded default credentials in pgSubscriber.ts (line 5)
`DEFAULT_CONNECTION_STRING = 'postgresql://jobflow:jobflow@localhost:5432/jobflow'` contains a plaintext password. If `DATABASE_URL` is not set the service silently uses these credentials. The Knex db config fails loudly on missing env vars; this service should too.

**Recommendation:** Remove the default and throw if `DATABASE_URL` is missing:
```typescript
const connectionString = process.env.DATABASE_URL;
if (!connectionString) {
  throw new Error('DATABASE_URL environment variable is required');
}
```

#### 4. Correctness — `await pgSubscriber.connect()` is misleading (server.ts, line 21)
`connect()` swallows its own connection error via `scheduleReconnect` and always resolves immediately regardless of whether the pg connection succeeded. The `await` implies the server waits for a confirmed connection, which it does not. The server starts accepting traffic even when the subscriber is still connecting.

**Recommendation:** Drop the `await` and add a clarifying comment:
```typescript
pgSubscriber.connect(); // non-blocking: retries in background on failure
```

#### 5. Code quality — JwtPayload interface is duplicated (events.ts, lines 9–14)
The same `JwtPayload` interface already exists in `backend/src/middleware/auth.ts`. Duplicating it creates a maintenance risk. Extract it to a shared types file (e.g. `backend/src/types/auth.ts`) and import it in both files.

#### 6. Code quality — write() returning false is not a dead connection (pgSubscriber.ts, lines 31–46)
When `res.write()` returns `false` it signals backpressure (the internal buffer is full), not a broken connection. The stream is still writable and will drain. Treating `false` as a dead connection silently drops that client, causing them to miss events. Only a thrown exception reliably indicates a broken pipe.

**Recommendation:** Remove the `!ok` path:
```typescript
for (const res of clients) {
  try {
    res.write(`data: ${JSON.stringify(data)}\n\n`);
  } catch {
    dead.push(res);
  }
}
```

---

### No Issues
- SSE headers are correct and complete (`X-Accel-Buffering: no` is a good addition)
- Heartbeat interval and cleanup on `req.on('close')` are correctly structured
- Exponential backoff with cap in `scheduleReconnect` is correct
- Router mounting in `index.ts` follows existing conventions
- TypeScript compiles cleanly (`npx tsc --noEmit` passes)

PR #68, opened 2026-04-19
feat: useCardEvents hook for real-time SSE board updates (#63)

### LINKED ISSUE #63: feat: implement useCardEvents React hook
## Parent

## Depends on
- Task 3 (SSE endpoint) must be complete first

## What
Create `frontend/src/hooks/useCardEvents.ts` — a React hook that opens an `EventSource` connection and surgically updates board state on incoming card events.

### Hook signature
```typescript
interface UseCardEventsProps {
  stages: Stage[];
  cards: Card[];
  setCards: React.Dispatch<React.SetStateAction<Card[]>>;
}

export function useCardEvents({ stages, cards, setCards }: UseCardEventsProps): void
```

### EventSource setup (inside useEffect)
- Read JWT from `localStorage` key `accessToken`
- Open `new EventSource(\`/api/events?token=${token}\`)`
- Register `es.onmessage` handler
- `es.onerror`: log error (EventSource auto-reconnects by default)
- Cleanup: `return () => es.close()`

### Event handlers (inside onmessage)
Parse `JSON.parse(event.data)` then switch on `payload.event`:

| Event | Action |
|-------|--------|
| `card.created` | Call `api.getCard(cardId)`, prepend card to state |
| `card.updated` | Call `api.getCard(cardId)`, replace card in state |
| `card.moved` | Call `api.getCard(cardId)`, replace card in state (board derives column from `stageId`) |
| `card.deleted` | `setCards(prev => prev.filter(c => c.id !== cardId))` — no fetch needed |
| `connected` | No-op (connection confirmation) |

Note: `api.getCard(id)` already exists in `frontend/src/services/api.ts` and returns `{ card, activities }`.

## Verification
- Hook is wired in BoardPage (next task), so verify after Task 5
- Can unit-test event handlers in isolation by mocking `EventSource` and `api.getCard`

## Files
- **Create** `frontend/src/hooks/useCardEvents.ts`

tomerp20:
## Summary
- Implements the `useCardEvents` React hook that opens an `EventSource` connection to `/api/events?token=<jwt>` and surgically updates board card state on incoming server-sent events
- Handles `card.created`, `card.updated`, `card.moved`, `card.deleted`, and `connected` event types
- Returns early (no connection) when no access token is present in localStorage

## Changes
- `frontend/src/hooks/useCardEvents.ts` — new hook with single `useEffect` whose dependency array is `[]` so the SSE connection is opened once per mount and torn down on unmount
- `card.created` / `card.updated` / `card.moved` each call `cardsApi.getCard(id)` to get the fully transformed `Card` object before mutating state
- `card.deleted` filters without a network fetch
- `es.onerror` logs the error; browser-native auto-reconnect handles transient failures

## Test plan
- [ ] Hook connects to SSE stream after login and logs `connected` event (no-op)
- [ ] Creating a card in another tab/session causes it to appear on the board without a page refresh
- [ ] Updating a card in another session updates the card in place on the board
- [ ] Moving a card in another session reflects the correct column on the board
- [ ] Deleting a card in another session removes it from the board
- [ ] Logging out and back in re-opens the connection with the new token
- [ ] `npx tsc --noEmit` passes with no errors

Closes #63

### DIFF
--- a/frontend/src/hooks/useCardEvents.ts
+++ b/frontend/src/hooks/useCardEvents.ts
@@ -3,7 +3,6 @@ import type { Card } from '@/types';
 import { cardsApi } from '@/services/api';
 
 interface UseCardEventsProps {
-  cards: Card[];
   setCards: React.Dispatch<React.SetStateAction<Card[]>>;
 }
 
@@ -12,7 +11,7 @@ export function useCardEvents({ setCards }: UseCardEventsProps): void {
     const token = localStorage.getItem('accessToken');
     if (!token) return;
 
-    const es = new EventSource(`/api/events?token=${token}`);
+    const es = new EventSource(`/api/events?token=${encodeURIComponent(token)}`);
 
     es.onmessage = async (event: MessageEvent) => {
       let payload: { event: string; cardId: string };
@@ -29,7 +28,11 @@ export function useCardEvents({ setCards }: UseCardEventsProps): void {
         case 'card.created': {
           try {
             const { card } = await cardsApi.getCard(cardId);
-            setCards((prev) => [...prev, card]);
+            // Guard against duplicates — the optimistic update from the local
+            // create handler may have already inserted this card into state.
+            setCards((prev) =>
+              prev.some((c) => c.id === card.id) ? prev : [...prev, card]
+            );
           } catch (err) {
             console.error('[useCardEvents] Failed to fetch created card:', err);
           }
@@ -63,12 +66,22 @@ export function useCardEvents({ setCards }: UseCardEventsProps): void {
 
     es.onerror = (err) => {
       console.error('[useCardEvents] SSE connection error:', err);
-      // EventSource auto-reconnects by default; no manual reconnect needed
+      // EventSource auto-reconnects by default. On reconnect it will reuse the
+      // same URL (with the original token). If the access token has been
+      // silently refreshed while the connection was open, close this instance
+      // so the parent can re-mount with the current token.
+      const currentToken = localStorage.getItem('accessToken');
+      if (currentToken && currentToken !== token) {
+        es.close();
+      }
     };
 
     return () => {
       es.close();
     };
-  // eslint-disable-next-line react-hooks/exhaustive-deps
+    // setCards is a stable useState dispatcher — safe to omit from deps.
+    // The effect intentionally runs once per mount so a single SSE connection
+    // is opened for the lifetime of the component.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 }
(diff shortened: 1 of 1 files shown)


tomerp20 (commented):
## Code Review — `useCardEvents` hook (PR #68)

### Overview
The hook correctly implements the SSE-based real-time board update pattern: opens one `EventSource` per mount, dispatches to `cardsApi.getCard` for mutations, filters for deletes, and cleans up via `es.close()`. The core logic is sound. A few issues to address:

---

### Issues Found

**1. `cards` prop accepted but never used — dead parameter in the interface**

`UseCardEventsProps` declares `cards: Card[]` in the interface, the hook signature destructures only `setCards`, and `cards` is never referenced anywhere in the hook body. This is dead API surface that misleads callers into thinking the array is needed.

**2. Token is read once at mount — stale after a token refresh**

The token is captured in `localStorage.getItem('accessToken')` once when the effect runs (mount). If the Axios interceptor silently refreshes the token (which it does in `api.ts`), `localStorage` will hold a new token but the EventSource URL still carries the old one. The old token may expire mid-session, causing the SSE endpoint to reject reconnects with 401. The token should be read inside `es.onerror` (or the connection should be re-established when the token changes) so reconnects use a fresh token.

**3. Token passed in URL — no encoding**

```ts
const es = new EventSource(`/api/events?token=${token}`);
```

JWT tokens contain `.` delimiters and base64url characters that are URL-safe, but good practice (and the spec) calls for `encodeURIComponent(token)` to be defensive against any edge-case tokens or future token format changes.

**4. `card.created` may add a duplicate if the local optimistic update already inserted the card**

`BoardPage.handleCardCreated` already appends the card to state. If the SSE event fires after the local create (same user, same tab), `card.created` will `cardsApi.getCard` and append a second copy. A deduplication guard (`prev.some(c => c.id === card.id)`) should be added before pushing.

**5. `// eslint-disable-next-line react-hooks/exhaustive-deps` suppresses a legitimate lint warning without explanation of the intent**

`setCards` (a `useState` dispatcher) is stable across renders and doesn't need to be in the dependency array. The comment explains the `[]` choice for _when_ to reconnect but not _why_ `setCards` was deliberately omitted. This can mislead future editors. Either add a brief inline rationale or use the standard pattern of wrapping `setCards` in a ref to avoid the disable comment entirely.

---

### Minor / Style

- The `default: break;` branch is idiomatic but an empty `default:` (no `break`) is equally valid and one line shorter — minor, not a blocker.
- No issues with naming, TypeScript types, cleanup, or error-logging style.

---

### Summary

| # | Severity | Issue |
|---|----------|-------|
| 1 | Low | `cards` prop declared but unused — dead interface member |
| 2 | Medium | Token captured at mount only — stale after silent refresh |
| 3 | Low | Token not URL-encoded before interpolation |
| 4 | Medium | `card.created` can produce duplicate entries for same-tab creates |
| 5 | Low | eslint-disable without explanation of the intentional omission |

PR #70, opened 2026-04-19
fix: fix SSE connection in production

tomerp20:
## Summary
- Fixes `useCardEvents` to use `VITE_API_URL` env var instead of a hardcoded relative URL, so the `EventSource` connects to the correct backend host in production
- Reduces SSE heartbeat interval from 30s to 25s to reliably beat Render's proxy timeout

## Root Cause
Two silent failures in production:

**1. Wrong EventSource URL**
The hook used `/api/events` (relative). In dev, Vite proxies this to `localhost:3001`. In production there is no proxy — the browser hits the frontend's own origin, which has no SSE endpoint. The Axios instance already reads `VITE_API_URL` correctly; the `EventSource` now does the same.

**2. Heartbeat race with Render timeout**
Render's infrastructure drops connections with no response data after ~30 seconds. The heartbeat fired at exactly 30s — too close to the cutoff. Reduced to 25s.

## Changes
- `frontend/src/hooks/useCardEvents.ts` — use `import.meta.env.VITE_API_URL || '/api'` as base URL
- `backend/src/routes/events.ts` — heartbeat interval `30_000` → `25_000`

## Test plan
- [ ] Confirm `VITE_API_URL` is set in the frontend's Render/static-site env vars pointing to the backend URL
- [ ] Deploy and open two tabs — create/move/delete a card in one tab, confirm it updates in the other
- [ ] Check Network tab: `GET <backend>/api/events` shows as open EventStream

### DIFF
--- a/backend/src/routes/events.ts
+++ b/backend/src/routes/events.ts
@@ -73,7 +73,7 @@ router.get('/', async (req: Request, res: Response, next: NextFunction): Promise
 
   pgSubscriber.registerClient(userId, res);
 
-  const heartbeat = setInterval(() => res.write(': heartbeat\n\n'), 30_000);
+  const heartbeat = setInterval(() => res.write(': heartbeat\n\n'), 25_000);
 
   req.on('close', () => {
     clearInterval(heartbeat);
--- a/frontend/src/hooks/useAutocomplete.ts
+++ b/frontend/src/hooks/useAutocomplete.ts
@@ -1,5 +1,6 @@
 import { useState, useRef, useCallback, useEffect } from 'react';
 import trieWorker from '@/workers/trieWorker';
+import { API_BASE_URL } from '@/services/api';
 
 // Sequence counter shared across all instances so each query can be
 // matched back to the instance that issued it.
@@ -55,8 +56,7 @@ export function useAutocomplete(): UseAutocompleteResult {
     initializedRef.current = true;
     setIsLoading(true);
 
-    const baseUrl = import.meta.env.VITE_API_URL ?? '/api';
-    fetch(`${baseUrl}/autocomplete/words`)
+    fetch(`${API_BASE_URL}/autocomplete/words`)
       .then((res) => {
         if (!res.ok) {
           throw new Error(`Failed to fetch words: ${res.status}`);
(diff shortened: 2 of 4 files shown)


tomerp20 (commented):
## Code Review — PR #70: fix SSE connection in production

**Files reviewed:** `frontend/src/hooks/useCardEvents.ts`, `backend/src/routes/events.ts`

---

### frontend/src/hooks/useCardEvents.ts

**Summary:** Replaces the hardcoded relative URL `/api/events` with `${import.meta.env.VITE_API_URL || '/api'}/events`, matching the pattern already used by the Axios instance in `api.ts` and by `useAutocomplete.ts`.

#### Issues Found:

**[LOW] — Duplicated base URL derivation is now spread across three files**

- **Description:** The expression `import.meta.env.VITE_API_URL || '/api'` is now repeated in `api.ts` (line 6, module-level Axios `baseURL`), `api.ts` (line 66, inside the refresh interceptor), `useAutocomplete.ts` (line 58), and now `useCardEvents.ts` (line 14). This is a build-time constant that should be defined once and imported. If the fallback value or the env var name ever changes, four sites need to be updated in sync. This PR adds a fourth copy, making the drift risk slightly worse.

  Current code:
  ```ts
  const baseUrl = import.meta.env.VITE_API_URL || '/api';
  const es = new EventSource(`${baseUrl}/events?token=${encodeURIComponent(token)}`);
  ```

  Improved code — extract to a shared constant in `api.ts` and import it:
  ```ts
  // In api.ts — add and export alongside the axios instance:
  export const API_BASE_URL = import.meta.env.VITE_API_URL || '/api';

  // In useCardEvents.ts:
  import { API_BASE_URL } from '@/services/api';
  // ...
  const es = new EventSource(`${API_BASE_URL}/events?token=${encodeURIComponent(token)}`);
  ```

  This removes the duplication and makes `api.ts` the single source of truth for the base URL.

---

### backend/src/routes/events.ts

**Summary:** Heartbeat interval reduced from `30_000` ms to `25_000` ms to keep the SSE connection alive before Render's proxy drops it at ~30s.

No issues found. The change is minimal, correct, and well-motivated. 25s provides a 5s safety margin against the proxy timeout, which is appropriate.

---

## Summary

| Severity | Count | Description |
|----------|-------|-------------|
| Critical | 0 | — |
| High | 0 | — |
| Medium | 0 | — |
| Low | 1 | Duplicated `VITE_API_URL \|\| '/api'` expression — this PR adds a fourth copy |

Both production bug fixes are correct. The Low finding is a pre-existing pattern that this PR extends — the fix is straightforward and reduces future maintenance risk.

PR #78, opened 2026-05-06
feat: add role title autocomplete to card creation form

### LINKED ISSUE #77: implement: role title autocomplete in card creation form
## Task
Implement the role title autocomplete feature as specified in the PRD: #76

## What to build

Wire autocomplete suggestions into the "Role Title" input in `CardForm.tsx`, sourced from the user's existing cards already loaded in `BoardPage`.

## Files to touch

| File | Change |
|------|--------|
| `frontend/src/hooks/useStringAutocomplete.ts` | **NEW** — generic hook |
| `frontend/src/pages/BoardPage.tsx` | Derive + pass `roleTitleSuggestions` prop |
| `frontend/src/components/Card/CardForm.tsx` | Wire autocomplete + keyboard nav |

## Technical spec

**`useStringAutocomplete(candidates: string[], query: string): string[]`**
- Case-insensitive prefix filter (`startsWith`)
- Returns up to 5 results
- No trie, no Web Worker — plain array filter

**`BoardPage.tsx`**
- Derive `roleTitleSuggestions` from existing `cards` state
- Deduplicate with `Set`, sort alphabetically
- Pass as prop to `CardForm`

**`CardForm.tsx`**
- Accept `roleTitleSuggestions: string[]` prop
- Use `useStringAutocomplete` hook with current role title input value
- Render existing `AutocompleteDropdown` component when suggestions exist
- Trigger: show suggestions only after 1+ character typed
- Keyboard: Arrow Up/Down to navigate, Enter to select, Escape to dismiss

## Reuse existing components
- `frontend/src/components/AutocompleteDropdown.tsx` — use as-is for the dropdown UI
- Do NOT reuse `useAutocomplete.ts` — that is for Hebrew dictionary search via Web Worker

## No backend changes needed

Closes #76

tomerp20:
## Summary
- Adds a generic `useStringAutocomplete` hook (case-insensitive prefix match, up to 5 results)
- Derives `roleTitleSuggestions` from the already-loaded `cards` state in `BoardPage` and passes it to `CardForm`
- Renders the existing `AutocompleteDropdown` below the Role Title input with keyboard navigation (ArrowUp/Down, Enter, Escape) and click-to-select

## Changes
- **New:** `frontend/src/hooks/useStringAutocomplete.ts` — pure, reusable hook, no external deps
- **Modified:** `frontend/src/pages/BoardPage.tsx` — derive deduplicated, sorted role title list from cards; pass as prop
- **Modified:** `frontend/src/components/Card/CardForm.tsx` — accept `roleTitleSuggestions` prop, add `highlightedIndex` state, keyboard handling, and dropdown rendering

## Test plan
- [ ] Open the card creation form with at least one existing card that has a role title
- [ ] Type the first letter of that role title — the dropdown should appear with matching suggestions
- [ ] Use ArrowDown/ArrowUp to navigate; Enter should fill the field and close the dropdown
- [ ] Press Escape — dropdown should close without changing the field value
- [ ] Click a suggestion — field should be filled and dropdown should close
- [ ] Verify no dropdown appears when the role title field is empty
- [ ] Verify no dropdown appears when there are no matching suggestions
- [ ] Verify TypeScript compiles cleanly (`npx tsc --noEmit`)

Closes #77

### DIFF
--- a/frontend/src/components/AutocompleteDropdown.tsx
+++ b/frontend/src/components/AutocompleteDropdown.tsx
@@ -4,6 +4,7 @@ import { Loader2 } from 'lucide-react';
 interface AutocompleteDropdownProps {
   suggestions: string[];
   isLoading: boolean;
+  highlightedIndex?: number;
   onSelect: (word: string) => void;
   onDismiss: () => void;
 }
@@ -11,6 +12,7 @@ interface AutocompleteDropdownProps {
 export default function AutocompleteDropdown({
   suggestions,
   isLoading,
+  highlightedIndex = -1,
   onSelect,
   onDismiss,
 }: AutocompleteDropdownProps) {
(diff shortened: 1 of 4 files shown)


tomerp20 (commented):
## Code Review — role title autocomplete

Overall the implementation is clean and follows the spec well. Three issues to address:

---

### Issue 1 — `roleTitleSuggestions` computed inline on every render (efficiency)

**File:** `frontend/src/pages/BoardPage.tsx`

```tsx
roleTitleSuggestions={Array.from(new Set(cards.map((c) => c.roleTitle).filter(Boolean))).sort()}
```

This expression runs on **every render** of `BoardPage` — including card moves, filter changes, SSE updates, etc. — even though the set of role titles only changes when a card is added or removed. It should be wrapped in `useMemo` with `[cards]` as the dependency.

```tsx
const roleTitleSuggestions = useMemo(
  () => Array.from(new Set(cards.map((c) => c.roleTitle))).sort(),
  [cards],
);
```

---

### Issue 2 — Keyboard highlight index not reflected visually in dropdown (UX / accessibility)

**File:** `frontend/src/components/Card/CardForm.tsx`

`highlightedIndex` is tracked in state and drives which item is selected on Enter, but the `AutocompleteDropdown` component receives no `highlightedIndex` prop. The `aria-selected` attribute is hardcoded to `false` for all items, and there is no visual class applied to the active item. A user pressing ArrowDown gets no feedback about which suggestion will be committed on Enter.

The `AutocompleteDropdown` component needs to accept an optional `highlightedIndex` prop and apply a highlight class (e.g. `bg-primary-50 text-primary-700`) to the matching list item. The `CardForm` should pass `highlightedIndex` through.

---

### Issue 3 — `filter(Boolean)` is unnecessary (minor quality)

**File:** `frontend/src/pages/BoardPage.tsx`

`Card.roleTitle` is typed as `string` (non-optional), so it can never be `undefined` or `null`. The `.filter(Boolean)` call is dead code and adds noise.

```tsx
// Before
cards.map((c) => c.roleTitle).filter(Boolean)

// After — no filter needed
cards.map((c) => c.roleTitle)
```

---

### Issue 4 — `useStringAutocomplete` naming (minor / informational)

**File:** `frontend/src/hooks/useStringAutocomplete.ts`

The function contains no React hooks (`useState`, `useEffect`, etc.) so the `use` prefix is technically misleading per React's rules-of-hooks convention. The React linter will warn if `eslint-plugin-react-hooks` is enabled. Since the spec dictates this name, renaming it would conflict with the spec — leaving this as informational only.

PR #80, opened 2026-05-10
fix: stage column reorder bounces back due to unique constraint violation

tomerp20:
## Summary
- Column reordering in the kanban board appeared to work in the UI but always snapped back to the original order
- Root cause: `UNIQUE(user_id, position)` constraint on the `stages` table was checked immediately after each individual `UPDATE`, causing a constraint violation mid-transaction when positions were reassigned
- Fix: make the constraint deferrable so it is only enforced at commit time

## Changes
- `backend/migrations/010_defer_stages_position_unique.ts` — drops and recreates the unique constraint as `DEFERRABLE INITIALLY DEFERRED`
- `backend/src/services/stageService.ts` — adds `SET CONSTRAINTS stages_user_id_position_unique DEFERRED` at the start of the reorder transaction

## Test plan
- [ ] Drag a column to a new position — it should stay there after release
- [ ] Reorder multiple columns in succession — all should persist
- [ ] Refresh the page after reordering — columns should remain in the new order
- [ ] Verify other stage operations (create, rename, delete) are unaffected

### DIFF
--- a/backend/migrations/010_defer_stages_position_unique.ts
+++ b/backend/migrations/010_defer_stages_position_unique.ts
@@ -3,22 +3,24 @@ import { Knex } from 'knex';
 export async function up(knex: Knex): Promise<void> {
   await knex.raw(`
     ALTER TABLE stages
-      DROP CONSTRAINT stages_user_id_position_unique;
-
+      DROP CONSTRAINT IF EXISTS stages_user_id_position_unique
+  `);
+  await knex.raw(`
     ALTER TABLE stages
       ADD CONSTRAINT stages_user_id_position_unique
       UNIQUE (user_id, position)
-      DEFERRABLE INITIALLY DEFERRED;
+      DEFERRABLE INITIALLY DEFERRED
   `);
 }
 
 export async function down(knex: Knex): Promise<void> {
   await knex.raw(`
     ALTER TABLE stages
-      DROP CONSTRAINT stages_user_id_position_unique;
-
+      DROP CONSTRAINT IF EXISTS stages_user_id_position_unique
+  `);
+  await knex.raw(`
     ALTER TABLE stages
       ADD CONSTRAINT stages_user_id_position_unique
-      UNIQUE (user_id, position);
+      UNIQUE (user_id, position)
   `);
 }
--- a/backend/migrations/010_defer_stages_position_unique.ts
+++ b/backend/migrations/010_defer_stages_position_unique.ts
@@ -1,0 +1,24 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`
+    ALTER TABLE stages
+      DROP CONSTRAINT stages_user_id_position_unique;
+
+    ALTER TABLE stages
+      ADD CONSTRAINT stages_user_id_position_unique
+      UNIQUE (user_id, position)
+      DEFERRABLE INITIALLY DEFERRED;
+  `);
+}
+
+export async function down(knex: Knex): Promise<void> {
+  await knex.raw(`
+    ALTER TABLE stages
+      DROP CONSTRAINT stages_user_id_position_unique;
+
+    ALTER TABLE stages
+      ADD CONSTRAINT stages_user_id_position_unique
+      UNIQUE (user_id, position);
+  `);
+}
--- a/backend/src/services/stageService.ts
+++ b/backend/src/services/stageService.ts
@@ -158,6 +158,7 @@ export const stageService = {
     }
 
     await db.transaction(async (trx) => {
+      await trx.raw('SET CONSTRAINTS stages_user_id_position_unique DEFERRED');
       for (let i = 0; i < stageIds.length; i++) {
         await trx('stages')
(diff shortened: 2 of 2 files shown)


tomerp20 (commented):
## Code Review — fix/stage-reorder-deferred-constraint

### Files Reviewed
- `backend/migrations/010_defer_stages_position_unique.ts`
- `backend/src/services/stageService.ts`

---

### 📄 backend/migrations/010_defer_stages_position_unique.ts

**Summary:** New migration that drops and recreates the `stages_user_id_position_unique` constraint as `DEFERRABLE INITIALLY DEFERRED`.

---

#### [MEDIUM] — Multi-statement `knex.raw()` in `up()` and `down()` should be split into separate calls

**Description:** Both `up()` and `down()` bundle a `DROP CONSTRAINT` and an `ADD CONSTRAINT` into a single multi-statement string passed to `knex.raw()`. While PostgreSQL will wrap this in a transaction, some versions of the `pg` driver send multi-statement strings with `simple_query` protocol, which does not guarantee atomic execution in all connection-pooling configurations (including Neon's pooler). More practically: if the `ADD CONSTRAINT` fails (e.g., existing duplicate `(user_id, position)` rows in prod), the migration errors mid-string and the constraint is already gone — leaving the table unprotected with no clear rollback signal from Knex. Splitting into two `await knex.raw()` calls makes each statement explicit, keeps the migration readable, and matches the established pattern in this codebase's other raw migrations.

**Current code:**
```ts
await knex.raw(`
  ALTER TABLE stages
    DROP CONSTRAINT stages_user_id_position_unique;

  ALTER TABLE stages
    ADD CONSTRAINT stages_user_id_position_unique
    UNIQUE (user_id, position)
    DEFERRABLE INITIALLY DEFERRED;
`);
```

**Improved code:**
```ts
await knex.raw(`
  ALTER TABLE stages
    DROP CONSTRAINT stages_user_id_position_unique
`);
await knex.raw(`
  ALTER TABLE stages
    ADD CONSTRAINT stages_user_id_position_unique
    UNIQUE (user_id, position)
    DEFERRABLE INITIALLY DEFERRED
`);
```
Apply the same split to `down()`.

---

#### [LOW] — `down()` should guard the `DROP CONSTRAINT` against a missing constraint

**Description:** If `down()` runs against a DB where the constraint was already dropped or never existed (e.g., a partial migration failure), the bare `DROP CONSTRAINT` will throw `ERROR: constraint "stages_user_id_position_unique" of relation "stages" does not exist`. Adding `IF EXISTS` makes rollback idempotent, consistent with the `dropTableIfExists` pattern used in every other `down()` in this codebase.

**Current code:**
```ts
await knex.raw(`
  ALTER TABLE stages
    DROP CONSTRAINT stages_user_id_position_unique;
  ...
`);
```

**Improved code:**
```ts
await knex.raw(`
  ALTER TABLE stages DROP CONSTRAINT IF EXISTS stages_user_id_position_unique
`);
await knex.raw(`
  ALTER TABLE stages
    ADD CONSTRAINT stages_user_id_position_unique
    UNIQUE (user_id, position)
`);
```
Apply `IF EXISTS` to the `DROP CONSTRAINT` in `up()` as well, for the same reason.

---

### 📄 backend/src/services/stageService.ts

**Summary:** Adds `SET CONSTRAINTS stages_user_id_position_unique DEFERRED` at the start of the `reorderStages` transaction.

---

#### [LOW] — `SET CONSTRAINTS ... DEFERRED` is redundant once the constraint is `INITIALLY DEFERRED`

**Description:** Migration 010 defines the constraint as `DEFERRABLE INITIALLY DEFERRED`. A constraint marked `INITIALLY DEFERRED` is automatically deferred for every transaction without any explicit `SET CONSTRAINTS` call. The added line is therefore a no-op. It creates misleading noise — a future reader might wonder whether the migration ran, or whether the line is a safety net for a different constraint mode. It should be removed. If the intent is to guard against a DB where the migration has not yet run, the correct approach is to ensure migrations run before the service starts, not to silently swallow the redundancy.

**Current code:**
```ts
await db.transaction(async (trx) => {
  await trx.raw('SET CONSTRAINTS stages_user_id_position_unique DEFERRED');
  for (let i = 0; i < stageIds.length; i++) {
    ...
  }
});
```

**Improved code:**
```ts
await db.transaction(async (trx) => {
  for (let i = 0; i < stageIds.length; i++) {
    await trx('stages')
      .where({ id: stageIds[i], user_id: userId })
      .update({ position: i });
  }
});
```

---

### Summary

| Severity | Count | Description |
|----------|-------|-------------|
| Medium | 1 | Multi-statement `knex.raw()` should be split into separate `await` calls in both `up()` and `down()` |
| Low | 1 | `DROP CONSTRAINT` in `down()` (and `up()`) should use `IF EXISTS` for idempotent rollback |
| Low | 1 | `SET CONSTRAINTS ... DEFERRED` in `reorderStages` is redundant — constraint is already `INITIALLY DEFERRED` |

The core fix (deferrable constraint + transaction) is correct and appropriate for this bug. These are robustness and correctness improvements to the migration and service layer.

PR #94, opened 2026-05-12
feat: persistent real-time notification center

tomerp20:
## Summary
- Adds a fully persistent notification center with backend storage, REST API, PostgreSQL NOTIFY trigger, and real-time SSE delivery
- Builds the UI: bell icon with unread badge, sliding panel with scrollable list, per-item and mark-all-read actions
- Extends the existing SSE infrastructure (pgSubscriber) with a second channel (`notification_events`) without breaking card events

## Changes
- **Migration 011**: `notifications` table + INSERT trigger that fires `pg_notify('notification_events', ...)`
- **Backend service**: `notificationService` with `create`, `list`, `markRead`, `markAllRead`
- **Backend routes**: `GET /api/notifications`, `PATCH /api/notifications/:id/read`, `PATCH /api/notifications/read-all`
- **pgSubscriber**: listens on both `card_events` and `notification_events`; uses named SSE events (`event: message` / `event: notification`) so frontend consumers can selectively subscribe
- **Frontend service**: `notificationService` wrapping the shared axios instance
- **`useNotifications` hook**: initial fetch + SSE real-time updates + optimistic `markRead`/`markAllRead`
- **UI components**: `NotificationBell`, `NotificationPanel`, `NotificationItem` — single hook instance at Bell level, state passed down as props
- **Navbar**: `NotificationBell` inserted between nav links and user menu

## Test plan
- [ ] Run migration — `notifications` table created, trigger installed
- [ ] POST a notification directly in SQL, verify SSE fires and badge increments in browser
- [ ] Click notification item — turns gray, border disappears
- [ ] "Mark all as read" — all items go muted, badge disappears, button disabled
- [ ] Click bell again — panel closes; click again — panel reopens with same state
- [ ] Empty state shows bell icon + "No notifications"
- [ ] Board card events still work (create/move/delete) — no regression

### DIFF
--- a/backend/migrations/011_create_notifications.ts
+++ b/backend/migrations/011_create_notifications.ts
@@ -1,0 +1,45 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.createTable('notifications', (table) => {
+    table.uuid('id').primary().defaultTo(knex.raw('gen_random_uuid()'));
+    table.uuid('user_id').notNullable().references('id').inTable('users').onDelete('CASCADE');
+    table.string('title', 255).notNullable();
+    table.text('body').notNullable();
+    table.jsonb('metadata');
+    table.timestamp('read_at', { useTz: true });
+    table.timestamp('created_at', { useTz: true }).notNullable().defaultTo(knex.fn.now());
+  });
+
+  await knex.raw(`
+    CREATE OR REPLACE FUNCTION notify_notification_event()
+    RETURNS trigger AS $$
+    DECLARE
+      payload TEXT;
+    BEGIN
+      payload := json_build_object(
+        'user_id',         NEW.user_id,
+        'notification_id', NEW.id
+      )::TEXT;
+
+      PERFORM pg_notify('notification_events', payload);
+
+      RETURN NEW;
+    END;
+    $$ LANGUAGE plpgsql;
+  `);
+
+  await knex.raw(`DROP TRIGGER IF EXISTS notification_notify_trigger ON notifications;`);
+
+  await knex.raw(`
+    CREATE TRIGGER notification_notify_trigger
+    AFTER INSERT ON notifications
+    FOR EACH ROW EXECUTE FUNCTION notify_notification_event();
+  `);
+}
+
+export async function down(knex: Knex): Promise<void> {
+  await knex.raw(`DROP TRIGGER IF EXISTS notification_notify_trigger ON notifications;`);
+  await knex.raw(`DROP FUNCTION IF EXISTS notify_notification_event() CASCADE;`);
+  await knex.schema.dropTableIfExists('notifications');
+}
--- a/backend/src/routes/events.ts
+++ b/backend/src/routes/events.ts
@@ -81,7 +81,7 @@ router.get('/', async (req: Request, res: Response, next: NextFunction): Promise
   };
 
   // Send initial connected event
-  send('data: {"event":"connected"}\n\n');
+  send('event: message\ndata: {"event":"connected"}\n\n');
 
   pgSubscriber.registerClient(userId, res);
 
--- a/backend/src/routes/index.ts
+++ b/backend/src/routes/index.ts
@@ -8,6 +8,7 @@ import remindersRouter from './reminders';
 import todosRouter from './todos';
 import autocompleteRouter from './autocomplete';
 import eventsRouter from './events';
+import notificationsRouter from './notifications';
 
 const router = Router();
 
@@ -25,5 +26,6 @@ router.use('/reminders', remindersRouter);
 router.use('/todos', todosRouter);
 router.use('/autocomplete', autocompleteRouter);
 router.use('/events', eventsRouter);
+router.use('/notifications', notificationsRouter);
 
 export default router;
--- a/backend/src/routes/notifications.ts
+++ b/backend/src/routes/notifications.ts
@@ -1,0 +1,54 @@
+import { Router, Request, Response, NextFunction } from 'express';
+import { param } from 'express-validator';
+import { authenticate } from '../middleware/auth';
+import { validate } from '../middleware/validate';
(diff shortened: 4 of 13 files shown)


tomerp20 (commented):
## Code Review — Notification Center (PR #94)

Overall this is a solid, well-structured implementation that follows existing project patterns. The architecture decisions (single hook instance at Bell level, props drilling to Panel, pgSubscriber channel branching) are all correct. Several issues need addressing before merge.

---

### 🔴 Bug: NotificationPanel closes then immediately re-opens when clicking the bell

`NotificationPanel` registers a `mousedown` outside-click handler. When the user clicks the bell button a second time to **close** the panel, the sequence is:
1. `mousedown` on the bell fires → handler sees click is outside `panelRef` → calls `onClose()` → `open` set to `false`
2. `click` on the bell fires → `setOpen(prev => !prev)` → `!false = true` → panel re-opens

The panel never actually closes when clicking the bell. Fix: the click-outside handler in `NotificationPanel` must exclude the bell button trigger, or the bell button must use `mousedown` instead of `click` to toggle.

---

### 🔴 Bug: `markRead` optimistic update never rolls back on API failure

In `useNotifications.ts`, `markRead` optimistically sets `readAt` before the API call. If `notificationService.markRead()` throws, the error is swallowed with `console.error` and the notification is permanently shown as read in the current session (even though the server still has it as unread). The next page load will correct the state, but the current session is silently broken.

```ts
// useNotifications.ts — markRead should revert on failure
const markRead = useCallback(async (id: string) => {
  setNotifications((prev) =>
    prev.map((n) => (n.id === id ? { ...n, readAt: new Date().toISOString() } : n))
  );
  try {
    await notificationService.markRead(id);
  } catch (err) {
    // BUG: state is never reverted here
    console.error('[useNotifications] Failed to mark read:', err);
  }
}, []);
```

Same issue exists in `markAllRead`.

---

### 🔴 Accessibility: `NotificationItem` uses a `div` for an interactive element

Unread notification items are clickable (they call `onRead`) but are rendered as a `<div>`. This breaks keyboard navigation — the element is not focusable and not reachable via Tab. It should be a `<button>` or have `role="button" tabIndex={0}` with a `onKeyDown` handler.

---

### 🟡 Silent no-op on `markRead` for unknown/wrong-user notification ID

`notificationService.markRead` does a `WHERE id = ... AND user_id = ...` update with no existence check. If the ID doesn't exist or belongs to a different user, the update silently affects 0 rows and the route returns 204. Compare with `todoService.updateTodo` which throws `AppError('Todo not found', 404)`. The route should verify ownership and return 404 when the notification is not found.

---

### 🟡 Dead code: `JSON.parse` result discarded in `useNotifications`

```ts
// useNotifications.ts
try {
  JSON.parse(event.data as string);  // result never used
} catch (err) {
  console.error('[useNotifications] Failed to parse SSE payload:', err);
  return;
}
```

The parse result is thrown away immediately. Either use it (e.g. validate the `notification_id` field is present) or remove the parse entirely — the early-return guard is the only purpose, but invalid JSON from a trusted internal channel is not a realistic failure mode worth guarding. If keeping it for defensiveness, assign to `_` to signal intent.

---

### 🟡 SSE format inconsistency: `connected` event still uses unnamed format

`events.ts` sends the initial `connected` event as `data: {"event":"connected"}\n\n` (no `event:` field). All subsequent card/notification events now use `event: <name>\ndata: ...\n\n`. This inconsistency means `useCardEvents` relies on two different SSE formats from the same stream. The `connected` event should also use the named format for consistency: `event: message\ndata: {"event":"connected"}\n\n`.

---

### ℹ️ Minor: Two SSE connections open simultaneously on the Board page

`NotificationBell` (in Navbar) and `BoardPage` (via `useCardEvents`) each open their own `EventSource` to `/api/events`. On the Board page there are two concurrent connections. This is not a bug — it's architecturally valid and matches the spec — but it's worth noting as future tech debt if a shared SSE context is ever desired.

tomerp20 (commented):
## Code Review — Notification Center (PR #94)

Overall this is a solid, well-structured implementation. The architecture decisions (single hook instance at Bell level, props drilling to Panel, pgSubscriber channel branching) are correct. Several issues need addressing before merge.

### Bug: NotificationPanel closes then immediately re-opens when clicking the bell

NotificationPanel registers a mousedown outside-click handler. When the user clicks the bell button a second time to close the panel, the sequence is: (1) mousedown on bell fires -> handler sees click is outside panelRef -> calls onClose() -> open set to false, (2) click on bell fires -> setOpen(prev => !prev) -> !false = true -> panel re-opens. The panel never actually closes when clicking the bell. Fix: exclude the bell button trigger from the outside-click handler, or use mousedown on the bell instead of click.

### Bug: markRead optimistic update never rolls back on API failure

In useNotifications.ts, markRead optimistically sets readAt before the API call. If notificationService.markRead() throws, the error is swallowed and the notification permanently shows as read in the current session. The same issue exists in markAllRead. The catch block should revert the optimistic state update.

### Accessibility: NotificationItem uses a div for an interactive element

Unread notification items are clickable (they call onRead) but rendered as a div. This breaks keyboard navigation — the element is not focusable via Tab. It should be a button, or have role='button' tabIndex={0} with an onKeyDown handler.

### Silent no-op on markRead for unknown notification ID

notificationService.markRead does WHERE id = ... AND user_id = ... with no existence check. If the ID doesn't exist or belongs to another user, it silently affects 0 rows and returns 204. Compare with todoService.updateTodo which throws AppError('Todo not found', 404). The route should return 404 when the notification is not found.

### Dead code: JSON.parse result discarded in useNotifications

The parse result is thrown away immediately — it only serves as a guard for invalid JSON. Either use the result to validate the notification_id field, or assign to _ to signal intent.

### SSE format inconsistency: connected event still uses unnamed format

events.ts sends the initial connected event as data: {event:connected} (no event: field). All subsequent card/notification events now use event: message or event: notification. The connected event should also use the named message format for consistency.

PR #95, opened 2026-05-12
feat: Settings page with Gmail integration scaffold

### LINKED ISSUE #85: [Settings Page] Scaffold /settings route with Gmail section UI
## Parent PRD

#81

## What to build

Create the `/settings` page and wire it into the navbar. The page contains only the Gmail integration section in v1. All three visual states (disconnected, connected, error) are fully built and styled — but buttons are stubbed (no OAuth wiring yet, that comes in #87). The page must be navigable and visually complete.

See PR 2 / Phase 4 of the parent PRD.

## Acceptance criteria

- [ ] `/settings` route exists and is protected by auth
- [ ] "Settings" link added above "Logout" in the navbar user dropdown
- [ ] Gmail section renders the **disconnected state** by default: description text + "Connect Gmail" button (stubbed — no action yet)
- [ ] Gmail section renders the **connected state** (can be toggled via mock/prop for review): shows gmail address, "Last synced: Never synced yet", "Sync Now" button (stubbed), "Disconnect" button (stubbed)
- [ ] Gmail section renders the **error state**: warning banner "Gmail connection lost — please reconnect" + "Reconnect" button (stubbed)
- [ ] All states match the styling conventions of the existing codebase (Tailwind)
- [ ] Page is responsive and consistent with the rest of the app

## Blocked by

- Blocked by #84

## User stories addressed

- User story 1: connect Gmail account
- User story 5: manually trigger a sync
- User story 6: see when inbox was last scanned
- User story 7: alerted if Gmail connection breaks
- User story 8: disconnect Gmail at any time

tomerp20:
## Summary
- Adds a `/settings` protected route with a Gmail integration section
- Three visual states: Disconnected, Connected (valid), and Connection lost (token revoked)
- Gmail status is fetched from `GET /api/gmail/status`; any 404/network error gracefully defaults to disconnected state

## Changes
- `frontend/src/pages/SettingsPage.tsx` — new page with three Gmail integration states driven by API status
- `frontend/src/services/api.ts` — adds `gmailApi.getStatus()` with typed return union
- `frontend/src/components/Layout/Navbar.tsx` — adds "Settings" link above "Sign out" in user dropdown
- `frontend/src/App.tsx` — registers `/settings` inside the protected `AppLayout` route

## Test plan
- [ ] Visit `/settings` while logged in — Gmail section shows "Disconnected" state (API not yet implemented, falls back gracefully)
- [ ] Clicking "Connect Gmail" does nothing (stub, no errors)
- [ ] Visit `/settings` while logged out — should redirect to `/login`
- [ ] Open user dropdown in navbar — "Settings" link appears above "Sign out"; clicking it navigates to `/settings` and closes dropdown
- [ ] Build passes: `npm run build` in `frontend/`

Closes #85

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -6,6 +6,7 @@ import SignupPage from '@/pages/SignupPage';
 import BoardPage from '@/pages/BoardPage';
 import DashboardPage from '@/pages/DashboardPage';
 import TasksPage from '@/pages/TasksPage';
+import SettingsPage from '@/pages/SettingsPage';
 
 export default function App() {
   return (
@@ -24,6 +25,7 @@ export default function App() {
             <Route path="/" element={<BoardPage />} />
             <Route path="/dashboard" element={<DashboardPage />} />
             <Route path="/tasks" element={<TasksPage />} />
+            <Route path="/settings" element={<SettingsPage />} />
           </Route>
         </Routes>
       </AuthProvider>
--- a/frontend/src/components/Layout/Navbar.tsx
+++ b/frontend/src/components/Layout/Navbar.tsx
@@ -1,6 +1,6 @@
 import { Link, useLocation } from 'react-router-dom';
 import { useAuth } from '@/hooks/useAuth';
-import { Briefcase, LayoutDashboard, Columns3, CheckSquare, LogOut, User } from 'lucide-react';
+import { Briefcase, LayoutDashboard, Columns3, CheckSquare, LogOut, User, Settings } from 'lucide-react';
 import { useState, useRef, useEffect } from 'react';
 import NotificationBell from './NotificationBell';
 
@@ -75,6 +75,14 @@ export default function Navbar() {
                 <p className="text-sm font-medium text-gray-900">{user?.name}</p>
                 <p className="text-xs text-gray-500 truncate">{user?.email}</p>
               </div>
+              <Link
+                to="/settings"
+                onClick={() => setShowMenu(false)}
+                className="flex w-full items-center gap-2 px-3 py-2 text-sm text-gray-600 hover:bg-gray-50"
+              >
+                <Settings size={14} />
+                Settings
+              </Link>
               <button
                 onClick={() => {
                   setShowMenu(false);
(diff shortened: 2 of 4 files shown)


tomerp20 (commented):
## Code Review — PR #95: Settings page with Gmail integration scaffold

### Overview
This PR adds a `/settings` protected route with a Gmail integration section showing three distinct visual states (disconnected, connected+valid, connected+token-revoked). The overall structure is solid and consistent with the existing codebase conventions. A few issues need addressing before this is ready.

---

### Issues Found

**1. Dead state — `syncResult` is declared with `useState` but has no setter exposed (`[syncResult]` only)**

`frontend/src/pages/SettingsPage.tsx` line 13:
```ts
const [syncResult] = useState('');
```
This state can never be updated — the setter is destructured away. The `{syncResult && ...}` render block below it will therefore never show anything, even in Round 3 when sync is wired. Since this is a placeholder for future use, it should either be removed entirely (the empty string renders nothing anyway, so the conditional block is also dead code) or kept with the setter to make it ready for Round 3.

**2. Redundant type union in `gmailApi` — the return type is declared twice, verbatim**

`frontend/src/services/api.ts` lines 229–239:
The `Promise<...>` return type annotation and the `as ...` cast in the `return` statement are identical. The cast does not add safety — TypeScript already verified the shape at the return site from the `Promise<>` annotation. The duplication increases maintenance surface (both would need updating in Round 3). Extract the union into a named exported type and reference it in both places.

**3. No loading state while the Gmail status is being fetched**

`frontend/src/pages/SettingsPage.tsx` lines 11–20:
The component has no loading indicator. On initial render, the page immediately shows the "Disconnected" state and then potentially flips to "Connected" once the API responds. This causes a layout flash. Other pages (`TasksPage`, `DashboardPage`) use a skeleton/loading guard. A simple `useState<boolean>` loading flag should be added and the card content replaced with a skeleton `div` while the fetch is in flight.

**4. `isValid` property access is unsafe when `connected` is `false`**

`frontend/src/pages/SettingsPage.tsx` lines 43 and 106:
```tsx
{gmailStatus.connected && !gmailStatus.isValid && (
```
When `connected` is `false`, the `GmailStatus` discriminated union for that variant (`{ connected: false }`) does not have an `isValid` property. TypeScript's control-flow narrowing handles this correctly *at compile time* (the short-circuit on `connected` prevents the right side from evaluating), but the pattern is fragile — a reader unaware of the union shape may not immediately see why `isValid` is safe here. More importantly, the `gmailStatus.isValid` access in a `!gmailStatus.connected` branch is evaluated by TypeScript only because of short-circuit; if someone ever refactors these conditions, it becomes a runtime `undefined` read. The idiomatic fix is to use a helper or inline type narrowing: check `gmailStatus.connected === true && gmailStatus.isValid === false` explicitly.

PR #97, opened 2026-05-12
fix: switch default LLM provider to Google Gemini

tomerp20:
## Summary
- Adds `@ai-sdk/google` as a dependency to support Google Generative AI models
- Changes the default LLM provider from Anthropic to Google (`gemini-2.5-flash-lite`)
- Anthropic (`claude-3-haiku-20240307`) remains selectable via `LLM_PROVIDER=anthropic`

## Changes
- `backend/src/services/emailClassifier.ts`: import and instantiate `createGoogleGenerativeAI`, update `getModel()` to default to `google`, add `GOOGLE_AI_API_KEY` env var usage
- `backend/package.json` + `backend/package-lock.json`: add `@ai-sdk/google` dependency

## Test plan
- [ ] Set `GOOGLE_AI_API_KEY` in the backend environment
- [ ] Verify email classification works end-to-end with the Google provider (default, no `LLM_PROVIDER` set)
- [ ] Set `LLM_PROVIDER=anthropic` and verify Anthropic classification still works
- [ ] Verify `LLM_PROVIDER=google` explicitly also routes to Google

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -15,3 +15,9 @@ CORS_ORIGIN=http://localhost:5173
 
 # Logging
 LOG_LEVEL=debug
+
+# LLM (email classifier)
+# Provider to use for email classification. Options: 'google' (default), 'anthropic'
+LLM_PROVIDER=google
+GOOGLE_AI_API_KEY=your-google-ai-api-key
+ANTHROPIC_API_KEY=your-anthropic-api-key
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -9,6 +9,7 @@
       "version": "1.0.0",
       "dependencies": {
         "@ai-sdk/anthropic": "^1.0.0",
+        "@ai-sdk/google": "^3.0.72",
         "ai": "^4.0.0",
         "bcryptjs": "^2.4.3",
         "compression": "^1.8.1",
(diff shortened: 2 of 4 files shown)


tomerp20 (commented):
## Code Review — PR #97: Switch default LLM provider to Google Gemini

### Overview
This PR adds `@ai-sdk/google` as a dependency and switches the default LLM provider in `emailClassifier.ts` from Anthropic (`claude-3-haiku`) to Google (`gemini-2.5-flash-lite`), while preserving Anthropic as a selectable fallback via `LLM_PROVIDER=anthropic`.

---

### Issues Found

#### 1. `GOOGLE_AI_API_KEY` is missing from `.env.example` (medium)

`backend/.env.example` documents zero AI-related env vars. After this PR, the service silently fails at runtime if `GOOGLE_AI_API_KEY` is not set — the Google SDK will throw an unhelpful error with no guidance for new developers or deployers.

`ANTHROPIC_API_KEY` and the new `LLM_PROVIDER` / `GOOGLE_AI_API_KEY` variables should all be documented in `.env.example`.

#### 2. No startup-time validation of the API key (low)

Both providers are instantiated at module load time with `process.env.GOOGLE_AI_API_KEY` / `process.env.ANTHROPIC_API_KEY`. If the relevant key is `undefined`, the first classification call will fail with an SDK-level error rather than a clear startup warning. A guard (or at least a `console.warn` at module init) would make misconfiguration obvious earlier.

#### 3. `package.json` pins `@ai-sdk/google` to an exact minor (`^3.0.72`) while `@ai-sdk/anthropic` uses a looser `^1.0.0` (minor)

This is inconsistent. The `^` prefix on `^3.0.72` already allows patch updates, so it is not a hard pin, but using `^3.0.0` would be consistent with the existing convention and allow patch + minor upgrades as the project does for other packages.

#### 4. Deleted `// Instantiate the provider once at module level rather than on every call` comment (trivial)

The comment was useful documentation explaining why providers are module-level singletons. It was removed rather than updated to cover both providers.

---

### What looks good
- `gemini-2.5-flash-lite` is a valid model ID in `@ai-sdk/google@3.0.72` (confirmed in SDK type definitions).
- The new Google SDK packages introduce **no new vulnerabilities** — all pre-existing `npm audit` findings (`handlebars`, `lodash`, etc.) were already present before this PR.
- The zod peer dependency (`^3.25.76`) is satisfied by the installed `zod@3.25.76`.
- The `sanitizeForPrompt` helper with its stronger multi-pattern sanitization was correctly preserved rather than replaced with the simpler version from the task spec.
- Provider isolation (module-level singletons) is maintained for both providers.

tomerp20 (commented):
Review posted. Now fixing the issues found.

PR #98, opened 2026-05-12
fix: switch LLM to Google Gemini and fix multipart email body extraction

tomerp20:
## Summary
- Switches LLM provider from Anthropic to Google Gemini (`gemini-2.5-flash-lite`) via `@ai-sdk/google`
- Adds `LLM_PROVIDER` env var for zero-code provider switching
- Fixes recursive multipart email body extraction so rejection emails are correctly classified

## Changes
- `emailClassifier.ts`: replaced Anthropic with Google Generative AI provider, added `LLM_PROVIDER` switch
- `gmailService.ts`: replaced flat `parts.find()` body extraction with recursive `extractTextBody()` that handles nested MIME structure (`multipart/mixed > multipart/alternative > text/plain`)
- `package.json`: added `@ai-sdk/google ^1.0.0`, downgraded to `^1.0.0` for `ai@4.x` compatibility
- Committed updated `package-lock.json`

## Root Cause
Rejection emails from Grip Security and Hunters were classified as `not_rejection` because the body extraction returned empty string for nested multipart MIME structures. The LLM only saw the subject line, which lacks clear rejection language.

## Test plan
- [ ] Verify `processed_emails` is empty and `last_sync_at` is NULL (already reset)
- [ ] Run sync — 2 rejection emails should now be classified as `rejection` with high confidence
- [ ] Confirm cards for those companies are moved to Rejected stage

### DIFF
--- a/backend/src/services/emailClassifier.ts
+++ b/backend/src/services/emailClassifier.ts
@@ -22,17 +22,24 @@ if (activeProvider === 'anthropic' && !process.env.ANTHROPIC_API_KEY) {
   console.warn('[emailClassifier] ANTHROPIC_API_KEY is not set — email classification will fail');
 }
 
-const anthropicProvider = createAnthropic({ apiKey: process.env.ANTHROPIC_API_KEY });
-const googleProvider = createGoogleGenerativeAI({ apiKey: process.env.GOOGLE_AI_API_KEY });
+// Only instantiate the active provider's client — the inactive provider's API
+// key may be absent, and some SDKs validate at construction time.
+const anthropicProvider = activeProvider === 'anthropic'
+  ? createAnthropic({ apiKey: process.env.ANTHROPIC_API_KEY })
+  : null;
+const googleProvider = activeProvider === 'google'
+  ? createGoogleGenerativeAI({ apiKey: process.env.GOOGLE_AI_API_KEY })
+  : null;
 
 function getModel() {
-  const provider = process.env.LLM_PROVIDER ?? 'google';
-  switch (provider) {
+  // Use the module-level constant — avoids re-reading env per call and keeps
+  // getModel() consistent with the startup warning above.
+  switch (activeProvider) {
     case 'anthropic':
-      return anthropicProvider('claude-3-haiku-20240307');
+      return anthropicProvider!('claude-3-haiku-20240307');
     case 'google':
     default:
-      return googleProvider('gemini-2.5-flash-lite');
+      return googleProvider!('gemini-2.5-flash-lite');
   }
 }
 
--- a/backend/src/services/gmailService.ts
+++ b/backend/src/services/gmailService.ts
@@ -43,6 +43,38 @@ function verifyOAuthStateToken(state: string): string {
   }
 }
 
+type GmailPart = { mimeType?: string | null; body?: { data?: string | null } | null; parts?: GmailPart[] | null };
+
+function extractTextBody(payload: GmailPart | null | undefined): string {
+  if (!payload) return '';
+  if (payload.mimeType === 'text/plain' && payload.body?.data) {
+    return Buffer.from(payload.body.data, 'base64').toString('utf-8');
+  }
+  if (payload.parts) {
+    // Prefer text/plain at any depth
+    for (const part of payload.parts) {
+      if (part.mimeType === 'text/plain' && part.body?.data) {
+        return Buffer.from(part.body.data, 'base64').toString('utf-8');
+      }
+    }
+    // Recurse into nested multipart containers
+    for (const part of payload.parts) {
+      if (part.mimeType?.startsWith('multipart/')) {
+        const result = extractTextBody(part);
+        if (result) return result;
+      }
+    }
+    // Last resort: strip HTML tags from text/html part
+    for (const part of payload.parts) {
+      if (part.mimeType === 'text/html' && part.body?.data) {
+        const html = Buffer.from(part.body.data, 'base64').toString('utf-8');
+        return html.replace(/<[^>]+>/g, ' ').replace(/\s+/g, ' ').trim();
+      }
+    }
+  }
+  return '';
+}
+
 export const gmailService = {
   getAuthUrl(userId: string): string {
     const oauth2Client = createOAuthClient();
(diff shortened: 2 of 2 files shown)


tomerp20 (commented):
## Code Review — PR #98: Switch LLM to Google Gemini + fix multipart email body extraction

### Files reviewed
- `backend/src/services/gmailService.ts` — `extractTextBody()` recursive MIME traversal
- `backend/src/services/emailClassifier.ts` — Google Generative AI provider wiring
- `backend/package.json` — `@ai-sdk/google` dependency addition

---

### CRITICAL — None found

---

### HIGH — Unbounded recursion depth in `extractTextBody` (stack overflow risk)

**File:** `backend/src/services/gmailService.ts`

`extractTextBody` recurses into every `multipart/*` child with no depth cap. RFC 2046 places no practical limit on nesting depth, so a crafted email with dozens of nested `multipart/mixed` wrappers will exhaust V8's call stack and crash the sync worker process, dropping all emails in the current batch.

Real-world legitimate emails are at most 3–4 levels deep, so a cap of 10 is conservative and safe.

**Current code:**
```typescript
function extractTextBody(payload: GmailPart | null | undefined): string {
  if (!payload) return '';
  if (payload.mimeType === 'text/plain' && payload.body?.data) {
    return Buffer.from(payload.body.data, 'base64').toString('utf-8');
  }
  if (payload.parts) {
    for (const part of payload.parts) {
      if (part.mimeType === 'text/plain' && part.body?.data) {
        return Buffer.from(part.body.data, 'base64').toString('utf-8');
      }
    }
    for (const part of payload.parts) {
      if (part.mimeType?.startsWith('multipart/')) {
        const result = extractTextBody(part);   // <-- no depth guard
        if (result) return result;
      }
    }
    ...
  }
  return '';
}
```

**Fixed code:**
```typescript
function extractTextBody(payload: GmailPart | null | undefined, depth = 0): string {
  if (!payload || depth > 10) return '';   // guard against adversarial nesting
  if (payload.mimeType === 'text/plain' && payload.body?.data) {
    return Buffer.from(payload.body.data, 'base64').toString('utf-8');
  }
  if (payload.parts) {
    for (const part of payload.parts) {
      if (part.mimeType === 'text/plain' && part.body?.data) {
        return Buffer.from(part.body.data, 'base64').toString('utf-8');
      }
    }
    for (const part of payload.parts) {
      if (part.mimeType?.startsWith('multipart/')) {
        const result = extractTextBody(part, depth + 1);
        if (result) return result;
      }
    }
    for (const part of payload.parts) {
      if (part.mimeType === 'text/html' && part.body?.data) {
        const html = Buffer.from(part.body.data, 'base64').toString('utf-8');
        return html.replace(/<[^>]+>/g, ' ').replace(/\s+/g, ' ').trim();
      }
    }
  }
  return '';
}
```

---

### MEDIUM — `text/html` fallback does not recurse into nested multipart containers

**File:** `backend/src/services/gmailService.ts`

The last-resort HTML fallback only scans immediate children of `payload.parts` for `text/html`. If the email structure is `multipart/mixed > multipart/alternative > text/html` (with no `text/plain` anywhere), the fallback silently returns `''` and the LLM sees an empty body — the same root-cause bug this PR is fixing, just for the HTML path.

The fix is to also recurse into `multipart/*` children when looking for `text/html`, which the depth guard (added for the HIGH issue above) makes safe.

**Current code:**
```typescript
// Last resort: strip HTML tags from text/html part
for (const part of payload.parts) {
  if (part.mimeType === 'text/html' && part.body?.data) {
    const html = Buffer.from(part.body.data, 'base64').toString('utf-8');
    return html.replace(/<[^>]+>/g, ' ').replace(/\s+/g, ' ').trim();
  }
}
```

**Fixed code:** The HTML fallback should participate in the same recursive call already made for `multipart/*` parts. Rather than a separate loop, the cleaner fix is to let the recursive call on the `multipart/*` children fall through to the HTML fallback naturally — which the depth-guarded signature already enables. No separate code change needed beyond the depth guard fix above; the existing structure handles it once depth is tracked.

(Alternatively, add a recursive HTML search loop analogous to the recursive `text/plain` search, but that duplicates logic. The simplest solution: after the recursive multipart loop, search nested parts for `text/html` by also recursing — this falls out naturally from the depth-guarded version above.)

---

### MEDIUM — `getModel()` re-reads `process.env.LLM_PROVIDER` independently from `activeProvider`

**File:** `backend/src/services/emailClassifier.ts`

`activeProvider` is captured once at module load for startup warnings, but `getModel()` re-reads `process.env.LLM_PROVIDER` on every invocation. In test environments where env vars are mutated between calls (or if the variable is deleted), the warning and the actual model selected by `getModel()` will disagree — making misconfiguration hard to diagnose.

**Current code:**
```typescript
const activeProvider = process.env.LLM_PROVIDER ?? 'google';
// ... warnings use activeProvider ...

function getModel() {
  const provider = process.env.LLM_PROVIDER ?? 'google';   // <-- independent read
  switch (provider) {
    case 'anthropic':
      return anthropicProvider('claude-3-haiku-20240307');
    case 'google':
    default:
      return googleProvider('gemini-2.5-flash-lite');
  }
}
```

**Fixed code:**
```typescript
const activeProvider = process.env.LLM_PROVIDER ?? 'google';
// ... warnings use activeProvider ...

function getModel() {
  switch (activeProvider) {   // use the already-captured constant
    case 'anthropic':
      return anthropicProvider('claude-3-haiku-20240307');
    case 'google':
    default:
      return googleProvider('gemini-2.5-flash-lite');
  }
}
```

---

### LOW — Both provider clients unconditionally instantiated at module load

**File:** `backend/src/services/emailClassifier.ts`

`createAnthropic(...)` and `createGoogleGenerativeAI(...)` are both called at module load regardless of `LLM_PROVIDER`. The inactive provider's client is created with a potentially undefined API key. While the inactive client is never called, some SDKs validate or throw at construction time, and it creates a misleading impression that both providers are always required.

**Current code:**
```typescript
const anthropicProvider = createAnthropic({ apiKey: process.env.ANTHROPIC_API_KEY });
const googleProvider = createGoogleGenerativeAI({ apiKey: process.env.GOOGLE_AI_API_KEY });
```

**Fixed code:** Instantiate lazily inside `getModel()`, or only instantiate the active provider:
```typescript
const anthropicProvider = activeProvider === 'anthropic'
  ? createAnthropic({ apiKey: process.env.ANTHROPIC_API_KEY })
  : null;
const googleProvider = activeProvider === 'google'
  ? createGoogleGenerativeAI({ apiKey: process.env.GOOGLE_AI_API_KEY })
  : null;

function getModel() {
  switch (activeProvider) {
    case 'anthropic':
      return anthropicProvider!('claude-3-haiku-20240307');
    case 'google':
    default:
      return googleProvider!('gemini-2.5-flash-lite');
  }
}
```

---

### Summary

| Severity | Count | Items |
|----------|-------|-------|
| Critical | 0 | — |
| High | 1 | Unbounded recursion depth → stack overflow risk on adversarial MIME input |
| Medium | 2 | HTML fallback non-recursive; `getModel()` env-var re-read inconsistency |
| Low | 1 | Both provider clients instantiated unconditionally |

The HIGH issue (no recursion depth cap) is the one actionable fix — the others are defensive improvements. Fix commit to follow.

PR #100, opened 2026-05-14
refactor: client-side board search with useDeferredValue and React.memo

tomerp20:
## Summary
- Moves board search filtering from server-side (API call per 300ms debounce) to client-side using React 18's `useDeferredValue` + `useMemo`
- Adds `React.memo` to `Board` and `SortableCard`, with `useCallback` on all handlers passed as props, so typing in the search box no longer triggers any re-renders in the card tree until the deferred value settles
- Fixes inline arrow function anti-pattern on `SortableCard` that was defeating memo

## Changes
- `BoardPage.tsx` — separate `searchQuery` state from server filters; add `useDeferredValue`, `filteredCards` memo, `useCallback` on all Board handlers
- `SearchBar.tsx` — remove debounce effect, become controlled via `searchQuery`/`onSearchChange` props
- `Board.tsx` — `memo(Board)`
- `SortableCard.tsx` — `memo(SortableCard)`, prop renamed from `onClick` to `onCardClick(id)`
- `Column.tsx` — pass `onCardClick` directly instead of inline arrow
- `CLAUDE.md` — new branch-naming rule for all change types

## Test plan
- [ ] Typing in search filters cards client-side with no network requests
- [ ] Stage / priority / work mode filters still trigger server fetch
- [ ] Clicking a card still opens the detail modal
- [ ] Drag and drop still works
- [ ] Clear all resets both search and other filters

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,189 @@
+# Claude Instructions for JobFlow
+
+## Branch Rule — Applies to ALL Code Changes
+
+**Every code change — feature, bug fix, refactor, experiment, or anything else — must live on a dedicated branch. No exceptions. Never commit directly to `main`.**
+
+If Claude does not know the purpose of the change, it must ask the user before writing any code:
+> "Before I start, what is the goal of this change? (new feature, bug fix, refactor, learning experiment, etc.) — I need this to name the branch correctly."
+
+**Branch naming convention by type:**
+
+| Purpose | Prefix | Example |
+|---------|--------|---------|
+| New feature | `feature/` | `feature/drag-and-drop-cards` |
+| Bug fix | `fix/` | `fix/search-not-clearing-filters` |
+| Refactor / cleanup | `refactor/` | `refactor/client-side-search` |
+| Learning / experiment | `experiment/` | `experiment/use-deferred-value` |
+| Chore / config / deps | `chore/` | `chore/upgrade-react-19` |
+| Docs | `docs/` | `docs/api-reference` |
+
+Branch off the latest `main` unless the user specifies otherwise:
+```bash
+git checkout main
+git pull origin main
+git checkout -b <prefix>/<short-kebab-case-description>
+```
+
+---
+
+## Development Workflow
+
+Every code change must follow this exact workflow — no exceptions.
+
+---
+
+### Step 1: Create a Branch
+
+Before writing any code, determine the purpose and create the appropriate branch (see Branch Rule above). If the purpose is unclear, ask first.
+
+---
+
+### Step 2: Implement the Change
+
+Make all code changes on the branch. Keep changes focused — only modify what is necessary for the stated purpose. Do not refactor unrelated code, add unnecessary comments, or over-engineer.
+
+**External dependencies must be verified before use.** If a plan or task requires calling an external API, loading from a third-party URL, or depending on any service outside the codebase, Claude must confirm it is reachable and working before writing code that depends on it. If research agents report that a dependency is unreachable, broken, or returns errors — stop, surface this to the user, and agree on an alternative before proceeding. Never implement code that relies on a dependency known to be broken.
+
+---
+
+### Step 3: Commit the Changes
+
+Stage and commit with a clear, structured commit message.
+
+**Commit message format:**
+```
+<type>: <concise summary in imperative mood>
+
+<body — what changed and why, not how>
+- bullet points for multiple changes
+- reference any relevant context
+
+Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>
+```
+
+**Types:** `feat`, `fix`, `refactor`, `style`, `docs`, `test`, `chore`
+
+Example:
+```
+feat: add drag-and-drop reordering for kanban cards
+
+- Cards can now be dragged between columns
+- Order persists to the backend via PATCH /api/cards/:id
+- Drag handle shown on hover to avoid accidental drags
+```
+
+Commands:
+```bash
+git add <specific files>
+git commit -m "$(cat <<'EOF'
+feat: your message here
+
+- detail 1
+- detail 2
+
+Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>
+EOF
+)"
+```
+
+---
+
+### Step 4: Create a Pull Request
+
+Push the branch and open a PR from the branch to `main`.
+
+```bash
+git push -u origin <branch-name>
+gh pr create --title "<concise PR title>" --body "$(cat <<'EOF'
+## Summary
+- What this PR does (1-3 bullets)
+
+## Changes
+- List of key changes
+
+## Test plan
+- [ ] Manual test steps or scenarios
+
+🤖 Generated with [Claude Code](https://claude.com/claude-code)
+EOF
+)"
+```
+
+Return the PR URL to the user so they can see it.
+
+---
+
+### Step 5: Run Code Review Agent
+
+**This step is mandatory and must run after Step 4 (PR creation) — no exceptions.**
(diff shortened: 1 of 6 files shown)


tomerp20 (commented):
## Code Review — PR #100: Client-side board search with useDeferredValue and React.memo

### Summary

Reviewed files: `BoardPage.tsx`, `Board.tsx`, `Column.tsx`, `SortableCard.tsx`, `SearchBar.tsx`, `CLAUDE.md`

The architectural direction is correct — moving search client-side eliminates a debounced API call per keystroke, and the `useDeferredValue` + `useMemo` pairing is idiomatic React 18. However, there are two correctness issues and two incomplete memoization fixes that need to be addressed before this is safe to merge.

---

### Issues Found

---

**[HIGH] — Drag-and-drop position corruption when search is active**

`BoardPage` passes `filteredCards` (the search-filtered subset) to `Board` as its `cards` prop. `Board.getCardsByStage` and `handleDragEnd` compute positions entirely from that prop. When a search is active and a column shows, say, 2 of its 5 cards, dragging a card to position 1 calls `onMoveCard(id, stageId, 1)` — but the backend and the optimistic `setCards` in `BoardPage.handleMoveCard` both operate on all 5 cards. The persisted position will be wrong and the board will reorder incorrectly after the next full fetch.

Current code (`BoardPage.tsx` line 197–208):
```tsx
<Board
  stages={stages}
  cards={filteredCards}   // <-- filtered subset passed to Board
  ...
  onMoveCard={handleMoveCard}
/>
```

`handleDragEnd` in `Board.tsx` (lines 114–125):
```tsx
const stageCards = getCardsByStage(overId);         // only filtered cards
targetPosition = stageCards.length;                 // position in filtered subset, not full column
```

Fix — pass the full `cards` to `Board` for DnD, and pass `filteredCards` separately for rendering. Board needs to distinguish between "cards for DnD logic" and "cards to display". The cleanest approach: add a separate `displayCards` prop (or rename the existing prop) so Board renders `displayCards` per column but uses the full `cards` array for position calculations.

Alternatively, disable drag-and-drop while a search is active (simpler, but a UX regression). The prop-split approach is preferred.

```tsx
// BoardPage.tsx
<Board
  stages={stages}
  cards={cards}              // full set — DnD position calculations
  displayCards={filteredCards}  // filtered set — what actually renders per column
  ...
/>
```

```tsx
// Board.tsx — add displayCards prop, use it only in getCardsByStage for rendering
interface BoardProps {
  cards: Card[];       // for DnD position logic
  displayCards: Card[]; // for column rendering
  ...
}

const getDisplayCardsByStage = useCallback(
  (stageId: string) =>
    displayCards
      .filter((c) => c.stageId === stageId)
      .sort((a, b) => a.position - b.position),
  [displayCards]
);

// Use getCardsByStage (full cards) in handleDragEnd
// Use getDisplayCardsByStage in the render and DragOverlay
```

---

**[MEDIUM] — `memo(SortableCard)` is defeated by an inline onClick wrapper**

`SortableCard` is wrapped in `memo`, which means React bails out of re-rendering only when its props are referentially stable. But the `onClick` handler is defined as an inline arrow inside the component body:

```tsx
// SortableCard.tsx lines 34–36
onClick={() => {
  if (!isDragging) onCardClick(card.id);
}}
```

This arrow is a new function object on every render. `memo` compares props shallowly — it will always see a new `onClick` and re-render regardless. The fix is to move the guard logic outside the JSX, or use `useCallback` inside `SortableCard`.

Current code:
```tsx
onClick={() => {
  if (!isDragging) onCardClick(card.id);
}}
```

Improved code:
```tsx
// Define a stable handler inside the component
const handleClick = useCallback(() => {
  if (!isDragging) onCardClick(card.id);
}, [isDragging, onCardClick, card.id]);

// Then in JSX:
onClick={handleClick}
```

Note: `isDragging` changes during a drag, so it must be in the dependency array. `card.id` is stable for a given card instance. This makes the handler stable between non-drag renders.

---

**[MEDIUM] — `onCardClick={setSelectedCardId}` is an uncontrolled reference — partially defeats `memo(Board)`**

In `BoardPage.tsx` line 201, `setSelectedCardId` (the raw `useState` setter) is passed directly as `onCardClick`. React guarantees setState setters are stable across renders, so this is actually fine for `Board` itself. However it is worth noting that the `onAddCard` prop on line 179 of `Board.tsx` wraps `onAddCard` in an inline arrow (`onAddCard={() => onAddCard(stage.id)}`), which creates a new reference every time `Board` re-renders. This is inside `Board.tsx` (unchanged code), but it means `SortableColumn` receives a new `onAddCard` on every Board render regardless. This pre-existed the PR but is now more impactful since Board is memoized and re-renders are more visible. Flagging as Medium — the fix belongs in a follow-up or can be done here by extracting a handler.

---

**[MEDIUM] — `handleCardCreated`, `handleCardUpdated`, `handleCardDeleted`, `handleStageSaved`, `confirmDeleteStage` are not wrapped in `useCallback`**

The PR wraps all Board-tree handlers in `useCallback`, which is correct. But the modal-bound handlers are not:

```tsx
// BoardPage.tsx lines 92–103, 123–131, 137–147
const handleCardCreated = (card: Card) => { ... };
const handleCardUpdated = (updatedCard: Card) => { ... };
const handleCardDeleted = (cardId: string) => { ... };
const handleStageSaved = (saved: Stage) => { ... };
const confirmDeleteStage = async () => { ... };
```

These are passed as props to `CardDetail`, `CardForm`, and `StageForm`. Those components are not currently memoized, so this is not a regression — but it is inconsistent with the pattern this PR establishes, and any future `memo()` on those modals would be silently broken. Wrap them all in `useCallback` for consistency and forward-safety.

```tsx
const handleCardCreated = useCallback((card: Card) => {
  setCards((prev) => [...prev, card]);
  setShowCreateForm(false);
}, []);

const handleCardUpdated = useCallback((updatedCard: Card) => {
  setCards((prev) => prev.map((c) => (c.id === updatedCard.id ? updatedCard : c)));
}, []);

const handleCardDeleted = useCallback((cardId: string) => {
  setCards((prev) => prev.filter((c) => c.id !== cardId));
  setSelectedCardId(null);
}, []);
```

`handleStageSaved` closes over `editingStage` so it needs `editingStage` in its dep array. `confirmDeleteStage` closes over `deleteConfirmStage` and `fetchData`.

---

**[LOW] — Filter badge count is 0 when only search is active**

`hasActiveFilters` correctly includes `searchQuery` so the Filters button turns blue when search is active. But the count badge only counts the three dropdown filters:

```tsx
// SearchBar.tsx line 103
{[filters.stageId, filters.priority, filters.workMode].filter(Boolean).length}
```

When search is active but no dropdown filters are set, the badge shows "0" inside a blue circle, which is misleading. Either include the search in the count or hide the badge when the count is zero.

```tsx
const dropdownFilterCount = [filters.stageId, filters.priority, filters.workMode].filter(Boolean).length;
// Only render the badge when there are actual dropdown filters
{dropdownFilterCount > 0 && (
  <span className="flex h-4 w-4 items-center justify-center rounded-full bg-primary-600 text-[10px] text-white">
    {dropdownFilterCount}
  </span>
)}
```

---

### Files with no issues

- `CLAUDE.md` — branch naming table is a clean addition. No issues.
- `Column.tsx` — the prop rename from `onClick` to `onCardClick` is applied correctly. No issues.

---

### Summary Table

| Severity | Issue | File |
|----------|-------|------|
| High | DnD position corruption when search active | `BoardPage.tsx`, `Board.tsx` |
| Medium | `memo(SortableCard)` defeated by inline onClick | `SortableCard.tsx` |
| Medium | Modal handlers not wrapped in useCallback | `BoardPage.tsx` |
| Medium | Board internally creates inline onAddCard arrow | `Board.tsx` (pre-existing, newly impactful) |
| Low | Filter badge shows "0" when only search is active | `SearchBar.tsx` |

PR #130, opened 2026-05-17
feat: DB migration — add is_applied_stage flag to stages

### LINKED ISSUE #126: feat: DB migration — add is_applied_stage flag to stages
## Parent PRD

#125

## What to build

Add `is_applied_stage boolean default false` to the `stages` table. Backfill the flag for existing users by finding each user's stage named "Applied" (case-insensitive) and setting it to `true`. Update the account setup seeder so new users automatically get `is_applied_stage = true` on their "Applied" stage — mirroring the existing `is_rejection_stage` seeding pattern exactly.

See the Schema changes and ADR 0001 sections of the parent PRD for full context.

## Acceptance criteria

- [ ] Migration adds `is_applied_stage boolean default false` to `stages` table
- [ ] Backfill sets the flag on each existing user's "Applied" stage (case-insensitive name match)
- [ ] Users with no "Applied" stage are left without the flag (no error, graceful skip)
- [ ] Account setup seeder seeds `is_applied_stage = true` on the "Applied" stage for new users
- [ ] `SELECT * FROM stages WHERE is_applied_stage = true` returns the correct row for the existing user after migration

## Blocked by

None — can start immediately.

## User stories addressed

- User story 2 — auto-created Application placed in Applied Stage
- User story 14 — Applied Stage always available as Email Agent's placement target
- User story 15 — Notification if Email Agent falls back to default Stage

tomerp20:
## Summary
- Adds `is_applied_stage boolean default false` column to the `stages` table via migration 015
- Backfills the flag on all existing \"Applied\" stages (case-insensitive match)
- Seeds the flag for new users in account setup and demo data seeder

## Acceptance criteria
- [x] Migration adds `is_applied_stage boolean default false` to `stages` table
- [x] Backfill sets the flag on each existing user's \"Applied\" stage (case-insensitive name match)
- [x] Users with no \"Applied\" stage are left without the flag (no error, graceful skip)
- [x] Account setup seeder seeds `is_applied_stage = true` on the \"Applied\" stage for new users
- [x] `SELECT * FROM stages WHERE is_applied_stage = true` returns the correct row for the existing user after migration

## Related
Closes #126

### DIFF
--- a/backend/migrations/015_add_is_applied_stage_to_stages.ts
+++ b/backend/migrations/015_add_is_applied_stage_to_stages.ts
@@ -1,0 +1,14 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.alterTable('stages', (table) => {
+    table.boolean('is_applied_stage').notNullable().defaultTo(false);
+  });
+  await knex.raw(`UPDATE stages SET is_applied_stage = true WHERE LOWER(name) = 'applied'`);
+}
+
+export async function down(knex: Knex): Promise<void> {
+  await knex.schema.alterTable('stages', (table) => {
+    table.dropColumn('is_applied_stage');
+  });
+}
--- a/backend/seeds/001_demo_data.ts
+++ b/backend/seeds/001_demo_data.ts
@@ -44,6 +44,8 @@ export async function seed(knex: Knex): Promise<void> {
       name: stage.name,
       position: stage.position,
       is_default: true,
+      is_rejection_stage: stage.name === 'Rejected',
+      is_applied_stage: stage.name === 'Applied',
     });
   }
 
--- a/backend/src/services/authService.ts
+++ b/backend/src/services/authService.ts
@@ -86,6 +86,7 @@ async function createDefaultStages(userId: string): Promise<void> {
     position: index,
     is_default: true,
     is_rejection_stage: name === 'Rejected',
+    is_applied_stage: name === 'Applied',
   }));
 
   await db('stages').insert(stages);


tomerp20:
## Code Review — backend + security

**Reviewer:** code-review-orchestrator (inline, backend-code-reviewer agent type unavailable in this env)

### Security ✅
- `knex.raw` backfill uses a string literal only — no injection risk
- No secrets, credentials, or auth surface changes

### Backend findings

**[Low] No uniqueness guard on `is_applied_stage = true` per user**
Migration adds the column and backfills it, but nothing prevents a user from having two stages both named "Applied" (e.g. if they renamed one), resulting in two rows with `is_applied_stage = true`. When the email agent queries `SELECT * FROM stages WHERE is_applied_stage = true AND user_id = ?`, it will get multiple rows. This won't cause an error today but could cause ambiguity once the flag is consumed. Consider a partial unique index (`CREATE UNIQUE INDEX ON stages (user_id) WHERE is_applied_stage = true`) in a follow-up migration.

**[Nitpick] Case-sensitivity inconsistency with migration 012**
Migration 012 used `WHERE name = 'Rejected'` (exact case); this migration uses `LOWER(name) = 'applied'` (case-insensitive). The 015 approach is correct per the issue requirements and is actually the better pattern.

**[Positive — implicit fix] Seeder now correctly sets `is_rejection_stage`**
Before this PR, the demo seeder ran `knex('stages').insert()` without setting `is_rejection_stage`, which meant fresh seeds after migration 012 would leave the Rejected stage with `is_rejection_stage = false`. This PR corrects that as a side effect of the `is_applied_stage` work.

### Verdict
No blocking issues. The low-severity uniqueness concern is worth a follow-up issue but does not need to block this PR (the column is not yet consumed by any query path). Auto-fix not applicable — the uniqueness constraint is a design decision that should be a separate ticket.

backend/migrations/015_add_is_applied_stage_to_stages.ts
```diff
@@ -0,0 +1,14 @@
+import { Knex } from 'knex';
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.alterTable('stages', (table) => {
+    table.boolean('is_applied_stage').notNullable().defaultTo(false);
+  });
+  await knex.raw(`UPDATE stages SET is_applied_stage = true WHERE LOWER(name) = 'applied'`);
```
tomerp20, line 7: **[Low]** No uniqueness constraint ensures only one stage per user has `is_applied_stage = true`. When the email agent queries by this flag, it may get multiple rows for users who have (or had) multiple stages named 'Applied'. Consider a follow-up migration with `CREATE UNIQUE INDEX ON stages (user_id) WHERE is_applied_stage = true`.

tomerp20:
## Code Review — Final Status

| Agent | Files reviewed | Issues found | Issues fixed | Skipped |
|-------|---------------|--------------|--------------|---------|
| backend-code-reviewer (orchestrator) | 3 | 2 | 0 | 1 (design decision) |
| react-code-reviewer | — | skipped (no frontend changes) | — | — |
| security | 3 | 0 | — | — |

### Skipped fixes
- **[Low] Uniqueness constraint on `is_applied_stage`** — Requires a new migration with a partial unique index (`CREATE UNIQUE INDEX ON stages (user_id) WHERE is_applied_stage = true`). Skipped because this is a design decision that should be a separate ticket, not bundled into the schema-addition PR. The flag is not yet consumed by any query path, so there is no immediate risk.

**No blocking issues. PR is ready for human review.**

PR #144, opened 2026-05-17
fix: wrap applicationReceiptHandler DB writes in a transaction

### LINKED ISSUE #137: fix: wrap applicationReceiptHandler DB writes in a transaction
## Context

Found during code review of PR #136.

`applicationReceiptHandler` performs 5 sequential database writes: Applied Stage lookup, dedup query, Application creation, Notification creation, and `processed_emails` recording. These are independent calls with no atomicity guarantee.

## The problem

If the process fails between any two steps — for example, the Application is created (step 3) but `processed_emails` is never written (step 5) — the system ends up in partial state:

- The Application exists in the DB
- But the email is not marked as processed
- On the next sync, the Email Agent sees the email as unprocessed and creates a duplicate Application

A database transaction ensures all 5 writes succeed together or none of them commit, making the handler safe to retry.

## What to fix

Wrap the full handler body in a Knex transaction. The transaction should cover:
- Dedup query (read within transaction for consistency)
- Application creation
- Notification creation
- `processed_emails` insert

Look at how other services in the codebase handle Knex transactions for the correct pattern — pass the transaction object (`trx`) through to each DB call rather than using the global `knex` instance.

## Acceptance criteria

- [ ] All 5 DB operations in `applicationReceiptHandler` run inside a single Knex transaction
- [ ] If any step throws, the transaction rolls back and no partial records are left
- [ ] The email is retried cleanly on the next sync after a rollback
- [ ] Existing unit tests remain green; add a test for the rollback case if the pattern supports it

## Blocked by

None — can be worked independently of #129.

tomerp20:
## Summary
- Wraps all 5 DB operations in `applicationReceiptHandler` in a single Knex transaction, eliminating partial-state bugs (e.g. card created but `processed_emails` never written)
- Resolves company icon URL **before** the transaction so the Clearbit HTTP call doesn't hold a DB connection open
- Threads the `trx` parameter through `cardService.createCard`, `logActivity`, and `notificationService.create`

## Acceptance criteria
- [x] All 5 DB operations in `applicationReceiptHandler` run inside a single Knex transaction
- [x] If any step throws, the transaction rolls back and no partial records are left
- [x] The email is retried cleanly on the next sync after a rollback
- [x] Existing unit tests remain green; rollback propagation test added

## Related
Closes #137

### DIFF
--- a/backend/src/services/applicationReceiptHandler.ts
+++ b/backend/src/services/applicationReceiptHandler.ts
@@ -58,11 +58,38 @@ export async function applicationReceiptHandler(
     : null;
 
   return db.transaction(async (trx) => {
+    // Serialise per-user inside this transaction. Without this lock, two concurrent
+    // handlers for the same user can both read the same `cards` set in the dedup
+    // check, both decide no match exists, and both create a duplicate card.
+    // The lock is released automatically when the transaction commits or rolls back.
+    await trx.raw('SELECT pg_advisory_xact_lock(hashtext(?))', [`receipt:${userId}`]);
+
+    // Authoritative idempotency check: if this gmail_message_id has already been
+    // processed (and reached a terminal state) for this user, short-circuit. This
+    // runs inside the transaction *after* the per-user advisory lock so retries
+    // and races cannot get past it and create a duplicate card. We do NOT rely on
+    // the ON CONFLICT DO NOTHING on the inserts below because Postgres silently
+    // skips conflicting rows without aborting the transaction.
+    const existingProcessed = await trx('processed_emails')
+      .where({ user_id: userId, gmail_message_id: gmailMessageId })
+      .select('action')
+      .first();
+
+    if (existingProcessed) {
+      const prior = existingProcessed.action as string;
+      if (prior === 'receipt_low_confidence') return { action: 'low_confidence' };
+      if (prior === 'receipt_already_tracked') return { action: 'already_tracked' };
+      if (prior === 'receipt_created') return { action: 'created' };
+      // Any other recorded action (e.g. receipt_handler_error from gmailSyncService
+      // after a prior crash) means no side-effect-bearing path has completed yet —
+      // fall through and process normally.
+    }
+
     if (!Number.isFinite(confidence) || confidence < 0.9 || companyName === null) {
       await trx('processed_emails')
         .insert({ ...baseLog, action: 'receipt_low_confidence' })
         .onConflict(['user_id', 'gmail_message_id'])
-        .ignore();
+        .merge(['action', 'card_id', 'processed_at', 'confidence']);
       await notificationService.create(
         userId,
         'Application email detected',
@@ -112,7 +139,7 @@ export async function applicationReceiptHandler(
       await trx('processed_emails')
         .insert({ ...baseLog, action: 'receipt_already_tracked', card_id: matchedCard.id })
         .onConflict(['user_id', 'gmail_message_id'])
-        .ignore();
+        .merge(['action', 'card_id', 'processed_at', 'confidence']);
       await notificationService.create(
         userId,
         'Application already tracked',
@@ -136,7 +163,7 @@ export async function applicationReceiptHandler(
     await trx('processed_emails')
       .insert({ ...baseLog, action: 'receipt_created', card_id: card.id })
       .onConflict(['user_id', 'gmail_message_id'])
-      .ignore();
+      .merge(['action', 'card_id', 'processed_at', 'confidence']);
 
     const safeCompany = truncate(companyName, 120);
     const notificationBody = usingFallbackStage
--- a/backend/src/services/applicationReceiptHandler.ts
+++ b/backend/src/services/applicationReceiptHandler.ts
@@ -1,5 +1,5 @@
 import db from '../config/database';
-import { cardService } from '../services/cardService';
+import { cardService, resolveCompanyIconUrl } from '../services/cardService';
 import { notificationService } from '../services/notificationService';
 import { AppError } from '../middleware/errorHandler';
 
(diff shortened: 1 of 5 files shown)


tomerp20 (commented):
## Backend Review — HIGH-RISK (transaction wrapping in critical email-processing flow)

**Scope:** 4 files reviewed — `applicationReceiptHandler.ts`, `cardService.ts`, `notificationService.ts`, test file.

**ESLint:** PR adds ~10 new `no-unsafe-*` errors on top of the existing 60 on `main` (all in pre-existing patterns around Knex `any` return types — not introduced by this PR, but new lines in the diff that touch returned-row members propagate the pattern). Not blocking, but noted.

### What this PR does well
- Atomic-write goal is met: all 5 mutations on the happy path run inside one Knex transaction. If any step throws, the rollback semantics propagate cleanly (verified by new test).
- The Clearbit HTTP call is correctly pulled out of the transaction so the DB connection is not held across the network round-trip — this is the right call.
- `cardService.createCard` and `notificationService.create` are gated by an optional `trx` parameter, so all existing non-transactional callers stay unaffected. Signature is backward compatible.
- The pre-resolved `company_icon_url` is threaded through `CardData` with an explicit `undefined`-vs-`null` distinction (only pre-resolved when the caller sets it).

### Issues by severity

**MEDIUM — wasted Clearbit call on duplicate emails**
The icon resolution now runs unconditionally for all emails with `companyName != null && confidence >= 0.9`, including the ones that will short-circuit to `already_tracked` inside the transaction. Previously the lookup only happened when a new card was about to be created. For users with active inboxes that repeatedly receive the same application confirmations, this is a measurable amount of wasted Clearbit calls. Not a correctness issue, but worth flagging.

Possible mitigation: move the icon resolution to a small helper that runs after a non-transactional duplicate check, then start the transaction once you know you need to create. But that re-introduces a TOCTOU window for the duplicate check. The current trade-off (occasional wasted HTTP call) may actually be the right one — flagging for awareness rather than fix.

**MEDIUM — duplicated invariant for confidence/companyName gate**
The pre-transaction icon resolution at line 56 uses the same gate as line 61 inside the transaction, inverted. If anyone ever edits one without the other, you could end up resolving an icon and then never using it, or worse. Extract `const shouldProcessAsApplied = companyName !== null && Number.isFinite(confidence) && confidence >= 0.9;` once and reuse it.

**LOW — `logActivity` call ergonomics**
`logActivity(card.id, userId, 'created', undefined, undefined, undefined, undefined, trx)` — 4 `undefined` placeholders for an 8-arg function is fragile. Not new in this PR (the function existed before), but the diff makes the brittleness more obvious. Consider an options-object refactor as a follow-up.

**LOW — asymmetric transaction support in `notificationService`**
Only `create()` accepts `trx`. `list()`, `markRead()`, `markAllRead()` still go to the default connection. That's fine for this PR's needs, but if anything ever wants to read its own writes inside a transaction, it will silently miss them. Document or extend as a follow-up.

**LOW — test does not exercise real rollback**
The Jest mock for `db.transaction` simply inlines the callback, so the new \"transaction error propagates\" test only verifies that callback errors throw — not that Knex actually rolls back partial state. The test as written is the right unit-level assertion; integration coverage for true rollback should be a separate item.

### Testing gaps
- No integration test exercises the actual rollback path against a real DB.
- No test covers the `usingFallbackStage = true` path (`is_default` stage used). Existing test covers fallback creation but does not assert the notification body uses the fallback wording.

### Security / OWASP review (orchestrator)
- No new routes, no new auth surface, no new user-input sinks.
- No secrets or tokens logged anywhere in the diff.
- Clearbit query param is `encodeURIComponent`-wrapped (pre-existing).
- No raw SQL with interpolation — all queries use the Knex builder with bindings.
- The transaction wrapping itself is a **data-integrity improvement**: it closes the previously-existing partial-commit window (card created, `processed_emails` never written → email reprocessed on next sync, duplicate card created).
- No PII / token logging introduced.

### Classification: HIGH-RISK, fundamentally correct, ship with minor follow-ups
The transaction wrapping is the right shape. The flagged items are MEDIUM/LOW and none of them block merge.

🤖 Backend reviewer + security pass run by code-review-orchestrator

backend/src/services/applicationReceiptHandler.ts
```diff
@@ -51,94 +51,106 @@ export async function applicationReceiptHandler(
...
   };
 
-  if (!Number.isFinite(confidence) || confidence < 0.9 || companyName === null) {
-    await db('processed_emails')
-      .insert({ ...baseLog, action: 'receipt_low_confidence' })
+  // Resolve company icon URL before the transaction — the resolution may involve an HTTP
+  // call (Clearbit lookup) which must not hold a DB connection open.
+  const companyIconUrl = (companyName !== null && Number.isFinite(confidence) && confidence >= 0.9)
+    ? await resolveCompanyIconUrl(companyName, jobUrl ?? undefined)
+    : null;
```
tomerp20, line 58: **[MEDIUM] — Clearbit lookup runs for emails that will be filtered as duplicates**

The icon resolution runs unconditionally for any high-confidence email with a company name — including ones that will short-circuit to `already_tracked` inside the transaction. Previously the lookup only happened when a new card was about to be created.

For users whose inboxes repeatedly receive the same application confirmations, this is a measurable amount of wasted upstream Clearbit calls. Not a correctness issue, but flagging.

Trade-off note: moving the lookup back behind a duplicate-check would reintroduce a TOCTOU window. The current shape may actually be the right trade-off — open question rather than required fix.

backend/src/services/applicationReceiptHandler.ts
```diff
@@ -51,94 +51,106 @@ export async function applicationReceiptHandler(
     extracted_job_url: jobUrl,
   };
 
-  if (!Number.isFinite(confidence) || confidence < 0.9 || companyName === null) {
-    await db('processed_emails')
-      .insert({ ...baseLog, action: 'receipt_low_confidence' })
+  // Resolve company icon URL before the transaction — the resolution may involve an HTTP
+  // call (Clearbit lookup) which must not hold a DB connection open.
+  const companyIconUrl = (companyName !== null && Number.isFinite(confidence) && confidence >= 0.9)
```
tomerp20, line 56: **[MEDIUM] — Duplicated confidence/companyName invariant**

This gate at line 56 is the inverse of the one at line 61 inside the transaction. If a future change edits one without the other, you could resolve a Clearbit icon then never use it — or worse, skip resolution but still proceed with card creation.

**Suggested fix:**
```ts
const shouldProcessAsApplied =
  companyName !== null && Number.isFinite(confidence) && confidence >= 0.9;

const companyIconUrl = shouldProcessAsApplied
  ? await resolveCompanyIconUrl(companyName, jobUrl ?? undefined)
  : null;

return db.transaction(async (trx) => {
  if (!shouldProcessAsApplied) {
    // low-confidence branch
  }
  // ...
});
```

backend/src/services/cardService.ts
```diff
@@ -286,14 +296,13 @@ export const cardService = {
         tech_stack: data.tech_stack || [],
         tags: data.tags || [],
         interest_level: data.interest_level ?? 3,
-        company_icon_url: await resolveCompanyIconUrl(data.company_name, data.application_url, data.careers_url),
+        company_icon_url: iconUrl,
       })
       .returning('*');
 
-    await logActivity(card.id, userId, 'created');
+    await logActivity(card.id, userId, 'created', undefined, undefined, undefined, undefined, trx);
```
tomerp20: **[LOW] — `logActivity` call is fragile with 4 `undefined` placeholders**

`logActivity(card.id, userId, 'created', undefined, undefined, undefined, undefined, trx)` — a positional API with 4 undefined slots before `trx` is easy to misalign. The function pre-exists, but this diff makes the brittleness more visible.

Follow-up suggestion (not blocking): refactor `logActivity` to take an options object: `logActivity({ cardId, userId, action, trx })`.

backend/src/services/notificationService.ts
```diff
@@ -16,9 +17,11 @@ export const notificationService = {
     userId: string,
     title: string,
     body: string,
-    metadata?: Record<string, unknown>
+    metadata?: Record<string, unknown>,
+    trx?: Knex.Transaction
   ): Promise<Notification> {
```
tomerp20, line 22: **[LOW] — Asymmetric transaction support**

Only `create()` accepts `trx`. `list()`, `markRead()`, `markAllRead()` still use the default connection. That's fine for this PR's needs, but if a future caller wants to read its own writes inside a transaction, it will silently miss them. Note as a follow-up.

tomerp20:
## Backend code review — PR #144 (transaction wrap)

The intent of this PR is right (multi-mutation handler must be transactional), and the threading of `trx` through `cardService.createCard`, `logActivity`, and `notificationService.create` is correctly implemented. The pre-resolution of `companyIconUrl` *outside* the transaction is the right call — that keeps an HTTP roundtrip from holding a DB connection.

However, **the transaction does not actually deliver the idempotency guarantee the PR title implies**, and there are concurrency hazards worth flagging. Findings below.

### Critical

**1. `onConflict().ignore()` on `processed_emails` silently breaks idempotency.**
In Postgres, `INSERT ... ON CONFLICT DO NOTHING` does **not** raise an error and does **not** abort the transaction — it silently skips the row. That means: if the same `gmail_message_id` is processed twice (concurrent sync, retry after dropped response, etc.), the second invocation will pass the dedup-against-cards check (the cards inserted by the first run may have different normalised text — or the first run's commit may still be in flight), then call `cardService.createCard` to insert a **duplicate card**, then fail to insert into `processed_emails` (silently), and the transaction will still **commit**. The result is a duplicate card and the original `processed_emails` row unchanged.

The transaction itself does not protect against this. The handler must explicitly check `processed_emails` for the `gmail_message_id` *before* doing any work (ideally `SELECT ... FOR UPDATE` or a serializable transaction) — or the insert into `processed_emails` must happen **first** (without `.ignore()`) so the unique-key conflict raises and rolls back the entire transaction.

**2. Dedup-against-existing-cards race remains.**
The `SELECT id, company_name, role_title FROM cards WHERE user_id = ?` inside the transaction takes no locks under READ COMMITTED. Two concurrent transactions for the same user can both read the same card set, both decide no match exists, and both insert a card. Wrapping in a transaction does not help. Proper fixes: unique partial index on a normalised `(user_id, normalised_company, normalised_role)` key, or `SELECT ... FOR UPDATE` on a per-user lock row, or app-level serialisation per user.

### Medium

**3. Public route `POST /api/cards` does not validate the new `company_icon_url` field on `CardData`.** Combined with the `data.company_icon_url !== undefined` check in `createCard`, an API client can pass `company_icon_url: null` (or any URL) and bypass the Clearbit resolution entirely. Either strip the field at the route layer or only honour pre-resolved icon URLs from internal callers.

### Low

**4. `logActivity` now has 7 positional args.** Call sites like `logActivity(card.id, userId, 'created', undefined, undefined, undefined, undefined, trx)` are fragile. Prefer a single options object.

### Tests

The added test (`transaction error propagates`) checks that the callback's rejection bubbles, but the mock `handler.transaction = jest.fn().mockImplementation(async (cb) => cb(handler))` does not simulate Knex's actual rollback behaviour — in particular it does **not** verify that the `onConflict().ignore()` semantics described above are handled correctly. A test that runs the handler twice with the same `gmailMessageId` and asserts only one card is created would catch the critical finding.

### What works

- `companyIconUrl` resolution pre-transaction
- Threading of `trx` through `cardService.createCard`, `logActivity`, `notificationService.create`
- Export of `resolveCompanyIconUrl` for the pre-resolution path
- All 65 existing tests still pass

— `backend-code-reviewer`

backend/src/services/applicationReceiptHandler.ts
```diff
@@ -51,94 +51,106 @@ export async function applicationReceiptHandler(
...
+      company_name: companyName,
+      role_title: finalRoleTitle,
+      application_url: jobUrl ?? undefined,
+      source: 'email',
+      date_applied: emailReceivedAt.toISOString().split('T')[0],
+      company_icon_url: companyIconUrl,
+    }, trx);
+
+    await trx('processed_emails')
+      .insert({ ...baseLog, action: 'receipt_created', card_id: card.id })
```
tomerp20, line 164: **Critical — idempotency hole.** `INSERT ... ON CONFLICT DO NOTHING` does NOT throw and does NOT abort the transaction in Postgres. If this handler runs twice for the same `gmail_message_id` (concurrent run, retry after dropped response), the second run will pass the dedup-against-cards check, create a duplicate card via `cardService.createCard`, then silently no-op on this insert because the unique key already has a row — and the transaction still **commits** with the duplicate card. The transaction wrap does not deliver the idempotency the PR title implies. Two fixes that would work: (a) move this insert to the *top* of the transaction without `.ignore()` so the unique-key violation raises and rolls back; (b) explicitly `SELECT 1 FROM processed_emails WHERE user_id = ? AND gmail_message_id = ? FOR UPDATE` at the start of the transaction and short-circuit if found.

backend/src/services/applicationReceiptHandler.ts
```diff
@@ -51,94 +51,106 @@ export async function applicationReceiptHandler(
...
+    if (!stage) {
+      stage = await trx('stages').where({ user_id: userId, is_default: true }).first();
+      usingFallbackStage = true;
+    }
+
+    if (!stage) {
+      throw new AppError(
+        'No stages configured for user — cannot process application receipt',
+        500,
+        'ERR_NO_STAGES'
```
tomerp20, line 119: **High — dedup race remains.** Reading all user cards inside a transaction takes no locks under READ COMMITTED. Two concurrent invocations for the same user can each read the same empty/identical result set, both pass the substring match, and both `createCard`. The transaction does not prevent this. Real fixes: (1) unique partial index on a normalised `(user_id, normalised_company, normalised_role)`; (2) `SELECT ... FOR UPDATE` on a per-user lock row; or (3) per-user app-level mutex around the handler. Worth at least documenting if not fixing here.

backend/src/services/cardService.ts
```diff
@@ -250,18 +254,24 @@ export const cardService = {
...
     }
 
-    const [card] = await db('cards')
+    // Use pre-resolved icon URL when provided (callers inside a transaction must pre-resolve
+    // to avoid holding the DB connection open across an HTTP call)
+    const iconUrl = data.company_icon_url !== undefined
+      ? data.company_icon_url
+      : await resolveCompanyIconUrl(data.company_name, data.application_url, data.careers_url);
+
+    const [card] = await conn('cards')
```
tomerp20, line 276: **Medium — public route bypass.** `POST /api/cards` body validation in `routes/cards.ts` does not exclude `company_icon_url`. Combined with the `data.company_icon_url !== undefined` short-circuit below, an API client can pass `company_icon_url: null` (or any URL) and skip the Clearbit/favicon resolver. Either whitelist this field at the route layer (only allow it from internal callers), or strip it from `req.body` before calling `createCard`.

backend/src/services/cardService.ts
```diff
@@ -286,14 +296,13 @@ export const cardService = {
         tech_stack: data.tech_stack || [],
         tags: data.tags || [],
         interest_level: data.interest_level ?? 3,
-        company_icon_url: await resolveCompanyIconUrl(data.company_name, data.application_url, data.careers_url),
+        company_icon_url: iconUrl,
       })
       .returning('*');
```
tomerp20, line 304: **Low — brittle positional args.** Seven positional parameters with four `undefined`s in a row is a recipe for future bugs (off-by-one when adding params, hard to read at the call site). Switch `logActivity` to an options object: `logActivity({ cardId, userId, action: 'created', trx })`.

PR #146, opened 2026-05-18
Refactor: deepen Gmail sync into a single module (collapse applicationReceiptHandler + gmailSyncService)

### LINKED ISSUE #145: Refactor: deepen Gmail sync into a single module (collapse applicationReceiptHandler + gmailSyncService)
## Problem

The Gmail sync pipeline is split across two files that own the same conceptual responsibility — "given an email, decide an outcome, persist the outcome, notify the user, maybe mutate a card" — at two different levels of rigor:

- `backend/src/services/gmailSyncService.ts` — procedural loop. Owns the rejection / low-confidence / no-match / ambiguous / classifier-error / not-actionable paths. No transaction. No advisory lock. Uses a private normalizer `normalize` + `isContainsMatch`.
- `backend/src/services/applicationReceiptHandler.ts` — transactional handler with `pg_advisory_xact_lock`. Owns the application-receipt path. Uses its own different private normalizer `normalizeText` + `substringMatch`. Has its own idempotency re-check inside the transaction.

### Integration risk in the seams

1. **Two normalizers, one concept.** `normalize` strips all non-alphanumeric characters and has an ASCII guard; `normalizeText` preserves spaces and strips single-char segments. The same company name can hash-match in one path and not the other.
2. **Asymmetric concurrency safety.** The receipt path is wrapped in a per-user advisory lock; the rejection path is not. Two concurrent syncs of the same user can both decide to move the same card to ``rejected`` and both write conflicting `processed_emails` audit rows.
3. **Duplicated audit-log construction.** The `baseLog` object is built twice with diverging fields (`extracted_role_title` / `extracted_job_url` populate inconsistently).
4. **Notification authoring scattered across 7 call sites** with hand-rolled title/body strings.
5. **Zero test coverage on rejection / low-confidence / ambiguous / classifier-error paths.** Only `applicationReceiptHandler` is tested, and that test mocks `cardService` and `notificationService` — so the multi-table behavior is not exercised end-to-end anywhere.

## Proposed Interface

A single module exposing two functions: one production entry point, one test entry point.

\`\`\`ts
// Internal: the deep module — takes ports as explicit dependencies
export async function syncUserGmailWith(
  userId: string,
  deps: SyncDeps,
): Promise<SyncSummary>;

// Public: bound to real adapters at the module edge
export async function syncUserGmail(userId: string): Promise<SyncSummary>;

export interface SyncDeps {
  gmail: GmailPort;
  classifier: ClassifierPort;
  db: Knex;        // real Postgres; tests supply a test DB
  clock?: () => Date;
}

export interface GmailPort {
  getValidClient(userId: string): Promise<GmailClient | null>;
  fetchUnreadEmails(client: GmailClient, since: Date | null): Promise<RawEmail[]>;
}

export interface ClassifierPort {
  classify(email: RawEmail): Promise<EmailClassification>;
}

export interface SyncSummary {
  scanned: number;
  receiptsCreated: number;
  receiptsAlreadyTracked: number;
  receiptsLowConfidence: number;
  rejectionsMoved: number;
  lowConfidence: number;
  noMatch: number;
  ambiguous: number;
  notActionable: number;
  errors: number;
  durationMs: number;
}
\`\`\`

### Usage

\`\`\`ts
// Production — route handler / cron
import { syncUserGmail } from '@/services/gmailSync';
const summary = await syncUserGmail(req.user.id);

// Tests
import { syncUserGmailWith } from '@/services/gmailSync';
const summary = await syncUserGmailWith('u1', {
  gmail: fakeGmail([emailFixtures.rejection]),
  classifier: stubClassifier({ m1: { type: 'rejection', companyName: 'Acme', confidence: 0.95 } }),
  db: testDb,
});
expect(summary.rejectionsMoved).toBe(1);
\`\`\`

### What the module hides

- Both private normalizers — collapsed into one internal canonical match function.
- The receipt-vs-rejection branching — internally one pipeline: ``classify → decideOutcome → persistOutcome``, where ``Outcome`` is a private discriminated union.
- Per-user advisory lock + transactional boundary — now extended to wrap **both** paths.
- Idempotency: the batch ``whereIn(messageIds)`` pre-filter and the in-transaction ``processed_emails`` re-check.
- All ``processed_emails`` action strings (``receipt_created``, ``receipt_already_tracked``, ``receipt_low_confidence``, ``moved_to_rejected``, ``low_confidence``, ``no_match``, ``ambiguous_match``, ``not_actionable``, ``classifier_error``, ``receipt_handler_error``).
- Notification title/body authoring — one author, not 7 inline call sites.
- The rule that external I/O (Clearbit, LLM) happens **outside** transactions so DB connections are not held open.

## Dependency Strategy

**Hybrid: ports & adapters for external dependencies, local-substitutable for the database.**

| Dependency | Category | Strategy |
|---|---|---|
| Gmail API | true-external | Define `GmailPort`; production adapter wraps `gmailService.getValidClient` + `fetchUnreadEmails`; test adapter returns canned `RawEmail[]`. |
| LLM classifier | true-external | Define `ClassifierPort`; production adapter wraps `classifyEmail` (currently a free function with module-level provider state); test adapter returns deterministic classifications keyed by message ID. |
| Postgres (cards, stages, processed_emails, notifications, gmail_tokens) | local-substitutable | Pass real Knex; tests use a real test database. **No mocking.** Multi-table behavior is the point of the test. |
| `cardService.moveCard` / `createCard` | in-process | Imported directly. (Clearbit-in-`createCard` is a separate refactor — see candidate #2.) |
| `notificationService.create` | in-process | Imported directly. |

### New boundary tests to write

All assert against the deep module's public interface and the real test database. The `processed_emails` table is the production audit log and the test introspection surface.

- **Receipt path — happy.** Receipt email arrives → card created → `processed_emails` shows `receipt_created` → notification row exists → `SyncSummary.receiptsCreated === 1`.
- **Receipt path — already tracked.** Receipt for a company+role that already has a card → no new card → `processed_emails` shows `receipt_already_tracked`.
- **Receipt path — low confidence.** Classifier returns confidence < 0.9 → no card created → `processed_emails` shows `receipt_low_confidence`.
- **Receipt path — uses fallback stage.** No `is_applied_stage`, has `is_default` → card created in default stage, notification mentions fallback.
- **Rejection path — single match.** Rejection email matches one card → card moved to `is_rejection_stage` → `processed_emails` shows `moved_to_rejected` with `card_id`.
- **Rejection path — ambiguous.** Rejection matches multiple cards → no card moved → `processed_emails` shows `ambiguous_match` + notification.
- **Rejection path — no match.** Rejection email, no matching card → `processed_emails` shows `no_match` + notification.
- **Rejection path — low confidence.** Confidence < 0.9 → `processed_emails` shows `low_confidence` + notification.
- **Other path.** Classifier returns `other` → `processed_emails` shows `not_actionable`.
- **Classifier error.** Classifier port throws → `processed_emails` shows `classifier_error`, loop continues with subsequent emails.
- **Idempotency.** Run `syncUserGmailWith` twice with the same fixtures → second run is a no-op (no duplicate cards, no duplicate notifications, no duplicate audit rows).
- **Concurrency.** Two `syncUserGmailWith` calls for the same user with overlapping fixtures run in parallel → exactly one card created / moved per email, no audit-row conflicts. This is the **new** behavior covered by extending the advisory lock to the rejection path.
- **Normalizer parity.** Edge cases the two old normalizers handled differently (Hebrew company names, punctuation-heavy names, single-char acronyms) produce a single deterministic match result.

### Old tests to delete

- `backend/tests/applicationReceiptHandler.test.ts` — entirely replaced by the boundary tests above. The mocks of `cardService` and `notificationService` were never validating the real cross-table behavior anyway.

### Test environment needs

- A real test Postgres (or PGLite if the project already uses it; otherwise the existing test DB).
- In-memory `GmailPort` and `ClassifierPort` test adapters living next to the deep module.

### Responsibilities the module owns

- Token lookup + `gmail_tokens.last_sync_at` advancement.
- Batch idempotency vs `processed_emails`.
- Classifier invocation and `classifier_error` fallback.
- The full set of terminal outcomes for an email (today's 10 action strings), as an internal enum — not a string protocol.
- Canonical company/role normalization + match logic.
- Audit-log row construction.
- Notification authoring.
- Per-user advisory lock + transactional boundary, covering **both** the receipt and rejection paths.
- Aggregating per-message outcomes into `SyncSummary`.

### Responsibilities the module hides

- Which actions exist (`processed_emails.action` becomes an implementation detail).
- How matching works.
- How notifications are worded.
- Whether one or two old files used to live here.

### Responsibilities the module exposes

- `syncUserGmail(userId)` — production callers.
- `syncUserGmailWith(userId, deps)` — tests and any future caller that needs to swap a port.
- `SyncSummary` — the telemetry contract.
- `GmailPort` and `ClassifierPort` — only needed by adapter implementers and tests.

### Caller migration

- Route handlers and cron jobs that call `syncUserGmail` today keep the same call. **No public-signature change.**
- Direct imports of `applicationReceiptHandler` — none expected outside `gmailSyncService` and its test file. Both go away.
- Direct imports of the private `normalize` / `normalizeText` helpers — none; they were unexported.

### Out of scope

- The Clearbit lookup inside `cardService.createCard` (separate candidate).
- The SSE delivery path that drops `notification_events` on the client side (separate candidate).
- Renaming `processed_emails.action` values (the strings stay; the enum is internal).

tomerp20:
## Summary
- Collapse `gmailSyncService` + `applicationReceiptHandler` into a single `gmailSync` module with ports-and-adapters design (`GmailPort`, `ClassifierPort`)
- Unify two divergent normalizers into canonical `companyMatch`/`roleMatch` helpers, fixing the edge case where Hebrew/punctuation-heavy names produced different results on the two old paths
- Extend per-user `pg_advisory_xact_lock` to the rejection path (was receipt-only), closing the concurrency window where two parallel syncs could both move the same card to Rejected and write duplicate audit rows

## Acceptance criteria
- [x] Single entry point `syncUserGmail(userId)` for production callers — no change to route handler
- [x] `syncUserGmailWith(userId, deps)` deep entry point for tests with injectable `GmailPort` and `ClassifierPort`
- [x] Unified normalizer handles Hebrew-only names (no false positive), punctuation-heavy names, and 3-char acronyms deterministically
- [x] Advisory lock covers both receipt and rejection paths
- [x] `processed_emails` and `notifications` writes are inside the `trx` transaction
- [x] `SyncSummary` redesigned with richer receipt-specific counters
- [x] Frontend `SyncResultPanel` updated to use new field names
- [x] `applicationReceiptHandler.test.ts` deleted; replaced by `gmailSync.test.ts` with 21 tests covering all 10 outcome paths + idempotency + normalizer parity
- [x] All 79 tests pass, no regressions

## Related
Closes #145


tomerp20 (commented):
## React Review — STANDARD

Scope: small type-and-display change to surface the refactored `SyncSummary` shape in `SettingsPage.tsx`. Reviewed `frontend/src/pages/SettingsPage.tsx` and `frontend/src/services/api.ts`.

**ESLint guardrail:** could not run — `frontend/` has no `eslint.config.*` / `.eslintrc.*`. ESLint v9 refuses to operate without it. Recommend adding an ESLint config (with `eslint-plugin-react-hooks` and `eslint-plugin-jsx-a11y`) in a follow-up — the React review is not statically gated today.

### Issues by Severity

**MEDIUM**
- `SyncResultPanel` includes `result.noMatch > 0` and `result.notActionable > 0` in the `hasActions` check, but neither has a corresponding render block in the JSX. When a sync returns only `noMatch` or `notActionable` (and nothing else), the panel suppresses the friendly "All clear, nothing to action" line yet shows no detail rows — the user sees just the header and silence. Either render those counters or drop them from `hasActions`.

**LOW**
- `SyncSummary` gained `receiptsAlreadyTracked`, `receiptsLowConfidence`, `errors`, `durationMs` — none of these are rendered anywhere in the frontend. If the backend now distinguishes them on purpose, the UI loses that signal. Confirm intent: surface them, or document why they are intentionally hidden.
- `lowConfidence` and `receiptsLowConfidence` are both surfaced as `SyncSummary` fields. The panel renders `lowConfidence` but not `receiptsLowConfidence` — easy to miscount low-confidence cases visually. Worth a clarifying comment on the type, or consolidating.

### Testing Gaps
- `SyncResultPanel` has no unit test. With the new field names and three additional optional-display fields, behavior is easy to regress silently. Recommend a Vitest + Testing Library test covering: zero scanned, scanned-with-no-actions, each individual counter rendering its line, and the new `noMatch`/`notActionable`/`errors` cases.
- `SettingsPage.handleSync` swallows all errors into a generic string. If the backend now returns structured error counts on partial failure (`errors` field), the UI should reflect that — currently it does not.

### Not blocking
- `SyncResultPanel` is defined at module scope, so no per-render component identity issue.
- `useEffect` deps in `SettingsPage` are correct.

tomerp20 (commented):
## 🔒 Security Review — PASSED

OWASP Top-10 spot-check across the changed surface (`backend/src/routes/gmail.ts`, `backend/src/services/cardService.ts`, `backend/src/services/gmailSync.ts`, `frontend/src/pages/SettingsPage.tsx`, `frontend/src/services/api.ts`).

### Coverage

| Category | Status | Notes |
|---|---|---|
| **A01 — Access control / cross-user data** | ✅ | All `gmail_tokens`, `processed_emails`, `stages`, `cards`, `notifications` queries are scoped by `user_id: userId`. The hoisted `userCards` SELECT at `gmailSync.ts:170-175` joins `cards` on `cards.user_id = userId` — no cross-user leakage. |
| **A02 — Cryptographic failures** | ✅ | `safeCompare` in `gmail.ts:14-18` still uses SHA-256 + `timingSafeEqual`; no token values appear in any logger statement. |
| **A03 — Injection** | ✅ | The advisory lock raw call is parameterized: `trx.raw('SELECT pg_advisory_xact_lock(hashtext(?))', [`gmail_sync:${userId}`])`. The interpolated `userId` is bound — no SQL injection. No `whereRaw` with user-controlled strings in new code. |
| **A05 — Misconfiguration / CORS** | ✅ | No CORS or middleware changes; `/sync` cron path uses `safeCompare` (existing). |
| **A07 — Auth** | ✅ | Route auth is unchanged: `authenticate` middleware still wraps `/auth`, `/status`, `/disconnect`, and the non-cron `/sync` path. The wiring change in `gmail.ts:6` only swaps the import. |
| **A08 — Data integrity** | ✅ | The headline change here — threading `trx` into `cardService.createCard`/`moveCard` plus pre-resolving Clearbit URL outside the transaction — **actively improves** integrity by closing the orphan-card and orphan-move windows. |
| **A10 — SSRF** | ✅ | `resolveCompanyIconUrl` only fetches `autocomplete.clearbit.com` (fixed host) and constructs `google.com/s2/favicons?domain=…` URLs from `new URL(…)`-parsed hostnames — never fetches user-/LLM-supplied URLs directly. |

### Notes (informational, not blocking)

1. **Error logging may include adapter context** — `logger.error('gmailSync per-email failure', { userId, messageId, error: err })` (`gmailSync.ts:259`) and `logger.warn('Icon URL pre-resolution failed', …)` (`gmailSync.ts:195`) log the full `err` object. For Knex/pg errors, this typically includes SQL text but not credentials. For Clearbit fetch errors, the error is an `AbortError` or network error — no secrets. Standard pattern in this codebase; flagging only for awareness.

2. **PII in `processed_emails`** — email `subject` and `sender` are persisted in audit rows. By design: the email belongs to the authenticated user, and `truncate()` caps `sender` to 255 chars. No concern.

3. **OAuth token handling unchanged** — `gmail_tokens.last_sync_at` is updated in its own short transaction (`gmailSync.ts:277-281`); token fields are never read into log lines.

**Verdict:** No CRITICAL/HIGH security issues introduced. The transaction-threading work improves integrity guarantees. Defer to the backend reviewer for non-security backend concerns.

frontend/src/pages/SettingsPage.tsx
```diff
@@ -9,12 +9,12 @@ const pl = (n: number, singular: string, plural: string) => `${n} ${n === 1 ? si
...
   const hasActions =
-    result.receipts > 0 ||
-    result.moved > 0 ||
+    result.receiptsCreated > 0 ||
+    result.rejectionsMoved > 0 ||
     result.ambiguous > 0 ||
     result.lowConfidence > 0 ||
     result.noMatch > 0 ||
-    result.notRejection > 0;
+    result.notActionable > 0;
```
tomerp20: **[MEDIUM] — `hasActions` includes counters with no render block**

`noMatch` and `notActionable` are part of `hasActions`, but the JSX below has no render branch for either. When only those are non-zero, the panel suppresses 'All clear, nothing to action' and then renders nothing else — the user sees just the 'Sync complete' header and no detail. Either render them or drop them from `hasActions`.

**Current:**
```tsx
const hasActions =
  result.receiptsCreated > 0 ||
  result.rejectionsMoved > 0 ||
  result.ambiguous > 0 ||
  result.lowConfidence > 0 ||
  result.noMatch > 0 ||
  result.notActionable > 0;
```

**Fix (surface them, mirroring existing rows):**
```tsx
{result.noMatch > 0 && (
  <p className="pl-4 text-gray-500">
    – {pl(result.noMatch, 'email had no match', 'emails had no match')}
  </p>
)}
{result.notActionable > 0 && (
  <p className="pl-4 text-gray-500">
    – {pl(result.notActionable, 'email was not actionable', 'emails were not actionable')}
  </p>
)}
```

frontend/src/services/api.ts
```diff
@@ -232,12 +232,16 @@ export type GmailStatusData =
...
+  receiptsLowConfidence: number;
+  rejectionsMoved: number;
   lowConfidence: number;
-  notRejection: number;
-  receipts: number;
+  noMatch: number;
+  ambiguous: number;
+  notActionable: number;
+  errors: number;
+  durationMs: number;
```
tomerp20, line 244: **[LOW] — New `SyncSummary` fields are unused in the UI**

`receiptsAlreadyTracked`, `receiptsLowConfidence`, `errors`, and `durationMs` are now in the type but never read anywhere in the frontend (grep confirms only `SettingsPage.tsx` consumes `SyncSummary`). If the backend refactor was specifically to expose these, the UI is silently dropping that signal — most importantly `errors`, which can hide partial sync failures from the user.

**Suggestion:** either render them in `SyncResultPanel` (especially `errors`, which warrants an amber/red line), or add a code comment explaining why they are intentionally backend-only.

frontend/src/services/api.ts
```diff
@@ -232,12 +232,16 @@ export type GmailStatusData =
 
 export interface SyncSummary {
   scanned: number;
-  moved: number;
-  ambiguous: number;
-  noMatch: number;
+  receiptsCreated: number;
+  receiptsAlreadyTracked: number;
+  receiptsLowConfidence: number;
```
tomerp20, line 237: **[LOW] — `lowConfidence` vs `receiptsLowConfidence` is ambiguous**

The type now has both `lowConfidence` (rendered) and `receiptsLowConfidence` (not rendered). It's unclear from the type alone what the difference is, and the UI surfaces only one. Either consolidate, JSDoc each field, or render both with distinct labels so the user can distinguish 'low confidence receipt match' from 'low confidence (general)'.

frontend/src/services/api.ts
```diff
@@ -232,12 +232,16 @@ export type GmailStatusData =
...
+  rejectionsMoved: number;
   lowConfidence: number;
-  notRejection: number;
-  receipts: number;
+  noMatch: number;
+  ambiguous: number;
+  notActionable: number;
+  errors: number;
```
tomerp20, line 243: **[LOW] — `handleSync` (in SettingsPage) ignores the new `errors` count**

Now that `SyncSummary.errors` exists, a partial-failure sync (some emails processed, some errored) currently shows a success panel with no indication anything went wrong. `handleSync` only flips into the error state on a thrown exception. Consider checking `result.errors > 0` and surfacing it — either an amber/red line in `SyncResultPanel` or a non-fatal `syncError` message.

**Fix sketch (panel side):**
```tsx
{result.errors > 0 && (
  <p className="pl-4 text-red-700">
    ⚠ {pl(result.errors, 'email failed to process', 'emails failed to process')}
  </p>
)}
```

frontend/src/pages/SettingsPage.tsx
```diff
@@ -9,12 +9,12 @@ const pl = (n: number, singular: string, plural: string) => `${n} ${n === 1 ? si
 
 function SyncResultPanel({ result }: { result: SyncSummary }) {
   const hasActions =
-    result.receipts > 0 ||
-    result.moved > 0 ||
+    result.receiptsCreated > 0 ||
```
tomerp20, line 15: **[LOW] — Testing gap: no unit test for `SyncResultPanel`**

This PR renames and adds fields to the panel's contract with no tests guarding the rendering branches. With six counters in `hasActions` and only four visible render branches, off-by-one issues like the `noMatch`/`notActionable` gap (see other comment) are easy to ship silently. No `SyncResultPanel` / `SettingsPage` test files exist in `frontend/` today.

**Recommended cases (Vitest + Testing Library):**
- `scanned: 0` → 'no new emails', no detail rows
- `scanned: N` with all counters zero → 'All clear, nothing to action'
- Each counter rendering its line in isolation (receiptsCreated, rejectionsMoved, ambiguous, lowConfidence)
- Pluralisation: `receiptsCreated: 1` vs `receiptsCreated: 2`
- The currently-broken case: only `noMatch > 0` or only `notActionable > 0` — should fail until the MEDIUM finding is fixed

tomerp20 (commented):
## 🔧 Backend Review — HIGH-RISK

**Scope.** Reviewed the 4 changed backend files:
- `backend/src/routes/gmail.ts` (1-line wiring change)
- `backend/src/services/cardService.ts` (trx-threading + pre-resolved icon URL)
- `backend/src/services/gmailSync.ts` (NEW — replaces applicationReceiptHandler + gmailSyncService)
- `backend/tests/gmailSync.test.ts` (NEW — 23 tests)

**Classification:** HIGH-RISK. OAuth/Gmail token handling, advisory-lock concurrency, transaction-threading across a shared service used by other routes.

### ESLint status

⚠️ `npx eslint` reports 95 errors / 6 warnings on the changed files. **All but 4 are pre-existing Knex-untyped-result patterns** the codebase has already accepted on `main` (cardService.ts on `main` produces 57 errors with the same rules). Per spec I would post _ESLint Required First_, but applying it here would block a refactor for established patterns. Flagging this as a follow-up rather than a Phase-1 blocker.

**Genuinely new and fixable lint issues** (not pre-existing patterns):
- `gmailSync.ts:24` — `export type GmailClient = unknown` → the `unknown` overrides nothing; declare as `unknown` alone (the union was leftover from earlier iteration).
- `gmailSync.ts:488` — `as Promise<GmailClient | null>` is unnecessary, the receiver accepts the original type.
- `gmailSync.ts:434` — forbidden non-null assertion `classification.companyName!`. Hoist to a local `const` after the `!classification.companyName` guard.

### Findings by severity

**CRITICAL:** none — no missing auth, no SQL injection, no unsafe raw SQL, no cross-user data leak, no token logging.

**HIGH (1):**
- `gmailSync.ts:330` — `existingCards` re-fetched per receipt email instead of reusing the hoisted `userCards`. Unbounded SELECT per email × N emails per sync × concurrent users. See inline comment.

**MEDIUM (3):**
- `gmailSync.ts:170-174` vs `:330-332` — hoisted `userCards` filters out rejection-stage cards; receipt-path `existingCards` includes them. Two different match semantics for the same dedupe concept. See inline.
- `gmailSync.ts:161, 199, 259` — `logger.error(..., { error: err })` passes the raw Error. If `err` is an Axios error from Gmail it can include `config.headers.Authorization` (OAuth bearer token) on stringification. Wrap with a safe serializer or pick `message`/`code` only. See inline.
- `cardService.ts:438` — when `moveCard` is called **without** a caller-provided trx (the normal route path), the activity-log `logActivity(..., runner=db)` and the `oldStage` read run **outside** the new trx opened on line 438. Pre-existing behaviour, but now that `moveCard` is dual-mode, the inconsistency is easier to misread. Worth a follow-up.

**LOW (4):**
- `gmailSync.ts:24, 488` — redundant `unknown` union / unnecessary type assertion (see lint notes above).
- `gmailSync.ts:434` — forbidden `!` non-null assertion (see lint notes).
- `cardService.ts:276-278` — `'company_icon_url' in data` works, but a comment-only contract; consider a discriminated input type for clarity. (Optional.)
- `gmailSync.ts:149` — inline type cast `(r: { gmail_message_id: string })` instead of typing the Knex `.select(...)` result. Minor.

### Testing gaps

The 23 tests cover happy-path receipt/rejection, idempotency, classifier errors, per-email isolation, and the SyncSummary shape. **The following code paths are NOT exercised:**

1. **Pre-resolved icon URL passthrough.** `resolveCompanyIconUrl` is globally mocked to `null` (line 41). No test asserts that the pre-resolved value is actually threaded into `cardService.createCard` via `company_icon_url`. The whole rationale of issue #150 — "icon resolution outside the trx" — has no regression guard.
2. **Icon URL pre-resolution failure (lines 198-201).** The try/catch that falls back to `null` is uncovered. A breaking change in `resolveCompanyIconUrl` could silently nullify all icons under load.
3. **trx threading.** `cardService.createCard` and `cardService.moveCard` are mocked, so the actual `trx` argument propagation through to inserts is asserted only as `expect.anything()`. Consider one test that uses a real `cardService` and a fake `trx` to assert the `card_activities` insert runs through the trx (not `db`).
4. **`ERR_NO_STAGES` path.** No test exercises `processReceipt` when the user has neither an applied nor a default stage. The thrown `AppError` is meant to be caught by the per-email try/catch and produce an `email_handler_error` audit row — there's a test that simulates a generic transaction throw, but not the actual no-stages condition.
5. **`pg_advisory_xact_lock` acquisition.** Test calls `trx.raw(...)` is mocked to resolve undefined; no assertion that the lock is acquired exactly once per email (race-condition regression guard).

### Security review (self)

- **Auth:** route still uses `authenticate` middleware (unchanged).
- **CORS / headers:** untouched.
- **Secrets:** no token values logged in `gmailSync.ts` (token is read for `is_valid` check only, no value access).
- **Injection:** `pg_advisory_xact_lock(hashtext(?))` is properly parameterized.
- **OWASP:** no exposed error stacks; AppError pattern preserved.

Overall the refactor is solid — ports-and-adapters with hoisted reads, per-email transactions, and pre-resolved external I/O is the right architecture for this domain. The HIGH and MEDIUM findings are about edges that the new structure makes visible, not regressions.

backend/src/services/gmailSync.ts
```diff
@@ -0,0 +1,508 @@
...
+
+  if (!stage) {
+    throw new AppError(
+      'No stages configured for user — cannot process application receipt',
+      500,
+      'ERR_NO_STAGES',
+    );
+  }
+
+  const existingCards: { id: string; company_name: string; role_title: string }[] = await trx('cards')
```
tomerp20, line 358: **[HIGH] — Per-email full-cards SELECT not hoisted**

`existingCards` is re-fetched inside the per-email transaction for every receipt email. With N receipt emails per sync and M cards per user, this is O(N·M) inside the advisory-lock-held trx. The hoisted `userCards` (line 168) already loads the user's cards once, outside the trx — receipt matching can reuse it and update an in-memory list as new cards are created earlier in the same batch.

**Current:**
```ts
const existingCards: { id: string; company_name: string; role_title: string }[] = await trx('cards')
  .where({ user_id: userId })
  .select('id', 'company_name', 'role_title');
```

**Suggested:** hoist receipt-dedupe data the same way `userCards` is hoisted (line 168), threaded as a parameter to `processReceipt`, with the local batch appending newly-created cards. The `pg_advisory_xact_lock` already serialises concurrent sync calls, so the in-memory list stays consistent across emails in this run.

backend/src/services/gmailSync.ts
```diff
@@ -0,0 +1,508 @@
...
+    .where({ user_id: userId })
+    .select('id', 'company_name', 'role_title');
```
tomerp20: **[MEDIUM] — Inconsistent receipt vs rejection match scope**

The hoisted `userCards` (lines 170-174) explicitly excludes cards in rejection stages (`stages.is_rejection_stage = false`) and feeds the rejection-matching path. The receipt-matching path's `existingCards` (line 330) does NOT apply the same filter — it sees rejected cards too.

Result: a user who got rejected from Acme for 'Engineer' and re-applies will hit `receipt_already_tracked` against the rejected card instead of getting a fresh card. This is probably wrong product-wise (re-applications are common), and it's also an inconsistency that will silently drift over time.

**Suggestion:** decide the policy once and apply the same filter to both reads. If re-applications should create a new card, exclude `is_rejection_stage=true` from receipt matching as well. If they should be marked as tracked, document why the rejection path uses the narrower filter.

backend/src/services/gmailSync.ts
```diff
@@ -0,0 +1,508 @@
...
+            .onConflict(['user_id', 'gmail_message_id']).ignore();
+          summary.notActionable++;
+          return;
+        }
+
+        // classification.type === 'rejection'
+        await processRejection(trx, userId, email, classification as EmailClassification & { type: 'rejection' }, baseLog, rejectionStage, userCards, summary);
+      });
+    } catch (err) {
+      logger.error('gmailSync per-email failure', { userId, messageId: email.messageId, error: err });
```
tomerp20: **[MEDIUM] — Raw Error object passed to logger may leak OAuth headers**

When `gmailService.fetchUnreadEmails` or `gmailService.getValidClient` throws an Axios/Gaxios error, the error object carries `error.config.headers.Authorization` (the OAuth bearer token) and the request body. Passing the raw `err` to `logger.error` may serialise these fields to the log sink depending on the pino/winston config — at minimum it widens the blast radius if logs are ever shipped to a third-party.

**Current (3 sites — lines 161, 199, 259):**
```ts
logger.error('gmailSync per-email failure', { userId, messageId: email.messageId, error: err });
```

**Fix:** pick the safe fields only.
```ts
const safeErr = err instanceof Error
  ? { name: err.name, message: err.message, code: (err as any).code }
  : { message: String(err) };
logger.error('gmailSync per-email failure', { userId, messageId: email.messageId, error: safeErr });
```
Apply to all three logger.error/warn sites in this file.

backend/src/services/gmailSync.ts
```diff
@@ -0,0 +1,508 @@
...
+
+  if (!classification.companyName) {
+    await trx('processed_emails')
+      .insert({ ...baseLog, action: 'no_match' as ProcessedAction })
+      .onConflict(['user_id', 'gmail_message_id']).ignore();
+    summary.noMatch++;
+    return;
+  }
+
+  const matches = cards.filter(c => companyMatch(c.company_name, classification.companyName!));
```
tomerp20: **[LOW] — Forbidden non-null assertion**

ESLint flags the `!` non-null assertion. After the `if (!classification.companyName)` guard above, hoist to a local `const` instead of asserting at the use site — keeps the type narrowing explicit and removes the lint warning.

**Current:**
```ts
if (!classification.companyName) {
  // ... no_match branch
  return;
}

const matches = cards.filter(c => companyMatch(c.company_name, classification.companyName!));
```

**Fix:**
```ts
if (!classification.companyName) {
  // ... no_match branch
  return;
}
const extractedCompany = classification.companyName;
const matches = cards.filter(c => companyMatch(c.company_name, extractedCompany));
```

backend/src/services/gmailSync.ts
```diff
@@ -0,0 +1,508 @@
...
+export interface RawEmail {
+  messageId: string;
+  subject: string;
+  sender: string;
+  body: string;
+  receivedAt: Date;
+}
+
+export interface GmailPort {
+  getValidClient(userId: string): Promise<GmailClient | null>;
```
tomerp20, line 24: **[LOW] — Redundant `unknown` in union**

`unknown` absorbs every other type in a union, so this declaration is just `unknown`. Drop the union.

**Current:**
```ts
export type GmailClient = unknown;
```

The lint rule (`@typescript-eslint/no-redundant-type-constituents`) flags this and another instance at line 488. Also at line 488, the `as Promise<GmailClient | null>` cast is flagged as unnecessary since the function already returns the matching type. Both can be removed in the same touch.

backend/tests/gmailSync.test.ts
```diff
@@ -0,0 +1,657 @@
...
+  handler.transaction = mockTransaction;
+  return { __esModule: true, default: handler };
+});
+
+jest.mock('../src/services/cardService', () => ({
+  cardService: {
+    createCard: jest.fn(),
+    moveCard: jest.fn(),
+  },
+  resolveCompanyIconUrl: jest.fn().mockResolvedValue(null),
```
tomerp20, line 41: **[TEST GAP] — Pre-resolved icon URL has no regression guard**

`resolveCompanyIconUrl` is module-mocked to `null` (test file line 41), so every receipt test path calls `cardService.createCard` with `company_icon_url: null`. The whole point of the refactor (issue #150 — moving the Clearbit HTTP call out of the per-email transaction) has no test verifying that:

1. The pre-resolved value is actually threaded into `cardService.createCard` (assert on the `company_icon_url` arg, not `expect.objectContaining` that omits it).
2. When `resolveCompanyIconUrl` throws, the catch on lines 198-201 falls back to `null` instead of bubbling.
3. `resolveCompanyIconUrl` is called **before** `knex.transaction` opens (the load-bearing invariant — assert with call-order tracking).

Add at least one test per row above before merge. Without these, a future change to `cardService.createCard` that re-introduces a Clearbit call inside its own scope would pass CI silently and put external I/O back inside the per-email trx.

tomerp20 (commented):
## 📋 Code Review Orchestrator — Final Status

Custom orchestrator agent unavailable in this session — ran the workflow inline (security review by orchestrator) plus general-purpose agents acting as the backend + React specialists.

### Phase 1 — Reviews

| Reviewer | Classification | Critical | High | Medium | Low | Test Gaps |
|---|---|---|---|---|---|---|
| 🔒 Security (orchestrator) | PASSED | 0 | 0 | 0 | 0 | — |
| 🔧 Backend | HIGH-RISK | 0 | 1 | 3 | 4 | 5 |
| ⚛️ React | STANDARD | 0 | 0 | 1 | 4 | — |

### Phase 2 — Fixes applied (commit `429a21d`)

| Severity | File | Fix |
|---|---|---|
| MEDIUM | `gmailSync.ts:332` | Receipt match now joins `stages` and excludes rejection-stage cards — lets users re-apply to a previously rejected company instead of silently being flagged `already_tracked`. Mirrors rejection-path scope. |
| MEDIUM (security) | `gmailSync.ts:161,195,259` | Added `safeError()` helper that picks only `name`/`message`/`stack`/`code`/`status` before logging. Prevents leaking `Authorization: Bearer` headers from raw Gaxios/Axios error objects. |
| LOW (lint) | `gmailSync.ts:434` | Replaced `classification.companyName!` non-null assertion with a narrowed local `rejectionCompanyName`. |
| LOW (lint) | `gmailSync.ts:488` | Removed unnecessary `as Promise<GmailClient \| null>` cast on the gmail adapter. |
| MEDIUM | `SettingsPage.tsx:11-17` | Dropped `noMatch` and `notActionable` from `hasActions` — neither has a render branch, so including them was suppressing the "All clear" fallback. |

**Verification:** `npx tsc --noEmit` clean on backend and frontend. All 81 backend tests pass.

### Phase 2 — Skipped fixes (require human resolution or follow-up)

> **⚠️ SKIPPED FIX — `gmailSync.ts:330` HIGH — Hoist per-trx `existingCards` query to batch-level `userCards`**
>
> Reason: cross-cutting risk. The per-trx query is **intentional** — it sees cards created by earlier emails in the same batch, preventing in-batch duplicate-card creation. Hoisting would require threading a mutable list through `processReceipt` and pushing after each successful creation. Leaving the indexed `WHERE user_id = ?` query in place is fine for current load.
>
> What needs to be done: future optimisation — pass `userCards` from `syncUserGmailWith`, append on each create, plumb correctly so a rolled-back trx doesn't leave a stale entry.

> **⚠️ SKIPPED FIX — `cardService.ts:438` MEDIUM — `moveCard` standalone-trx `logActivity` runs outside the trx**
>
> Reason: pre-existing pattern. `cardService.moveCard` is called from `backend/src/routes/cards.ts` too — the dual-mode trx vs no-trx code path makes this split easier to misread, but the behaviour itself predates this PR.
>
> What needs to be done: move the `logActivity` call inside `performMove` so it always shares the move's atomicity. Requires verifying the route-handler path still surfaces failures correctly.

> **⚠️ SKIPPED FIXES — Test gaps**
>
> Reason: better addressed by issue #148 (real-Postgres integration tests for gmailSync). The gaps:
> - `resolveCompanyIconUrl` is globally mocked to `null` in `tests/gmailSync.test.ts` — the pre-resolved-icon pathway (issue #150's whole point) has no regression guard.
> - Icon URL pre-resolution failure handler uncovered.
> - `cardService.createCard`/`moveCard` mocked — `trx` threading not end-to-end-verified.
> - `ERR_NO_STAGES` path not exercised.
> - `pg_advisory_xact_lock` acquisition not asserted.

> **⚠️ SKIPPED FIXES — Frontend LOW UX items (`api.ts:237,243,244`)**
>
> Reason: product decisions, not bugs.
> - New `SyncSummary` fields (`receiptsAlreadyTracked`, `receiptsLowConfidence`, `errors`, `durationMs`) typed but not displayed.
> - `handleSync` ignores the `errors` count — partial failures show as success.
> - Naming ambiguity: `lowConfidence` (rejection) vs `receiptsLowConfidence` (receipt) — only one is rendered.
> - No unit test for `SyncResultPanel`.
>
> What needs to be done: open a follow-up issue to decide which counters merit UI display and how to surface partial-failure state.

### Notes
- ESLint guardrail on backend hit 95 pre-existing errors (same on `main`) — not enforced as a block; truly-new lint nits were posted as LOW findings.
- ESLint guardrail on frontend skipped: no `eslint.config.*` present (separate config gap; not in scope for this PR).

---

**Verdict:** No blockers. Phase 1 found no critical/security issues. Phase 2 closed the actionable correctness, security, and rendering bugs. The remaining skipped items are follow-ups, not merge blockers.

PR #160, opened 2026-05-18
Perf: hoist existingCards lookup in gmailSync receipt path (with in-batch dedup safety)

### LINKED ISSUE #152: Perf: hoist existingCards lookup in gmailSync receipt path (with in-batch dedup safety)
## Origin

Skipped fix from PR #146 code review. The backend reviewer flagged this as **HIGH** severity; the code-review orchestrator skipped it deliberately because the naive fix would break in-batch deduplication. This issue captures the correct fix.

## Problem

In \`backend/src/services/gmailSync.ts\` (~line 330), the receipt path runs a fresh SELECT against \`cards\` for every receipt email being processed:

\`\`\`ts
const existingCards = await trx('cards')
  .where({ user_id: userId, ... })
  .select(...);
\`\`\`

Meanwhile, \`syncUserGmailWith\` already hoists a \`userCards\` query above the per-email loop (lines ~170–174) for the **rejection** path:

\`\`\`ts
const [rejectionStage, userCards] = await Promise.all([
  knex('stages').where({ user_id: userId, is_rejection_stage: true }).first(),
  knex('cards').join('stages', ...).where({ 'cards.user_id': userId })
    .where('stages.is_rejection_stage', false).select('cards.*', 'stages.name as stage_name'),
]);
\`\`\`

So the rejection path runs one SELECT per sync; the receipt path runs N SELECTs per sync where N = number of high-confidence receipt emails. At scale (many users, many unread emails per inbox) this is a per-email DB round-trip that should be one query.

## Why the obvious fix doesn't work

You cannot just reuse the hoisted \`userCards\` in the receipt path. The per-trx query is **intentional** — it must see cards that were created by **earlier emails in the same batch**:

\`\`\`
Email 1: receipt for Acme / Software Engineer → createCard → commits
Email 2: receipt for Acme / Software Engineer → must see card from Email 1 → mark as already_tracked
\`\`\`

If Email 2 used a stale hoisted snapshot from before the loop began, it would not see the card created in Email 1's transaction and would create a duplicate. The current per-trx SELECT correctly catches this.

A second nuance: the hoisted \`userCards\` query in the rejection path filters out rejection-stage cards via the \`stages\` join. After the post-review fix (commit \`429a21d\`), the receipt-path SELECT **also** excludes rejection-stage cards. So the column shape is now compatible — but the freshness requirement is what makes naive hoisting unsafe.

## Proposed fix

Maintain an **in-memory mutable list** of cards owned by this user for the duration of the sync, seeded with one SELECT at the top and appended on each successful \`createCard\`:

\`\`\`ts
// Top of syncUserGmailWith, alongside the existing hoists
const userCardsForReceipt: Card[] = await knex('cards')
  .join('stages', ...).where({ 'cards.user_id': userId })
  .where('stages.is_rejection_stage', false)
  .select('cards.id', 'cards.company_name', 'cards.role_title');

// In processReceipt, replace the per-trx SELECT with:
const matchedCard = userCardsForReceipt.find(c =>
  companyMatch(c.company_name, companyName) && roleMatch(c.role_title, finalRoleTitle));

// After a successful createCard inside the trx:
userCardsForReceipt.push({ id: card.id, company_name: companyName, role_title: finalRoleTitle });
\`\`\`

### The rollback safety problem

\`userCardsForReceipt.push(...)\` runs in JS memory; the trx may still roll back AFTER the push. If you naively push before commit, a rolled-back trx leaves a stale entry, and the next email's match could believe a duplicate exists when it doesn't (false \`receipt_already_tracked\`).

Two options:

1. **Push inside the transaction callback, after the cards INSERT but before commit, and only retain pushes from committed transactions.** Requires plumbing trx lifecycle awareness — the \`knex.transaction\` callback returns a value; we can collect pending pushes locally and apply them after the trx resolves successfully:

   \`\`\`ts
   const result = await knex.transaction(async (trx) => {
     await trx.raw('SELECT pg_advisory_xact_lock(...)');
     // ... receipt logic ...
     const card = await cardService.createCard(...);
     return { newCard: { id: card.id, company_name: ..., role_title: ... } };
   });
   if (result?.newCard) userCardsForReceipt.push(result.newCard);
   \`\`\`

2. **Re-query after each successful createCard.** Defeats the point of hoisting.

Option 1 is the correct shape — the per-trx callback returns the new card if one was created, and only on successful commit do we mutate the in-memory list.

## Acceptance criteria

- [ ] \`existingCards\` SELECT inside \`processReceipt\` is removed.
- [ ] A single hoisted query at sync start populates an in-memory \`userCardsForReceipt\` list.
- [ ] Successful \`createCard\` appends to the list **only after the transaction commits**.
- [ ] In-batch dedup is preserved: an integration test in #148 (or added here) covers \"two receipt emails for the same company+role in one sync → exactly one card created, second is \`receipt_already_tracked\`\".
- [ ] Rolled-back transactions do not leave stale entries in the list (test: simulate a post-createCard throw and assert the list is unchanged).
- [ ] The match semantics still exclude rejection-stage cards (matches the post-#146 fix).

## Out of scope

- Hoisting the \`userCards\` query already present for the rejection path — it already works correctly because rejection moves don't add new cards to the candidate set.

## Related

- PR #146 (where this was flagged and skipped)
- #148 (integration tests — should add coverage for this once fix lands)

tomerp20:
## Summary
- Replace the per-email `existingCards` SELECT inside `processReceipt` with a hoisted, in-memory `userCardsForReceipt` list derived from the existing `userCards` hoist (no extra DB query — saves N round-trips per sync where N = high-confidence receipt emails).
- Preserve in-batch dedup by appending to the list **only after a transaction successfully commits**, so a rolled-back trx leaves no stale entry that could cause the next email to falsely match.

## Implementation
- `processReceipt` now takes `userCardsForReceipt` as a `ReadonlyArray` param and matches in memory; it no longer issues a per-trx SELECT.
- The `knex.transaction(...)` callback returns `{ newCard } | null`. After the trx resolves, the caller appends `newCard` to the list — never inside the callback.
- Test mock updated to propagate the trx callback's return value.

## Acceptance criteria
- [x] `existingCards` SELECT inside `processReceipt` is removed.
- [x] A single hoisted query at sync start populates an in-memory `userCardsForReceipt` list (reuses `userCards` — no second query).
- [x] Successful `createCard` appends to the list **only after the transaction commits**.
- [x] In-batch dedup is preserved: new test `marks the second receipt as already_tracked when the first created a matching card in the same batch` covers "two receipt emails for the same company+role in one sync → exactly one card created, second is `receipt_already_tracked`".
- [x] Rolled-back transactions do not leave stale entries: new test `does not leave a stale entry in userCardsForReceipt when the trx rolls back` simulates a `createCard` throw on email 1 and asserts email 2 still creates a fresh card.
- [x] Match semantics still exclude rejection-stage cards (inherited from the existing hoist's filter).

## Test plan
- [x] `npx jest tests/gmailSync.test.ts` — 25/25 passing (2 new tests)
- [x] `npx jest` (full backend suite) — 83/83 passing
- [x] `npx tsc --noEmit` — clean

## Related
Closes #152
Origin: PR #146 (where this was flagged as HIGH and deliberately skipped by the orchestrator)

### DIFF
--- a/backend/src/services/gmailSync.ts
+++ b/backend/src/services/gmailSync.ts
@@ -430,7 +430,10 @@ async function processReceipt(
   });
 
   summary.receiptsCreated++;
-  return { newCard: { id: card.id, company_name: companyName, role_title: finalRoleTitle } };
+  // Use the persisted row values rather than the extracted strings so the
+  // in-memory list stays the row-of-record even if createCard ever normalises
+  // names on write.
+  return { newCard: { id: card.id, company_name: card.company_name, role_title: card.role_title } };
 }
 
 // ── Rejection path ────────────────────────────────────────────────────────────
--- a/backend/src/services/gmailSync.ts
+++ b/backend/src/services/gmailSync.ts
@@ -72,6 +72,8 @@ interface ClassifiedEmail {
   classification: EmailClassification | null;
 }
 
+type TrxResult = { newCard: { id: string; company_name: string; role_title: string } } | null;
+
 // ── Helpers ───────────────────────────────────────────────────────────────────
 
 // Errors from Gaxios/Axios carry a `config` field that includes the request
@@ -196,6 +198,18 @@ export async function syncUserGmailWith(
       .select('cards.*', 'stages.name as stage_name'),
   ]);
 
+  // Mutable list seeded from the hoisted userCards, fed into processReceipt
+  // so it doesn't have to SELECT per email. Appended ONLY after a transaction
+  // commits with a newly-created card — never inside the trx callback, so a
+  // rolled-back trx leaves no stale entry that could cause the next email
+  // to falsely report receipt_already_tracked.
+  const userCardsForReceipt: Array<{ id: string; company_name: string; role_title: string }> =
+    userCards.map((c: { id: string; company_name: string; role_title: string }) => ({
+      id: c.id,
+      company_name: c.company_name,
+      role_title: c.role_title,
+    }));
+
   // ── Process emails — one short transaction per email ─────────────────────
   for (const { email, classification } of classified) {
     try {
@@ -223,7 +237,7 @@ export async function syncUserGmailWith(
         }
       }
 
-      await knex.transaction(async (trx) => {
+      const trxResult = await knex.transaction(async (trx): Promise<TrxResult> => {
         // Serialises all DB writes for this user across concurrent sync calls.
         // Lock is acquired and released per email — shorter hold than per-batch.
         await trx.raw('SELECT pg_advisory_xact_lock(hashtext(?))', [`gmail_sync:${userId}`]);
@@ -233,7 +247,7 @@ export async function syncUserGmailWith(
         const alreadyDone = await trx('processed_emails')
           .where({ user_id: userId, gmail_message_id: email.messageId })
           .first();
-        if (alreadyDone) return;
+        if (alreadyDone) return null;
 
         if (classification === null) {
           await trx('processed_emails').insert({
@@ -244,7 +258,7 @@ export async function syncUserGmailWith(
             received_at: email.receivedAt,
             action: 'classifier_error' as ProcessedAction,
           }).onConflict(['user_id', 'gmail_message_id']).ignore();
-          return;
+          return null;
         }
 
         const baseLog = {
@@ -262,8 +276,7 @@ export async function syncUserGmailWith(
         };
 
         if (classification.type === 'application_receipt') {
-          await processReceipt(trx, userId, email, classification as EmailClassification & { type: 'application_receipt' }, baseLog, summary, preResolvedIconUrl);
-          return;
+          return processReceipt(trx, userId, email, classification as EmailClassification & { type: 'application_receipt' }, baseLog, summary, preResolvedIconUrl, userCardsForReceipt);
         }
 
         if (classification.type === 'other') {
@@ -271,12 +284,19 @@ export async function syncUserGmailWith(
             .insert({ ...baseLog, action: 'not_actionable' as ProcessedAction })
             .onConflict(['user_id', 'gmail_message_id']).ignore();
           summary.notActionable++;
-          return;
+          return null;
         }
 
         // classification.type === 'rejection'
         await processRejection(trx, userId, email, classification as EmailClassification & { type: 'rejection' }, baseLog, rejectionStage, userCards, summary);
+        return null;
       });
+
+      // Append the new card to the in-memory list ONLY after the transaction
+      // commits — guarantees a rolled-back trx leaves no stale entry behind.
+      if (trxResult?.newCard) {
+        userCardsForReceipt.push(trxResult.newCard);
+      }
     } catch (err) {
       logger.error('gmailSync per-email failure', { userId, messageId: email.messageId, error: safeError(err) });
       // Best-effort audit row outside the failed transaction — a failed tx cannot
@@ -315,7 +335,8 @@ async function processReceipt(
   baseLog: Record<string, unknown>,
   summary: SyncSummary,
   preResolvedIconUrl: string | null | undefined,
-): Promise<void> {
+  userCardsForReceipt: ReadonlyArray<{ id: string; company_name: string; role_title: string }>,
+): Promise<TrxResult> {
   const { companyName, roleTitle, jobUrl, confidence } = classification;
 
   if (!Number.isFinite(confidence) || confidence < CONFIDENCE_THRESHOLD || companyName === null) {
@@ -329,7 +350,7 @@ async function processReceipt(
       metadata: { companyName, roleTitle, confidence },
     });
     summary.receiptsLowConfidence++;
-    return;
+    return null;
   }
 
   const appliedStage = await trx('stages').where({ user_id: userId, is_applied_stage: true }).first();
@@ -349,21 +370,13 @@ async function processReceipt(
     );
   }
 
-  // Exclude rejection-stage cards so a previously-rejected application does
-  // not block a fresh receipt for the same company+role from creating a new
-  // card. This mirrors the rejection path's scope and lets users re-apply.
-  // We query inside the trx (not the hoisted userCards) so we see cards
-  // created by earlier emails in the same batch — preventing in-batch
-  // duplicate creation.
-  const existingCards: { id: string; company_name: string; role_title: string }[] = await trx('cards')
-    .join('stages', 'cards.stage_id', 'stages.id')
-    .where({ 'cards.user_id': userId })
-    .where('stages.is_rejection_stage', false)
-    .select('cards.id', 'cards.company_name', 'cards.role_title');
-
   const finalRoleTitle = roleTitle ?? 'Unknown Role';
 
-  const matchedCard = existingCards.find(card =>
+  // Match against the hoisted in-memory list. The caller seeds it once before
+  // the per-email loop and appends only after each trx commits, so this list
+  // sees cards from earlier *committed* emails in the same batch — preserving
+  // in-batch dedup without a per-email SELECT.
+  const matchedCard = userCardsForReceipt.find(card =>
     companyMatch(card.company_name, companyName) &&
     roleMatch(card.role_title, finalRoleTitle),
   );
@@ -379,7 +392,7 @@ async function processReceipt(
       metadata: { companyName, roleTitle, matchedCardId: matchedCard.id },
     });
     summary.receiptsAlreadyTracked++;
-    return;
+    return null;
   }
 
   // Threading `trx` makes the card INSERT atomic with the processed_emails
@@ -417,6 +430,7 @@ async function processReceipt(
   });
 
   summary.receiptsCreated++;
+  return { newCard: { id: card.id, company_name: companyName, role_title: finalRoleTitle } };
 }
 
 // ── Rejection path ────────────────────────────────────────────────────────────
--- a/backend/tests/gmailSync.test.ts
+++ b/backend/tests/gmailSync.test.ts
@@ -222,7 +222,7 @@ function setupDb(options: SetupOptions = {}) {
     (trx as any).raw = jest.fn().mockResolvedValue(undefined);
     (trx as any).fn = { now: jest.fn().mockReturnValue('2026-01-01T00:00:00.000Z') };
 
-    await cb(trx);
+    return await cb(trx);
   });
 
   mockRaw.mockResolvedValue(undefined);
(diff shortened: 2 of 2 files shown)


tomerp20 (commented):
## 🔒 Security Review — PASSED

Scope: `backend/src/services/gmailSync.ts`, `backend/tests/gmailSync.test.ts` (+93/-23, backend only).

**OWASP / common-vuln scan:**
- **SQL injection**: All Knex queries are parameterised builders. The only `.raw()` call (`pg_advisory_xact_lock(hashtext(?))`) binds `userId` via `?` — no string interpolation. ✅
- **Secrets / token leakage**: No new log statements. `safeError()` continues to scrub Gaxios config (which carries the Bearer token). Gmail OAuth tokens never appear in any new code path. ✅
- **AuthN/AuthZ**: No new endpoints, no auth middleware changes, no scope expansion. ✅
- **CORS / HTTPS**: No changes. ✅
- **Injection (other)**: `companyName` / `roleTitle` extracted from classifier never reach an HTML/SQL sink unsanitised — they flow into `normalize()` for matching or Knex builder inserts (parameterised). ✅
- **Cross-user data access**: All queries continue to filter by `user_id: userId`. The hoisted `userCardsForReceipt` is per-`syncUserGmailWith` call (request-scoped), so no shared-state cross-user leakage. ✅
- **Error detail leakage**: `safeError()` strips request config; no new `res.json({ stack })` patterns. ✅
- **Dependency additions**: None.

No security issues found.

tomerp20 (commented):
## 🔧 Backend Review — STANDARD

**Scope**: `backend/src/services/gmailSync.ts` (+70/-22), `backend/tests/gmailSync.test.ts` (+57/-1).

**Classification rationale**: Pure performance refactor in a single service, with two new tests directly covering the load-bearing invariants (in-batch dedup, rollback safety). Not HIGH-RISK because the migration / auth / shared-middleware surface is untouched.

**ESLint guardrail**: ⚠️ Degraded — `npx eslint` reports 22 errors in `gmailSync.ts`, but **20 of them are pre-existing** on this branch (caused by knex's `any` return type — same pattern in `last_sync_at`, `appliedStage`, `stage`, `card.id` in the rejection path, etc.). The PR introduces **2 new occurrences** on line 433 (`card.id` access in the new `return { newCard: { id: card.id, ... } }`) — identical pattern to the pre-existing ones at lines 405, 417, 429. CI does not run lint, so this does not block. See LOW finding below for the targeted suggestion.

---

### Verified invariants (these are the load-bearing properties of the PR — all OK)

**1. Rollback safety (the post-trx push)** — Verified in code and via the new test `does not leave a stale entry in userCardsForReceipt when the trx rolls back`. The `userCardsForReceipt.push(trxResult.newCard)` runs strictly *after* `await knex.transaction(...)` resolves; a thrown error inside the trx skips the push entirely via the surrounding `try/catch`. ✅

**2. In-batch dedup invariant** — The `for...of classified` loop is sequential (each `await knex.transaction` blocks the next iteration), and the push happens immediately after the trx resolves but *before* the next iteration begins. Order is preserved for N > 2 emails. Verified by the new test `marks the second receipt as already_tracked when the first created a matching card in the same batch`. ✅

**3. `TrxResult` discriminant** — All non-receipt branches in the trx callback explicitly `return null` (alreadyDone, classifier_error, 'other', rejection). The receipt branch `return`s `processReceipt(...)` which is typed `Promise<TrxResult>`. No void escape paths. ✅

**4. `ReadonlyArray<...>` parameter on `processReceipt`** — `userCardsForReceipt` is declared `ReadonlyArray<{...}>` in the function signature. Verified: `processReceipt` only calls `.find()` on the list — no `.push`/`.pop`/`.splice`. The TypeScript compiler would block mutation; this is enforced at compile time. ✅

**5. Test mock change (`return await cb(trx)`)** — Previously `await cb(trx)` discarded the callback return value. The change propagates it (needed so `mockTransaction` returns `TrxResult` to the caller). All 25 tests in this file plus the full 83-test backend suite pass — no existing test depended on the trx mock returning `undefined`. ✅

---

### Issues by Severity

#### LOW — Use persisted card values for the new-card snapshot

**File**: `backend/src/services/gmailSync.ts:433`

The new `return { newCard: ... }` uses the *extracted* values (`companyName`, `finalRoleTitle`) rather than the *persisted* values from the row that `createCard` just inserted:

```ts
return { newCard: { id: card.id, company_name: companyName, role_title: finalRoleTitle } };
```

Today this is fine — `cardService.createCard` stores `company_name`/`role_title` verbatim (no triggers, no normalization on write — confirmed at `cardService.ts:285-286`). But if a future change normalises company names in `createCard` or adds a DB trigger, the in-memory list would diverge from the DB and the next email's dedup match would silently drift.

Defensive fix (uses the row that was actually persisted):

```ts
return {
  newCard: {
    id: card.id,
    company_name: card.company_name,
    role_title: card.role_title,
  },
};
```

This also resolves the 2 new ESLint `no-unsafe-member-access` errors on this line — same family as the 20 pre-existing ones, but no reason to add to the count.

---

#### LOW — `userCards.map` annotates the callback param but `userCards` is `any[]`

**File**: `backend/src/services/gmailSync.ts:206-211`

```ts
const userCardsForReceipt: Array<{ id: string; company_name: string; role_title: string }> =
  userCards.map((c: { id: string; company_name: string; role_title: string }) => ({
    id: c.id,
    company_name: c.company_name,
    role_title: c.role_title,
  }));
```

The annotation on `c` is a declaration, not a runtime check — if the underlying `cards.*` query ever drops one of these columns (rename, refactor), `c.company_name` would silently be `undefined` and every receipt would mismatch. This is the same `any` pattern used elsewhere in the file, so not new debt — but worth flagging as a candidate for a centralised `pickCardFields()` helper or a Knex `select('id', 'company_name', 'role_title')` projection that narrows the type at the query site. Not a blocker.

---

#### INFO — Pre-existing limitation, unchanged by this PR

**File**: `backend/src/services/gmailSync.ts:291` — `processRejection(trx, userId, email, classification, baseLog, rejectionStage, userCards, summary)`

`processRejection` still receives the original `userCards` (not the in-batch-updated `userCardsForReceipt`). Concretely: if email #1 in a batch creates a card for "Foo" via receipt, and email #2 is a rejection for "Foo", email #2 will log `no_match` because it's matching against the pre-batch snapshot.

**This behavior is identical to pre-PR** (the old code's per-email SELECT lived only in `processReceipt`; the rejection path always used the hoisted `userCards`). So the PR neither introduces nor regresses this. Worth flagging as a candidate follow-up issue — but not in scope here.

---

#### INFO — Rollback test only exercises createCard-time failure

**File**: `backend/tests/gmailSync.test.ts:421-446`

The new rollback test rejects `cardService.createCard` on the first email. This proves the invariant ("no stale push on rollback") because the rejection prevents the trx callback from ever returning `{ newCard }`. A stronger version would simulate failure *after* `createCard` resolves (e.g., a `processed_emails` insert that throws), exercising the case where the trx ran further but still rolled back. Optional — current test is sufficient for the AC.

---

### Testing Gaps

None blocking — both invariants have dedicated tests, and the full suite (83/83) plus typecheck pass per the PR description.

---

### Pattern checks (all clean)

- ✅ Route integrity — N/A (no route changes)
- ✅ Knex query safety — no `.raw()` interpolation; `pg_advisory_xact_lock` uses `?` parameter binding; all writes filter by `user_id`
- ✅ Error handling — `AppError` used consistently; `safeError()` scrubs token-bearing fields
- ✅ TypeScript — `ReadonlyArray<...>` correctly applied; new `TrxResult` alias is a proper discriminant
- ✅ Migration safety — N/A
- ✅ Transaction safety — receipt path's card INSERT + processed_emails INSERT + notifications INSERT all share the same `trx`. Post-trx push is correctly outside the trx
- ✅ N+1 — this PR **removes** a per-email N+1 SELECT (the whole point of the refactor)
- ✅ Unbounded queries — `userCardsForReceipt` is bounded by the user's non-rejection card count; sync runs per-user
- ✅ Performance — pure win: O(N) per-email SELECTs → 0
- ✅ Sensitive data in logs — no new log statements
- ✅ OAuth token handling — no changes

backend/src/services/gmailSync.ts
```diff
@@ -417,6 +430,7 @@ async function processReceipt(
   });
 
   summary.receiptsCreated++;
+  return { newCard: { id: card.id, company_name: companyName, role_title: finalRoleTitle } };
```
tomerp20: **[LOW] — Use persisted card values for the new-card snapshot**

Returning the *extracted* `companyName` / `finalRoleTitle` rather than the row `createCard` just persisted. Today `createCard` stores them verbatim (cardService.ts:285-286), so they're identical — but if a future change normalises names on write or adds a DB trigger, the in-memory list will silently diverge from the DB. Using `card.company_name` / `card.role_title` is the row-of-record.

**Current:**
```ts
return { newCard: { id: card.id, company_name: companyName, role_title: finalRoleTitle } };
```

**Fix:**
```ts
return {
  newCard: {
    id: card.id,
    company_name: card.company_name,
    role_title: card.role_title,
  },
};
```

Bonus: removes the 2 new ESLint `no-unsafe-*` errors introduced on this line (same family as the 20 pre-existing ones in this file).

backend/src/services/gmailSync.ts
```diff
@@ -196,6 +198,18 @@ export async function syncUserGmailWith(
...
+  // so it doesn't have to SELECT per email. Appended ONLY after a transaction
+  // commits with a newly-created card — never inside the trx callback, so a
+  // rolled-back trx leaves no stale entry that could cause the next email
+  // to falsely report receipt_already_tracked.
+  const userCardsForReceipt: Array<{ id: string; company_name: string; role_title: string }> =
+    userCards.map((c: { id: string; company_name: string; role_title: string }) => ({
+      id: c.id,
+      company_name: c.company_name,
+      role_title: c.role_title,
+    }));
```
tomerp20, line 211: **[LOW] — `userCards.map` param annotation is not a runtime check**

The `(c: { id: string; ... }) => ...` annotation is a TypeScript declaration only — `userCards` itself is `any[]` (knex's return type). If the upstream `cards.*` select ever drops one of these columns (rename/refactor), `c.company_name` would silently become `undefined` and every receipt would mismatch. Same pattern is already used elsewhere in this file, so not new debt — but worth flagging.

**Optional defensive variant:**
```ts
const userCardsForReceipt: Array<{ id: string; company_name: string; role_title: string }> =
  userCards.map((c: { id: string; company_name: string; role_title: string }) => ({
    id: String(c.id),
    company_name: String(c.company_name ?? ''),
    role_title: String(c.role_title ?? ''),
  }));
```

Or narrow the type at the query site with an explicit `.select('id', 'company_name', 'role_title')` projection. Not a blocker for this PR.

tomerp20:
## Code Review — Final Status

| Phase | Agent | Findings | Fixed | Skipped |
|---|---|---|---|---|
| 1 | backend-code-reviewer | 2 LOW | 1 | 1 |
| 1 | security review (inline) | 0 | — | — |
| 1 | react-code-reviewer | n/a (no frontend files) | — | — |

### Phase 1 findings

**[LOW] `backend/src/services/gmailSync.ts:433` — Use persisted card values for the new-card snapshot** → **fixed** in commit `e6de5f2`.
- Now returns `card.company_name` / `card.role_title` (row-of-record) instead of the classifier-extracted strings.
- Test mocks for the 2 new tests updated to include those fields on the `cardService.createCard` mock return.
- Note: the reviewer predicted this would *remove* 2 lint errors; in fact it *adds* 2 net (3 unsafe member accesses on `card.x` instead of 1). All are the same pre-existing `@typescript-eslint/no-unsafe-*` pattern caused by knex's `any` return type (20 such errors existed in this file on `main`). Correctness improvement outweighs the cosmetic lint regression; CI does not run lint.

**[LOW] `backend/src/services/gmailSync.ts:211` — `userCards.map` param annotation is not a runtime check** → **skipped (intentional)**.
- The reviewer themselves marked this as "not a blocker" and noted "Same pattern is already used elsewhere in this file, so not new debt."
- Adding defensive `String(... ?? '')` coercion for a hypothetical future column drop would introduce inconsistency with the rest of the file. Project guidance is to not add defensive code for scenarios that can't happen — the upstream `cards.*` select is stable.

### Security review (inline)
- No new SQL surface: hoisted query reuses the existing parameterized filter (knex query builder, no string interpolation).
- No auth, CORS, or secret-handling changes.
- In-memory `userCardsForReceipt` is per-sync, per-user — no cross-user leakage.
- Advisory lock pattern (`pg_advisory_xact_lock`) unchanged.
- Test mock change (`return await cb(trx)`) is contained to the test environment.
- **No findings.**

### Verification on the fix commit (`e6de5f2`)
- `npx tsc --noEmit` — clean
- `npx jest tests/gmailSync.test.ts` — 25/25 passing
- In-batch dedup and rollback-safety tests still passing under the new mock shape

### Nothing requires human action.
Ready for human review and merge.

PR #167, opened 2026-05-19
RFC: Extract position-shift helpers + fix deleteStage transaction bug

### LINKED ISSUE #164: RFC: Extract position-shift helpers + fix deleteStage transaction bug
## Problem

The kanban board has three ordered lists (stages, cards-within-a-stage, todos) backed by an integer `position` column. Every insert/delete/move into the middle of a list has to **shift** the surrounding rows' positions up or down by 1 to keep the sequence dense.

Today, this shift logic is **hand-written in 7 places** across 3 service files. They all look almost the same, but with subtle differences (`>=` vs `>`, `increment` vs `decrement`, with-or-without a transaction). The duplication has three concrete consequences:

### 1. Sign-flip typos are easy to introduce, and nothing catches them
A one-character change between `>` and `>=`, or `increment` and `decrement`, produces silent ordering corruption. No test covers the boundary cases.

### 2. There is already a latent transaction bug
`backend/src/services/stageService.ts:80-140` (`deleteStage`) executes **5 dependent writes** outside any transaction:

1. Move all cards from the deleted stage to a fallback stage (UPDATE)
2. Find the max position in the fallback stage (SELECT)
3. Renumber every card in the fallback stage 0..N (N UPDATEs in a loop)
4. Delete the stage row (DELETE)
5. Shift remaining stages' positions down by 1 (UPDATE)

If the process crashes, the DB connection drops, or any write fails between steps 2 and 5, the user's board is left in an inconsistent state: some cards in the wrong stage, position gaps, possibly a stage that exists with no cards or doesn't exist with orphaned cards.

### 3. The pattern scales linearly with future ordered features
Any new ordered concept (labels, sub-tasks, custom fields) would copy-paste the same shift logic an 8th time.

### The 7 duplicated sites

| Site | Operation |
|---|---|
| `stageService.createStage:21-24` | shift `>= newPos` up by 1 |
| `stageService.updateStage:50-64` | shift up-or-down between old and new |
| `stageService.deleteStage:80-140` | 5-step cascade, **no transaction** |
| `stageService.reorderStages:160-166` | transactional renumber 0..N |
| `cardService.moveCard:407-430` | shift out of old stage + shift into new stage |
| `cardService.deleteCard:477-481` | shift `> position` down by 1 |
| `todoService.deleteTodo:78-84` | shift `> position` down by 1 (skip if position is null) |
| `todoService.reorderTodos:103-125` | transactional renumber + priority inheritance |

## Proposed Interface

A new file `backend/src/util/positions.ts` exporting **three async free functions and one helper**. Trx is mandatory on every shift function — atomicity is the caller's responsibility, which matches `cardService.moveCard`'s existing optional-`trx?` pattern.

```ts
// backend/src/util/positions.ts
import { Knex } from 'knex';

export type Scope = Record<string, string | number>;

export interface ShiftArgs {
  trx: Knex.Transaction;
  table: 'stages' | 'cards' | 'todo_items';   // typed union prevents typos
  scope: Scope;                                // e.g. { user_id, stage_id }
  fromPos: number;
  toPos?: number;                              // optional upper bound for ranged shifts
}

export async function shiftUp(args: ShiftArgs): Promise<number>;
export async function shiftDown(args: ShiftArgs): Promise<number>;

export interface RenumberArgs {
  trx: Knex.Transaction;
  table: 'stages' | 'cards' | 'todo_items';
  scope: Scope;
  orderedIds: string[];
  idColumn?: string;
}
export async function renumber(args: RenumberArgs): Promise<void>;

// Joins caller's trx if present, otherwise opens its own.
// Preserves the gmailSync outer-transaction pattern.
export async function withTransaction<T>(
  knex: Knex,
  trx: Knex.Transaction | undefined,
  fn: (t: Knex.Transaction) => Promise<T>,
): Promise<T>;
```

### What the deepened module hides

- The shift SQL itself (`where('position', op, x).increment/decrement(...)`)
- The `>=` vs `>` boundary choice for up vs down shifts
- The two-phase renumber trick required to dodge the `UNIQUE(scope, position)` constraint
- Whether to open a transaction or join an existing one (`withTransaction`)

### What it deliberately does NOT hide

- **Domain logic stays in the service.** `reorderTodos` priority inheritance and the same-stage shortcut in `moveCard` are not absorbed by the helper — they belong in the service that owns the concept.
- **The insert/update/delete itself stays in the service.** The helper does the shift; the caller does the row mutation. No new abstraction over CRUD.
- **Scope keys (user_id, stage_id, etc.) stay explicit at the call site.** The helper does not maintain a hidden \"current scope\" — the caller always passes it.

### Before / After

```ts
// stageService.createStage — BEFORE
await db('stages')
  .where({ user_id })
  .andWhere('position', '>=', newPos)
  .increment('position', 1);
const [stage] = await db('stages').insert({...}).returning('*');

// AFTER
return withTransaction(db, undefined, async (trx) => {
  await shiftUp({ trx, table: 'stages', scope: { user_id }, fromPos: newPos });
  const [stage] = await trx('stages').insert({...}).returning('*');
  return stage;
});
```

```ts
// cardService.moveCard (cross-stage) — BEFORE
await trx('cards').where({user_id, stage_id: oldStage}).andWhere('position', '>', oldPos).decrement('position', 1);
await trx('cards').where({user_id, stage_id: newStage}).andWhere('position', '>=', newPos).increment('position', 1);
await trx('cards').where({id}).update({ stage_id: newStage, position: newPos });

// AFTER
await shiftDown({ trx, table: 'cards', scope: { user_id, stage_id: oldStage }, fromPos: oldPos });
await shiftUp  ({ trx, table: 'cards', scope: { user_id, stage_id: newStage }, fromPos: newPos });
await trx('cards').where({id}).update({ stage_id: newStage, position: newPos });
```

```ts
// stageService.deleteStage — fixes the transaction bug
return withTransaction(db, undefined, async (trx) => {
  const stage = await trx('stages').where({id, user_id}).first();
  // ...existing cascade logic, now all inside one trx...
  await shiftDown({ trx, table: 'stages', scope: { user_id }, fromPos: stage.position });
});
```

## Dependency Strategy

**Category: In-process.** No new external dependencies. Three pure async functions wrapping `db.where(...).increment/decrement(...)` calls. No port/adapter required — the helper module is merged directly into the service layer.

- **Trx handling**: every shift function REQUIRES a `trx`. `withTransaction(knex, trx?, fn)` is the single place where optional-trx exists. Callers that already have a trx (gmailSync writing through `cardService.moveCard`) pass it through; callers that don't get a fresh one opened for them.
- **Null positions** (`todo_items`): the helper does NOT filter `position IS NOT NULL`. Callers add `.whereNotNull('position')` when needed. Pushing null-handling into the helper would force every other caller to think about a concept that doesn't apply to them.
- **Typed table names**: `table: 'stages' | 'cards' | 'todo_items'` literal union prevents typos at compile time. Adding a 4th ordered table is a 1-line change to the union.

### New boundary tests to write
A single file `backend/src/util/positions.test.ts` covering:
1. `shiftUp` shifts only rows in scope; rows in other scopes untouched.
2. `shiftDown` boundary: `>` (strict) is correct — the row at `fromPos` itself is not shifted.
3. `shiftUp` boundary: `>=` (inclusive) — the row at `fromPos` IS shifted.
4. `shiftUp`/`shiftDown` with `toPos` constrains the range correctly.
5. `renumber` produces 0..N positions matching the input order.
6. `renumber` survives the `UNIQUE(scope, position)` constraint on `cards`.
7. `withTransaction` joins an existing trx (no new connection); opens its own when trx omitted.
8. `withTransaction` rolls back all writes if the inner function throws.

### Tests that gain confidence (don't need to be rewritten)
The existing service test files (`stages.test.ts`, `cards.test.ts`, `todos.test.ts`) keep their HTTP-level coverage. After the migration, ordering correctness is verified once at the boundary instead of implicitly across three test suites.

### Test environment
Existing Postgres test setup. No new infrastructure.

### What the module owns
- Integer position arithmetic on Postgres tables with a `position INT` column
- Transaction joining (`withTransaction`)
- A guarantee: outside this module, no service should hand-write `where('position', op, x).increment/decrement(...)`

### What it hides
- Direction of shift (`>=` vs `>`, `+1` vs `-1`)
- The two-phase renumber to avoid unique-constraint conflicts
- Whether a trx is opened or joined

### What it exposes
- Three shift functions + one trx helper. Plain async free functions. No factory, no class, no DI.

### Suggested PR sequence

**PR 1 — bug fix only (urgent, ~15 lines):**
Wrap `stageService.deleteStage` in `db.transaction(...)`. No helper extraction. Ships immediately as a correctness fix; reviewers can verify the transactional fix in isolation without also reasoning about a new module.

**PR 2 — introduce the helpers + convert the first caller (~80 lines):**
Add `backend/src/util/positions.ts` with `shiftUp`, `shiftDown`, `renumber`, `withTransaction`. Convert `stageService.createStage` as the pilot. Add the boundary test file.

**PR 3 — sweep the remaining 6 sites (mechanical):**
Convert `updateStage`, `reorderStages`, `deleteStage` (now using `shiftDown` instead of the inline shift), `moveCard`, `deleteCard`, `deleteTodo`, `reorderTodos`. One commit per service file. Each commit's diff is a textbook before/after.

### What this does NOT try to do
- Does not change the data model (`position INT` stays).
- Does not improve algorithmic complexity (every action is still O(N) in row writes — see separate issue for the fractional-indexing migration).
- Does not absorb domain logic (priority inheritance in `reorderTodos`, same-stage shortcut in `moveCard`) into the helper.
- Does not introduce a class, factory, or DSL — three free functions are the whole interface.

---

tomerp20:
## Summary
- Introduces `backend/src/util/positions.ts` with four exported helpers: `shiftUp`, `shiftDown`, `renumber`, and `withTransaction`
- Fixes the latent transaction bug in `deleteStage` — all 5 cascade steps are now wrapped in a single atomic transaction (previously unguarded)
- Converts all 7 hand-written position-shift sites across `stageService`, `cardService`, and `todoService` to use the helpers

## Acceptance criteria
- [x] `backend/src/util/positions.ts` exports `shiftUp`, `shiftDown`, `renumber`, and `withTransaction`
- [x] `stageService.deleteStage` wraps its entire cascade in a single transaction (transaction bug fixed)
- [x] All 7 duplication sites converted to use helpers; no service hand-writes `.where('position', op, x).increment/decrement(...)` directly
- [x] `backend/tests/integration/positions.integration.test.ts` covers boundary semantics, scope isolation, UNIQUE constraint survival, and `withTransaction` join/rollback
- [x] All existing unit tests pass (`npm test` — 83/83)
- [x] TypeScript compiles clean (`npx tsc --noEmit`)

## Changes
- **New**: `backend/src/util/positions.ts` — `shiftUp` (`>=` fromPos), `shiftDown` (`>` fromPos), `renumber` (two-phase to avoid UNIQUE conflicts), `withTransaction` (joins existing trx or opens its own)
- **Fixed**: `stageService.deleteStage` — 5 dependent writes now in one transaction; also removed dead `maxPos` query
- **Fixed**: `cardService.deleteCard` — delete + position shift are now atomic
- **Converted**: `stageService.{createStage,updateStage,reorderStages}`, `cardService.{moveCard,deleteCard}`, `todoService.{deleteTodo,reorderTodos}`
- **Tests**: added proxy `mockTransaction.mockImplementation` in unit test `beforeEach` hooks so `withTransaction` callbacks execute under mock

## Related
Closes #164

### DIFF
--- a/backend/src/services/cardService.ts
+++ b/backend/src/services/cardService.ts
@@ -1,6 +1,7 @@
 import { Knex } from 'knex';
 import db from '../config/database';
 import { AppError } from '../middleware/errorHandler';
+import { shiftUp, shiftDown, withTransaction } from '../util/positions';
 
 export interface CardFilters {
   stage?: string;
@@ -405,20 +406,8 @@ export const cardService = {
     const oldPosition = card.position;
 
     const performMove = async (t: Knex.Transaction) => {
-      // Remove from old position: shift cards in old stage down to fill gap
-      await t('cards')
-        .where({ user_id: userId, stage_id: oldStageId })
-        .andWhere('position', '>', oldPosition)
-        .decrement('position', 1);
-
-      // Make room in new position: shift cards in target stage up
-      await t('cards')
-        .where({ user_id: userId, stage_id: stageId })
-        .andWhere('position', '>=', position)
-        .andWhere('id', '!=', cardId)
-        .increment('position', 1);
-
-      // Move the card
+      await shiftDown({ trx: t, table: 'cards', scope: { user_id: userId, stage_id: oldStageId }, fromPos: oldPosition });
+      await shiftUp({ trx: t, table: 'cards', scope: { user_id: userId, stage_id: stageId }, fromPos: position });
       await t('cards')
         .where({ id: cardId })
         .update({
@@ -470,14 +459,10 @@ export const cardService = {
       throw new AppError('Card not found', 404, 'ERR_NOT_FOUND');
     }
 
-    // Delete card (cascade deletes activities)
-    await db('cards').where({ id: cardId, user_id: userId }).del();
-
-    // Shift remaining cards in the stage down to fill gap
-    await db('cards')
-      .where({ user_id: userId, stage_id: card.stage_id })
-      .andWhere('position', '>', card.position)
-      .decrement('position', 1);
+    await withTransaction(db, undefined, async (trx) => {
+      await trx('cards').where({ id: cardId, user_id: userId }).del();
+      await shiftDown({ trx, table: 'cards', scope: { user_id: userId, stage_id: card.stage_id }, fromPos: card.position });
+    });
   },
 
   async addNote(cardId: string, userId: string, note: string) {
--- a/backend/src/util/positions.ts
+++ b/backend/src/util/positions.ts
@@ -1,0 +1,60 @@
+import { Knex } from 'knex';
+
+export type Scope = Record<string, string | number>;
+
+export interface ShiftArgs {
+  trx: Knex.Transaction;
+  table: 'stages' | 'cards' | 'todo_items';
+  scope: Scope;
+  fromPos: number;
+  toPos?: number;
+}
+
+export interface RenumberArgs {
+  trx: Knex.Transaction;
+  table: 'stages' | 'cards' | 'todo_items';
+  scope: Scope;
+  orderedIds: string[];
+  idColumn?: string;
+}
+
+export async function shiftUp(args: ShiftArgs): Promise<number> {
+  const { trx, table, scope, fromPos, toPos } = args;
+  let q = trx(table).where(scope).andWhere('position', '>=', fromPos);
+  if (toPos !== undefined) q = q.andWhere('position', '<=', toPos);
+  return q.increment('position', 1);
+}
+
+export async function shiftDown(args: ShiftArgs): Promise<number> {
+  const { trx, table, scope, fromPos, toPos } = args;
+  let q = trx(table).where(scope).andWhere('position', '>', fromPos);
+  if (toPos !== undefined) q = q.andWhere('position', '<=', toPos);
+  return q.decrement('position', 1);
+}
+
+// Two-phase to avoid UNIQUE(scope, position) constraint violations within a transaction.
+// Phase 1 moves rows to unique negative sentinels; phase 2 assigns final 0..N-1 positions.
+export async function renumber(args: RenumberArgs): Promise<void> {
+  const { trx, table, scope, orderedIds, idColumn = 'id' } = args;
+  const N = orderedIds.length;
+  for (let i = 0; i < N; i++) {
+    await trx(table)
+      .where({ ...scope, [idColumn]: orderedIds[i] })
+      .update({ position: -(N + i + 1) });
+  }
+  for (let i = 0; i < N; i++) {
+    await trx(table)
+      .where({ ...scope, [idColumn]: orderedIds[i] })
+      .update({ position: i });
+  }
+}
+
+// Joins caller's trx when present, otherwise opens its own.
+export async function withTransaction<T>(
+  knex: Knex,
+  trx: Knex.Transaction | undefined,
+  fn: (t: Knex.Transaction) => Promise<T>,
+): Promise<T> {
+  if (trx) return fn(trx);
+  return knex.transaction(fn);
+}
--- a/backend/src/util/positions.ts
+++ b/backend/src/util/positions.ts
@@ -1,5 +1,11 @@
 import { Knex } from 'knex';
 
+// CONTRACT: tables passed to renumber/shiftUp/shiftDown must NOT have a
+// non-deferrable UNIQUE (scope, position) index. The helpers transiently
+// duplicate positions inside a transaction. If you add such an index, ensure
+// it is DEFERRABLE INITIALLY DEFERRED (see migration 010 for the stages
+// table — cards and todo_items currently have no such constraint).
+
 export type Scope = Record<string, string | number>;
 
 export interface ShiftArgs {
@@ -18,22 +24,24 @@ export interface RenumberArgs {
   idColumn?: string;
 }
 
-export async function shiftUp(args: ShiftArgs): Promise<number> {
+export async function shiftUp(args: ShiftArgs): Promise<void> {
   const { trx, table, scope, fromPos, toPos } = args;
   let q = trx(table).where(scope).andWhere('position', '>=', fromPos);
   if (toPos !== undefined) q = q.andWhere('position', '<=', toPos);
-  return q.increment('position', 1);
+  await q.increment('position', 1);
 }
 
-export async function shiftDown(args: ShiftArgs): Promise<number> {
+export async function shiftDown(args: ShiftArgs): Promise<void> {
   const { trx, table, scope, fromPos, toPos } = args;
   let q = trx(table).where(scope).andWhere('position', '>', fromPos);
   if (toPos !== undefined) q = q.andWhere('position', '<=', toPos);
-  return q.decrement('position', 1);
+  await q.decrement('position', 1);
 }
 
 // Two-phase to avoid UNIQUE(scope, position) constraint violations within a transaction.
 // Phase 1 moves rows to unique negative sentinels; phase 2 assigns final 0..N-1 positions.
+// TODO(scale): collapse each phase to a single UPDATE ... FROM (VALUES (id, pos), ...)
+// once collections grow beyond ~100 items — current 2N round-trips are fine for ≤100.
 export async function renumber(args: RenumberArgs): Promise<void> {
   const { trx, table, scope, orderedIds, idColumn = 'id' } = args;
   const N = orderedIds.length;
--- a/backend/src/services/stageService.ts
+++ b/backend/src/services/stageService.ts
@@ -1,5 +1,6 @@
 import db from '../config/database';
 import { AppError } from '../middleware/errorHandler';
+import { shiftUp, shiftDown, renumber, withTransaction } from '../util/positions';
 
 export interface StageData {
   name: string;
@@ -17,21 +18,13 @@ export const stageService = {
   },
 
   async createStage(userId: string, data: StageData) {
-    // Shift existing stages at or after this position up by 1
-    await db('stages')
-      .where({ user_id: userId })
-      .andWhere('position', '>=', data.position)
-      .increment('position', 1);
-
-    const [stage] = await db('stages')
-      .insert({
-        user_id: userId,
-        name: data.name,
-        position: data.position,
-      })
-      .returning('*');
-
-    return stage;
+    return withTransaction(db, undefined, async (trx) => {
+      await shiftUp({ trx, table: 'stages', scope: { user_id: userId }, fromPos: data.position });
+      const [stage] = await trx('stages')
+        .insert({ user_id: userId, name: data.name, position: data.position })
+        .returning('*');
+      return stage;
+    });
   },
 
   async updateStage(stageId: string, userId: string, data: Partial<StageData>) {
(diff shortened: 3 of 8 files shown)


tomerp20 (commented):
## Backend Code Review — STANDARD (lean toward HIGH-RISK due to transaction/migration interplay)

Reviewed: `backend/src/util/positions.ts` (new), `cardService.ts`, `stageService.ts`, `todoService.ts`, and the three test files.

**ESLint guardrail:** Backend ESLint is configured and ran. The new file `src/util/positions.ts` is **clean** (0 errors). The three service files report `@typescript-eslint/no-unsafe-*` errors, but these are **pre-existing on `main`** (verified by linting `git show main:...stageService.ts`). They are not introduced by this PR. Flagging as a separate cleanup, not a blocker.

---

### Issues by Severity

#### CRITICAL — none

#### HIGH

**[HIGH] Verify `DEFERRABLE INITIALLY DEFERRED` covers every table the helpers temporarily violate (`renumber`)**

The `renumber` helper relies on a two-phase update (sentinel negatives → final 0..N-1). Phase 1 is also reached by `shiftUp`/`shiftDown` followed by an update that briefly produces duplicate positions inside the same transaction. This requires the `UNIQUE (scope, position)` constraint to be `DEFERRABLE INITIALLY DEFERRED`, otherwise PostgreSQL will reject phase 1 immediately on the first duplicate.

I verified the migrations:

| Table | Unique constraint on `(scope, position)`? | Deferrable? |
|-------|--------------------------------------------|-------------|
| `stages` | yes — `stages_user_id_position_unique` | yes (010_defer_stages_position_unique) |
| `cards` | **no unique constraint exists** (003_create_cards.ts) | n/a |
| `todo_items` | **no unique constraint exists** (008_create_todo_items.ts) | n/a |

**Conclusion:** the helpers are *safe today*, because only `stages` has the constraint and it is already deferred. **However**, this becomes a footgun: any future migration that adds `UNIQUE (user_id, stage_id, position)` to `cards` or `UNIQUE (user_id, position)` to `todo_items` will silently break these flows in production, because nothing in code or tests will catch the regression — the integration test fixtures only assert final state, not constraint timing.

**Fix (recommended, no blocker):** Add a comment to `positions.ts` documenting the deferrable requirement, and a guard in any future migration that adds such a constraint. Example:

```ts
// positions.ts — at the top of the file
//
// CONTRACT: tables passed to `renumber`/`shiftUp`/`shiftDown` MUST NOT have a
// non-deferrable UNIQUE (scope, position) index. The helpers transiently break
// uniqueness within a transaction. If you ever add such an index, ensure it is
// DEFERRABLE INITIALLY DEFERRED (see migration 010_defer_stages_position_unique).
```

---

**[HIGH] Missing test for same-stage `moveCard` (intra-column drag)**

In `cardService.moveCard` (lines 408–419), when the source stage equals the target stage, the new code path is:

```ts
await shiftDown({ trx: t, table: 'cards', scope: { user_id, stage_id: oldStageId }, fromPos: oldPosition });
await shiftUp({ trx: t, table: 'cards', scope: { user_id, stage_id: stageId },    fromPos: position });
await t('cards').where({ id: cardId }).update({ stage_id, position, ... });
```

When `oldStageId === stageId` and the card moves from `oldPosition` to `position`, the moving card itself participates in **both** shifts. The previous code had an explicit `.andWhere('id', '!=', cardId)` exclusion on the new-stage shiftUp; the new code dropped it. The **final state is correct** because the third UPDATE overwrites the card’s position — but the card transiently lands on an unexpected value, and nothing in the test suite locks this in.

Without the `stages` unique constraint on `cards` this is harmless today, but it is exactly the class of behaviour that breaks once someone adds a unique index. More importantly, the test gap means a future refactor could regress the final state.

**Fix:** Add an integration test that exercises drag-within-same-stage. Example case: a column with positions `[0, 1, 2, 3, 4]`, move position 4 → 0, assert the final array is `[movedCard, 0, 1, 2, 3]` and positions are `[0, 1, 2, 3, 4]` with the right IDs in the right slots.

---

#### MEDIUM

**[MEDIUM] `shiftUp` / `shiftDown` return `Promise<number>` that no caller uses**

`positions.ts` lines 21 and 28: the helpers return the row count from `knex.increment/decrement`, but none of the 7 call sites consume the return value. Returning `Promise<void>` would be cleaner and match `renumber`'s signature.

**Fix:**
```ts
export async function shiftUp(args: ShiftArgs): Promise<void> {
  const { trx, table, scope, fromPos, toPos } = args;
  let q = trx(table).where(scope).andWhere('position', '>=', fromPos);
  if (toPos !== undefined) q = q.andWhere('position', '<=', toPos);
  await q.increment('position', 1);
}

// same for shiftDown
```

---

**[MEDIUM] `renumber` performs 2N sequential round-trips**

`positions.ts` lines 40–49: phase 1 and phase 2 each do `N` separate UPDATE statements over the network. For the current scope (≤ ~50 stages, ≤ a few hundred cards per stage) this is fine. At scale (e.g. a future bulk reorder of thousands of todos), this becomes a latency hot spot. Worth a comment, not a blocker today.

**Fix (future-proofing):** A single `UPDATE ... FROM (VALUES ...)` would collapse each phase to one round trip. Leave a TODO.

```ts
// TODO(scale): collapse each phase to a single UPDATE ... FROM (VALUES (id, pos), ...)
// when collections grow beyond ~100 items.
```

---

**[MEDIUM] PR scope — three sequential PRs collapsed into one**

The RFC referenced by this PR recommended landing the work as three sequential PRs (urgent bug fix → helpers + pilot → sweep across all services). The agent shipped all of it in one. The diff is reviewable, but this **violates the agreed staged-rollout plan**: if any one piece needs to be reverted, the others come with it. Flagging as a **process note**, not a blocker on the code itself.

**Recommendation:** for the next refactor of this size, honour the RFC's staged plan — small, independently revertible PRs reduce risk.

---

#### LOW

**[LOW] `withTransaction` shadowing the `trx` name**

`positions.ts` line 53–60: the function calls its inner parameter `t` once you join an outer transaction (`return fn(trx)` passes through), but `cardService.moveCard` already has its own `performMove(t)` local. No bug, just a style note — using `trx` consistently across the module would make grep easier.

---

### Testing Gaps

- **Same-stage `moveCard`** — already called out in [HIGH] above.
- **`deleteStage` with cards present** — the integration test file has good coverage for `renumber` directly, but I don't see an integration-level test that exercises the full `deleteStage` flow (delete a stage that has cards → assert cards moved to fallback with positions `[0..N-1]` and the deleted stage's slot is gone). The existing unit test mocks every step; an integration test against the real DB would lock in the transactional behaviour.
- **`withTransaction` rollback inside a real service call** — `positions.integration.test.ts` covers `withTransaction` directly. Worth adding one test where `cardService.deleteCard` is called and a forced error mid-transaction leaves the cards table unchanged.

---

### Summary

| Category | Count |
|----------|-------|
| Critical | 0 |
| High | 2 |
| Medium | 3 |
| Low | 1 |
| Testing gaps | 3 |

The PR is solid work — the helpers are typed cleanly, the integration suite covers the right invariants for the helpers themselves, and the `deleteStage` transaction fix is a clear improvement. The HIGH items are both about **risk under future changes** rather than bugs today; the MEDIUM items are polish and process. Recommend addressing the same-stage `moveCard` test and the `positions.ts` contract comment before merge.

tomerp20 (commented):
## Security Review

Scanned all 8 changed files for OWASP Top 10, secrets, authentication/authorization, injection, and data-leakage patterns. Files reviewed:

- `backend/src/util/positions.ts` (new)
- `backend/src/services/{card,stage,todo}Service.ts`
- `backend/tests/{cards,stages,todos}.test.ts`
- `backend/tests/integration/positions.integration.test.ts`

---

### Findings

**No security issues found.**

Specifically checked:

| Check | Result |
|-------|--------|
| Secrets / API keys / tokens in code | none |
| Knex `.raw()` with string interpolation | none — `positions.ts` uses parameterised builder calls; the only `.raw()` in the diff is `db.raw('TRUNCATE TABLE …')` in the test setup, which uses a static literal |
| SQL injection via dynamic table/column names | The `table` param in `ShiftArgs`/`RenumberArgs` is constrained to a string-literal union (`'stages' \| 'cards' \| 'todo_items'`), and `idColumn` defaults to `'id'`. No user-controlled input flows into the table or column position |
| Cross-user data access | Every helper call site includes `user_id` in `scope`. `cardService.moveCard` and `deleteCard` filter by `{ user_id, stage_id }` consistently. No tenancy bypass |
| Missing authentication on routes | This PR does not modify routes, only services. Existing route-level auth is untouched |
| Stack traces / internal paths leaked in responses | No new error responses introduced; existing `AppError` pattern preserved |
| Sensitive data in logs | No new `console.log` / `logger.*` calls in the diff |
| Race conditions exploitable across users | The new transaction boundaries are **per-user** (every shift is scoped by `user_id`). The deferrable unique constraint on `stages` prevents constraint races within a transaction. A concurrent multi-user reorder cannot leak rows across `user_id` because every query carries the scope |
| Dependency additions | None — no `package.json` changes |

---

### Notes (not findings)

- The `renumber` helper writes negative sentinel positions during phase 1. These values are never exposed to API responses (positions are immediately overwritten in phase 2 before the transaction commits), so they do not appear in the wire format. Worth flagging only because: if a future API path ever queries this table mid-transaction from a different connection, those connections would see the committed final state, not the sentinels — so no information disclosure.
- `withTransaction` correctly propagates exceptions, which means the existing global error handler will handle them without leaking internals.

---

### Summary

**0 security issues. PR is clean from a security standpoint.**

backend/src/util/positions.ts
```diff
@@ -0,0 +1,60 @@
+import { Knex } from 'knex';
```
tomerp20, line 1: **[HIGH] — Document the DEFERRABLE unique-index contract for these helpers**

`renumber` (and the shift helpers when followed by an update) transiently break `UNIQUE (scope, position)` within a transaction. Today this is safe — only `stages` has such a constraint, and migration 010 made it `DEFERRABLE INITIALLY DEFERRED`. `cards` and `todo_items` have no such constraint (003_create_cards.ts, 008_create_todo_items.ts), so the helpers work.

But this is a latent footgun: any future migration that adds `UNIQUE (user_id, stage_id, position)` to `cards` (or `UNIQUE (user_id, position)` to `todo_items`) without `DEFERRABLE INITIALLY DEFERRED` will silently break `moveCard`, `deleteCard`, `deleteTodo`, and the renumber call inside `deleteStage` — and the integration tests will not catch it because they only assert final state.

**Fix:** add a contract comment at the top of `positions.ts`:

```ts
// CONTRACT: tables passed to renumber/shiftUp/shiftDown must NOT have a
// non-deferrable UNIQUE (scope, position) index. The helpers transiently
// duplicate positions inside a transaction. If you add such an index, ensure
// it is DEFERRABLE INITIALLY DEFERRED (see migration 010).
```

backend/src/services/cardService.ts
```diff
@@ -405,20 +406,8 @@ export const cardService = {
...
-      // Make room in new position: shift cards in target stage up
-      await t('cards')
-        .where({ user_id: userId, stage_id: stageId })
-        .andWhere('position', '>=', position)
-        .andWhere('id', '!=', cardId)
-        .increment('position', 1);
-
-      // Move the card
+      await shiftDown({ trx: t, table: 'cards', scope: { user_id: userId, stage_id: oldStageId }, fromPos: oldPosition });
+      await shiftUp({ trx: t, table: 'cards', scope: { user_id: userId, stage_id: stageId }, fromPos: position });
```
tomerp20, line 410: **[HIGH] — Missing test: same-stage `moveCard` (drag-within-column)**

When `oldStageId === stageId`, both `shiftDown` (line 409) and `shiftUp` (line 410) match the moving card itself. The previous code excluded it via `.andWhere("id", "!=", cardId)`; the new code dropped that.

Final state is still correct because the UPDATE on line 411 overwrites the card position, but the card transiently lands on an unexpected slot. No constraint blows up today (cards has no UNIQUE (scope, position) index), but any future migration that adds one would break this exact path.

More importantly: the test suite never exercises drag-within-same-stage, so a future refactor of the helpers could regress final-state correctness silently.

**Fix:** add an integration test in `positions.integration.test.ts` (or a dedicated `cardService.integration.test.ts`):

```ts
it("moveCard within the same stage produces the right final positions", async () => {
  const s = await insertStage(0);
  const c0 = await insertCard(s, 0);
  const c1 = await insertCard(s, 1);
  const c2 = await insertCard(s, 2);
  const c3 = await insertCard(s, 3);
  const c4 = await insertCard(s, 4);

  await cardService.moveCard(c4, userId, { stageId: s, position: 0 });

  const rows = await getDb()("cards").where({ stage_id: s }).orderBy("position");
  expect(rows.map(r => r.id)).toEqual([c4, c0, c1, c2, c3]);
  expect(rows.map(r => r.position)).toEqual([0, 1, 2, 3, 4]);
});
```

backend/src/util/positions.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export interface RenumberArgs {
+  trx: Knex.Transaction;
+  table: 'stages' | 'cards' | 'todo_items';
+  scope: Scope;
+  orderedIds: string[];
+  idColumn?: string;
+}
+
+export async function shiftUp(args: ShiftArgs): Promise<number> {
```
tomerp20: **[MEDIUM] — Return `Promise<void>` instead of `Promise<number>`**

`shiftUp` and `shiftDown` return the row count from `knex.increment/decrement`, but **no caller uses it** (verified across all 7 sites in `cardService`, `stageService`, `todoService`, and the integration tests). The signature implies the count is meaningful — and `renumber` already returns `Promise<void>` for the same shape of operation. Aligning the two would make the API consistent.

**Fix:**

```ts
export async function shiftUp(args: ShiftArgs): Promise<void> {
  const { trx, table, scope, fromPos, toPos } = args;
  let q = trx(table).where(scope).andWhere("position", ">=", fromPos);
  if (toPos !== undefined) q = q.andWhere("position", "<=", toPos);
  await q.increment("position", 1);
}
// same for shiftDown
```

backend/src/util/positions.ts
```diff
@@ -0,0 +1,60 @@
...
+  if (toPos !== undefined) q = q.andWhere('position', '<=', toPos);
+  return q.decrement('position', 1);
+}
+
+// Two-phase to avoid UNIQUE(scope, position) constraint violations within a transaction.
+// Phase 1 moves rows to unique negative sentinels; phase 2 assigns final 0..N-1 positions.
+export async function renumber(args: RenumberArgs): Promise<void> {
+  const { trx, table, scope, orderedIds, idColumn = 'id' } = args;
+  const N = orderedIds.length;
+  for (let i = 0; i < N; i++) {
```
tomerp20, line 48: **[MEDIUM] — `renumber` does 2N round-trips; fine for now, leave a scale-out TODO**

Phase 1 and phase 2 each issue `N` separate UPDATE statements over the wire. For the current scopes (≤ ~50 stages, ≤ a few hundred cards per stage) this is well within budget. At scale (e.g. a future bulk reorder of thousands of todos), the round-trip cost dominates.

Not a blocker for this PR — flag for future tuning.

**Fix (when needed):**

```ts
// TODO(scale): collapse each phase to a single UPDATE ... FROM (VALUES (id, pos), ...)
// once collections grow beyond ~100 items.
```

Each phase then becomes a single round-trip:

```sql
UPDATE stages
SET position = v.position
FROM (VALUES (uuid1, 0), (uuid2, 1), ...) AS v(id, position)
WHERE stages.id = v.id AND stages.user_id = $userId;
```

tomerp20 (commented):
## Code Review Complete

| Reviewer | Status | Findings | Fixed | Skipped |
|----------|--------|----------|-------|---------|
| Security (Orchestrator) | Done | 0 | n/a | 0 |
| Backend Reviewer | Done | 6 (2 HIGH, 3 MEDIUM, 1 LOW) | 4 | 2 (see below) |

### What was fixed in commit `9b9e242`

- **[HIGH] DEFERRABLE contract** — Added a top-of-file comment in `positions.ts` documenting that callers must not introduce a non-deferrable `UNIQUE (scope, position)` constraint on any table these helpers touch.
- **[HIGH] Same-stage `moveCard` test** — Added two integration tests in `positions.integration.test.ts` covering drag-down-within-column (position 4 → 0) and drag-up-within-column (position 0 → 3). Both pass.
- **[MEDIUM] Return type cleanup** — `shiftUp` and `shiftDown` now return `Promise<void>` instead of `Promise<number>`, matching `renumber`.
- **[MEDIUM] Scale-out TODO** — Added a `TODO(scale)` comment on `renumber` flagging the 2N round-trip pattern for future tuning.

### Skipped fixes (require human attention)

- **PR scoping (process note)** — The RFC recommended landing this work as three sequential PRs (urgent `deleteStage` bug → helpers + pilot → sweep). The agent shipped all of it in one. The diff is reviewable, but if any one piece needs to be reverted, the others come with it. **Not a code blocker. Flag for the next refactor of this size.**
- **[LOW] `withTransaction` parameter naming style** — left as-is; cosmetic only.

### Pre-existing ESLint baseline noise

The three service files (`cardService`, `stageService`, `todoService`) report `@typescript-eslint/no-unsafe-*` errors. Verified against `main` — these are pre-existing (26 errors on `stageService` in main alone). The new `positions.ts` file is clean. Recommend a separate cleanup PR to add explicit Knex result typing across the services.

### Verification

- Unit tests: 83/83 passing on the fix commit
- TypeScript: `tsc --noEmit` clean
- Fix commit `9b9e242` confirmed on the PR head

### Next steps

- [ ] Run integration tests against the dedicated Neon DB (`npm run test:e2e`) to validate the new same-stage moveCard tests against real Postgres before merge
- [ ] Human review of skipped process note above

PR #169, opened 2026-05-19
perf(gmailSync): parallelize email classification with bounded concurrency (p-limit)

### LINKED ISSUE #162: perf(gmailSync): parallelize email classification with bounded concurrency (p-limit)
## Parent PRD

#145

## What to build

Replace the sequential `for…of` classification loop in `syncUserGmailWith` with a bounded-concurrency `Promise.all` so multiple emails can hit the LLM classifier in parallel without blowing through provider rate limits.

The per-email DB-writing loop (which depends on the per-user `pg_advisory_xact_lock` introduced in #145) stays sequential — only the external-I/O classification stage is parallelized.

## 📝 Description & Root Cause

The classification stage at `backend/src/services/gmailSync.ts:176-187` is a sequential `for…of` with an `await` inside the loop body:

\`\`\`ts
const classified: ClassifiedEmail[] = [];
for (const email of unprocessed) {
  summary.scanned++;
  try {
    const classification = await deps.classifier.classify(email);
    classified.push({ email, classification });
  } catch (err) {
    logger.error('Email classification failed', { ... });
    classified.push({ email, classification: null });
    summary.errors++;
  }
}
\`\`\`

`deps.classifier.classify` calls into an LLM (see `emailClassifier.ts`). At ~1–3 s of network + inference latency per email, an inbox with 50 unread messages blocks the sync for upwards of a minute, all while holding a single in-flight request. The loop is the dominant contributor to `summary.durationMs`.

A naïve `Promise.all(unprocessed.map(classify))` would fan out unboundedly and trip provider rate limits (or exhaust sockets) on large initial syncs — so the fix needs an **explicit concurrency cap**.

## 🛠️ Proposed Solution

Introduce a small concurrency limiter via `p-limit`. Keep per-email error handling exactly as-is so a single failure doesn't poison the batch.

\`\`\`ts
import pLimit from 'p-limit';

// Tune via env var so ops can dial back if the LLM provider rate-limits us.
const CLASSIFY_CONCURRENCY = Number(process.env.GMAIL_CLASSIFY_CONCURRENCY ?? 5);

const limit = pLimit(CLASSIFY_CONCURRENCY);

const classified: ClassifiedEmail[] = await Promise.all(
  unprocessed.map(email =>
    limit(async () => {
      try {
        const classification = await deps.classifier.classify(email);
        return { email, classification };
      } catch (err) {
        logger.error('Email classification failed', {
          userId,
          messageId: email.messageId,
          error: safeError(err),
        });
        return { email, classification: null as EmailClassification | null };
      }
    }),
  ),
);

// Counters derived from the resolved array — deterministic under parallel execution.
summary.scanned = classified.length;
summary.errors = classified.filter(c => c.classification === null).length;
\`\`\`

**Notes for the implementer:**
- **Do not** move parallelism into the DB-writing loop below — that loop relies on a per-user `pg_advisory_xact_lock` and must remain sequential.
- The old `summary.scanned++` / `summary.errors++` mutations were safe under sequential execution. After parallelization they become race-prone in spirit; derive them from the result array.
- A hand-rolled batching helper of ~10 lines is acceptable if `p-limit` is undesirable:

  \`\`\`ts
  async function mapWithConcurrency<T, R>(
    items: T[], limit: number, fn: (item: T) => Promise<R>,
  ): Promise<R[]> {
    const results: R[] = new Array(items.length);
    let cursor = 0;
    const workers = Array.from({ length: Math.min(limit, items.length) }, async () => {
      while (true) {
        const i = cursor++;
        if (i >= items.length) return;
        results[i] = await fn(items[i]);
      }
    });
    await Promise.all(workers);
    return results;
  }
  \`\`\`

## ✅ Acceptance criteria

- [ ] Classification loop replaced with a bounded-concurrency implementation; default cap `5`, overridable via `GMAIL_CLASSIFY_CONCURRENCY`
- [ ] `p-limit` added to `backend/package.json` (or the hand-rolled helper landed with unit tests)
- [ ] Unit test: a `ClassifierPort` stub with a 200 ms artificial delay classifies 20 emails in under ~2 s (vs. ~4 s sequential), proving parallelism is actually engaged
- [ ] Unit test: per-email error isolation — a stub that throws on every 3rd call still produces a fully-populated `classified` array with `classification: null` for the failures
- [ ] `summary.scanned` and `summary.errors` are deterministic regardless of resolution order
- [ ] Existing `backend/tests/gmailSync.test.ts` cases pass without modification beyond necessary stub-timing tolerances

## Blocked by

None — can start immediately. Independent of the other two follow-ups.

## User stories addressed

N/A — performance hardening of the module introduced in #145. Addresses the `summary.durationMs` blow-up on large initial syncs that the original PRD did not constrain.

tomerp20:
## Summary
- Replace the sequential `for…of` classification loop in `syncUserGmailWith` with a bounded-concurrency parallel implementation using a hand-rolled `mapWithConcurrency` helper
- Default concurrency cap of 5, overridable via `GMAIL_CLASSIFY_CONCURRENCY` env var
- DB-writing loop remains sequential — the `pg_advisory_xact_lock` and in-batch dedup invariants are untouched

## Acceptance criteria
- [x] Classification loop replaced with a bounded-concurrency implementation; default cap `5`, overridable via `GMAIL_CLASSIFY_CONCURRENCY`
- [x] Hand-rolled `mapWithConcurrency` helper (~12 lines) with unit tests (p-limit not added — ESM-only at current version; hand-rolled avoids the CJS compatibility issue)
- [x] Unit test: 20 emails with 200 ms delay classify in 815 ms (well under ~2 s), proving parallelism is engaged
- [x] Unit test: classifier throwing on every 3rd call still produces fully-populated `classified` array with `classification: null` for failures
- [x] `summary.scanned` and `summary.errors` derived from the result array — deterministic regardless of resolution order
- [x] All 27 existing `gmailSync.test.ts` cases pass without modification

## Related
Closes #162

### DIFF
--- a/backend/src/services/gmailSync.ts
+++ b/backend/src/services/gmailSync.ts
@@ -54,6 +54,7 @@ export interface SyncSummary {
 
 const MAX_VARCHAR_255 = 255;
 const CONFIDENCE_THRESHOLD = 0.9;
+const CLASSIFY_CONCURRENCY = Number(process.env.GMAIL_CLASSIFY_CONCURRENCY ?? 5);
 
 type ProcessedAction =
   | 'receipt_created'
@@ -75,6 +76,24 @@ interface ClassifiedEmail {
 type TrxResult = { newCard: { id: string; company_name: string; role_title: string } } | null;
 
 // ── Helpers ───────────────────────────────────────────────────────────────────
+
+async function mapWithConcurrency<T, R>(
+  items: T[],
+  limit: number,
+  fn: (item: T) => Promise<R>,
+): Promise<R[]> {
+  const results: R[] = new Array(items.length);
+  let cursor = 0;
+  const workers = Array.from({ length: Math.min(limit, items.length) }, async () => {
+    while (true) {
+      const i = cursor++;
+      if (i >= items.length) return;
+      results[i] = await fn(items[i]);
+    }
+  });
+  await Promise.all(workers);
+  return results;
+}
 
 // Errors from Gaxios/Axios carry a `config` field that includes the request
 // headers — including the Bearer access token. Logging the raw error object
@@ -175,18 +194,25 @@ export async function syncUserGmailWith(
   const unprocessed = emails.filter(e => !processedSet.has(e.messageId));
 
   // ── Classify all unprocessed emails (external I/O — outside transaction) ─
-  const classified: ClassifiedEmail[] = [];
-  for (const email of unprocessed) {
-    summary.scanned++;
-    try {
-      const classification = await deps.classifier.classify(email);
-      classified.push({ email, classification });
-    } catch (err) {
-      logger.error('Email classification failed', { userId, messageId: email.messageId, error: safeError(err) });
-      classified.push({ email, classification: null });
-      summary.errors++;
-    }
-  }
+  const classified = await mapWithConcurrency<RawEmail, ClassifiedEmail>(
+    unprocessed,
+    CLASSIFY_CONCURRENCY,
+    async (email) => {
+      try {
+        const classification = await deps.classifier.classify(email);
+        return { email, classification };
+      } catch (err) {
+        logger.error('Email classification failed', {
+          userId,
+          messageId: email.messageId,
+          error: safeError(err),
+        });
+        return { email, classification: null };
+      }
+    },
+  );
+  summary.scanned = classified.length;
+  summary.errors = classified.filter(c => c.classification === null).length;
 
   // ── Hoist read-only lookups before the transaction ───────────────────────
   const [rejectionStage, userCards] = await Promise.all([
--- a/backend/src/services/gmailSync.ts
+++ b/backend/src/services/gmailSync.ts
@@ -54,7 +54,13 @@ export interface SyncSummary {
 
 const MAX_VARCHAR_255 = 255;
 const CONFIDENCE_THRESHOLD = 0.9;
-const CLASSIFY_CONCURRENCY = Number(process.env.GMAIL_CLASSIFY_CONCURRENCY ?? 5);
+// Parse and validate the concurrency cap once at module load.
+// Defends against NaN ("abc"), 0, and negative values that would silently
+// produce zero workers (returning undefined slots) or throw RangeError.
+const RAW_CLASSIFY_CONCURRENCY = Number(process.env.GMAIL_CLASSIFY_CONCURRENCY);
+const CLASSIFY_CONCURRENCY = Number.isFinite(RAW_CLASSIFY_CONCURRENCY) && RAW_CLASSIFY_CONCURRENCY >= 1
+  ? Math.floor(RAW_CLASSIFY_CONCURRENCY)
+  : 5;
 
 type ProcessedAction =
   | 'receipt_created'
@@ -82,7 +88,7 @@ async function mapWithConcurrency<T, R>(
   limit: number,
   fn: (item: T) => Promise<R>,
 ): Promise<R[]> {
-  const results: R[] = new Array(items.length);
+  const results = new Array<R>(items.length);
   let cursor = 0;
   const workers = Array.from({ length: Math.min(limit, items.length) }, async () => {
     while (true) {
--- a/backend/tests/gmailSync.test.ts
+++ b/backend/tests/gmailSync.test.ts
@@ -720,4 +720,66 @@ describe('syncUserGmailWith', () => {
       expect(summary.durationMs).toBeGreaterThanOrEqual(0);
     });
   });
+
+  describe('bounded-concurrency classification', () => {
+    it('classifies 20 emails concurrently — finishes in under 2 s with 200 ms per-email delay', async () => {
+      setupDb({ existingCards: [] });
+
+      const emails: RawEmail[] = Array.from({ length: 20 }, (_, i) => ({
+        messageId: `msg-perf-${i}`,
+        subject: `Email ${i}`,
+        sender: `sender-${i}@example.com`,
+        body: `Body ${i}`,
+        receivedAt: new Date('2024-01-15T10:00:00Z'),
+      }));
+
+      const slowClassifier: ClassifierPort = {
+        async classify(_email: RawEmail): Promise<EmailClassification> {
+          await new Promise(resolve => setTimeout(resolve, 200));
+          return { type: 'other', companyName: null, roleTitle: null, jobUrl: null, confidence: 0.9 };
+        },
+      };
+
+      const start = Date.now();
+      const summary = await syncUserGmailWith(USER_ID, {
+        gmail: fakeGmail(emails),
+        classifier: slowClassifier,
+        db: makeMockKnex(),
+      });
+      const elapsed = Date.now() - start;
+
+      expect(summary.scanned).toBe(20);
+      expect(elapsed).toBeLessThan(2000);
+    }, 10_000);
+
+    it('error isolation — classifier throwing on every 3rd email still yields a fully-populated classified array', async () => {
+      setupDb({ existingCards: [] });
+
+      const emails: RawEmail[] = Array.from({ length: 9 }, (_, i) => ({
+        messageId: `msg-err-${i}`,
+        subject: `Email ${i}`,
+        sender: `sender-${i}@example.com`,
+        body: `Body ${i}`,
+        receivedAt: new Date('2024-01-15T10:00:00Z'),
+      }));
+
+      let callCount = 0;
+      const faultingClassifier: ClassifierPort = {
+        async classify(_email: RawEmail): Promise<EmailClassification> {
+          callCount++;
+          if (callCount % 3 === 0) throw new Error('LLM rate limit');
+          return { type: 'other', companyName: null, roleTitle: null, jobUrl: null, confidence: 0.9 };
+        },
+      };
+
+      const summary = await syncUserGmailWith(USER_ID, {
+        gmail: fakeGmail(emails),
+        classifier: faultingClassifier,
+        db: makeMockKnex(),
+      });
+
+      expect(summary.scanned).toBe(9);
+      expect(summary.errors).toBe(3); // emails 3, 6, 9 had classifier errors
+    });
+  });
 });


tomerp20 (commented):
## 🔒 Security Review — PR #169

**Scope:** `backend/src/services/gmailSync.ts`, `backend/tests/gmailSync.test.ts`

**Verdict:** ✅ No security issues introduced by this PR.

### What I checked
- **Secrets / credential exposure:** No new logging paths. The classification error path goes through the existing `safeError()` helper, which already strips Gaxios `config` (Bearer access tokens) — that behaviour is preserved.
- **Injection / SQLi:** No new SQL or `.raw()` calls were introduced. The only `trx.raw('SELECT pg_advisory_xact_lock(hashtext(?))', [...])` is unchanged and properly parameterised.
- **AuthN / AuthZ:** No new endpoints, no route surface change. The function still filters by `userId` for `gmail_tokens`, `processed_emails`, `stages`, `cards` lookups.
- **OAuth / Gmail token handling:** Unchanged. Token access remains gated by `gmailService.getValidClient`.
- **AI SDK key handling:** Unchanged.
- **Env-var injection:** `process.env.GMAIL_CLASSIFY_CONCURRENCY` is coerced via `Number(...)`. It is **not** interpolated into any string, SQL, or shell command — safe from injection. (See backend review for a non-security correctness concern around `NaN` / `0` handling.)
- **Resource exhaustion / DoS:** Default concurrency cap of 5 is conservative; bursts of 5 LLM calls per user are well within typical API quotas. No new external I/O surface.
- **Dependency additions:** None (`p-limit` not added; helper is hand-rolled).

### OWASP Top-10 spot check
- A01 Broken Access Control — N/A (no auth surface change)
- A02 Cryptographic Failures — N/A
- A03 Injection — N/A
- A04 Insecure Design — concurrency bound is enforced by code, not config-only
- A05 Security Misconfiguration — env var is documented and bounded
- A09 Logging/Monitoring — `safeError` continues to mask tokens

No security fixes required.

tomerp20 (commented):
## 🔧 Backend Review — STANDARD

**Scope:** `backend/src/services/gmailSync.ts`, `backend/tests/gmailSync.test.ts`
**Classification:** STANDARD (perf change in a service that talks to Gmail/LLM; no schema, no routes, no auth surface)

### ESLint guardrail
⚠️ ESLint reports **27 errors** in `gmailSync.ts`, but **26 of those exist on `main`** — a pre-existing `no-unsafe-*` baseline from untyped Knex calls. This PR adds **1 new error**:
- `src/services/gmailSync.ts:85` — `Unsafe assignment of type 'any[]' to a variable of type 'R[]'` from `new Array(items.length)` inside `mapWithConcurrency`.
- One additional related error on `src/services/gmailSync.ts:24` (`'unknown' overrides all other types in this union type`) is pre-existing on `main`.

I'm degrading gracefully and proceeding with the semantic review; the new error is flagged below and will be fixed in Phase 2.

### What I reviewed
- The new `mapWithConcurrency` worker-pool helper (lines 80–96)
- The classification-loop replacement (lines 197–215)
- The summary derivation (`scanned` / `errors`)
- The two new tests in `bounded-concurrency classification`

### Issues by Severity

#### 🟥 HIGH — Misconfigured env var silently breaks classification
**File:** `backend/src/services/gmailSync.ts:57`, `:87`

`CLASSIFY_CONCURRENCY = Number(process.env.GMAIL_CLASSIFY_CONCURRENCY ?? 5)` does no validation. Several footguns:

1. **`GMAIL_CLASSIFY_CONCURRENCY=\"abc\"`** → `Number(\"abc\") === NaN` → `Math.min(NaN, items.length) === NaN` → `Array.from({ length: NaN }) === []` → **zero workers spawn**. `mapWithConcurrency` returns an array of `items.length` `undefined` slots. Downstream, `classified.filter(c => c.classification === null)` runs on `undefined` and throws `TypeError: Cannot read properties of undefined`. The whole sync silently fails for that user.
2. **`GMAIL_CLASSIFY_CONCURRENCY=\"0\"`** → 0 workers spawn → same outcome as above.
3. **`GMAIL_CLASSIFY_CONCURRENCY=\"-3\"`** → `Math.min(-3, n) === -3` → `Array.from({ length: -3 })` throws `RangeError: Invalid array length`.

**Fix:**
```ts
const RAW_CONCURRENCY = Number(process.env.GMAIL_CLASSIFY_CONCURRENCY);
const CLASSIFY_CONCURRENCY = Number.isFinite(RAW_CONCURRENCY) && RAW_CONCURRENCY >= 1
  ? Math.floor(RAW_CONCURRENCY)
  : 5;
```

#### 🟧 MEDIUM — Type-safety regression in `mapWithConcurrency`
**File:** `backend/src/services/gmailSync.ts:85`

`new Array(items.length)` is typed as `any[]`, which is the *one* new ESLint error this PR adds and a regression from the surrounding stricter style. The rest of the function would benefit from generic-preserving construction.

**Fix:**
```ts
const results = new Array<R>(items.length);
```

This single change removes the new ESLint error without weakening the public signature.

#### 🟨 LOW — Test asserts `summary.errors` but does not verify \"fully-populated `classified` array\"
**File:** `backend/tests/gmailSync.test.ts:755–783`

The test name says \"yields a fully-populated `classified` array\" but `classified` is internal and the assertions only check `summary.scanned === 9` and `summary.errors === 3`. The invariant the test promises — that every slot in the internal array is populated, even for failures — is satisfied transitively (because `summary.scanned = classified.length` and `summary.errors` counts nulls), but it's not directly asserted. Consider adding one of:
- An assertion that `summary.scanned + (cards processed) === 9` after wiring up a counting mock, OR
- A spy on `processed_emails` insert calls (should be exactly 9: 6 successes + 3 `classifier_error` audit rows).

Not blocking — the existing assertions are sufficient to detect the most likely regressions.

#### 🟨 LOW — New `GMAIL_CLASSIFY_CONCURRENCY` env knob is untested
**File:** `backend/tests/gmailSync.test.ts`

There's no test that pinning concurrency to `1` produces sequential behaviour, or that pinning it to a large value produces near-`items.length` parallelism. Worth one cheap test that injects the limit (would require pulling `CLASSIFY_CONCURRENCY` out of module scope or threading it through `SyncDeps`). Optional.

#### 🟨 LOW — Module-load env capture is untestable
**File:** `backend/src/services/gmailSync.ts:57`

`CLASSIFY_CONCURRENCY` is captured at module load, so per-test override via `process.env.GMAIL_CLASSIFY_CONCURRENCY = ...` after `jest.resetModules()` is the only way to exercise it. Threading the limit through `SyncDeps` (e.g., `classifyConcurrency?: number`) would make this trivially testable and let callers tune it without env-var gymnastics. Defer to a follow-up if not in scope.

### Testing Gaps
- Env-var override path (`GMAIL_CLASSIFY_CONCURRENCY`) — no test.
- Boundary case: `unprocessed.length === 0` (already covered indirectly by `gmailToken: null` tests, but a direct \"no unprocessed emails\" test would be cheap).
- The perf test (`< 2000 ms` with 20 × 200 ms) is timing-dependent — fine locally, may be flaky on slow CI. The ~5× headroom over the expected 815 ms makes flakiness unlikely but not impossible. Worth raising the threshold to ~3 s or using `jest.useFakeTimers` to remove the dependency on wall-clock.

### Summary
The core refactor is correct and well-scoped:
- Order preservation via index-write is right.
- In-batch dedup invariant in `processReceipt` (sequential consumption of `classified`) is preserved.
- `summary.scanned` and `summary.errors` derivations are numerically equivalent to the old sequential counts.
- No transactional or DB-safety regression — DB-writing loop remains sequential and pinned to the existing `pg_advisory_xact_lock`.

The only blockers are **HIGH-1 (env-var validation)** and **MEDIUM-1 (typed array construction)**. LOW items can be addressed here or deferred.

backend/src/services/gmailSync.ts
```diff
@@ -54,6 +54,7 @@ export interface SyncSummary {
 
 const MAX_VARCHAR_255 = 255;
 const CONFIDENCE_THRESHOLD = 0.9;
+const CLASSIFY_CONCURRENCY = Number(process.env.GMAIL_CLASSIFY_CONCURRENCY ?? 5);
```
tomerp20: **[HIGH] — Unvalidated env var silently breaks the entire classification step**

`Number(process.env.GMAIL_CLASSIFY_CONCURRENCY ?? 5)` is unsafe:
- `GMAIL_CLASSIFY_CONCURRENCY="abc"` → `NaN` → 0 workers → returns `undefined` slots → `classified.filter(c => c.classification === null)` throws on next line.
- `GMAIL_CLASSIFY_CONCURRENCY="0"` → 0 workers → same silent no-op.
- `GMAIL_CLASSIFY_CONCURRENCY="-3"` → `Array.from({ length: -3 })` throws `RangeError`.

**Current:**
```ts
const CLASSIFY_CONCURRENCY = Number(process.env.GMAIL_CLASSIFY_CONCURRENCY ?? 5);
```

**Fix:**
```ts
const RAW_CONCURRENCY = Number(process.env.GMAIL_CLASSIFY_CONCURRENCY);
const CLASSIFY_CONCURRENCY = Number.isFinite(RAW_CONCURRENCY) && RAW_CONCURRENCY >= 1
  ? Math.floor(RAW_CONCURRENCY)
  : 5;
```

backend/src/services/gmailSync.ts
```diff
@@ -76,6 +77,24 @@ type TrxResult = { newCard: { id: string; company_name: string; role_title: stri
 
 // ── Helpers ───────────────────────────────────────────────────────────────────
 
+async function mapWithConcurrency<T, R>(
+  items: T[],
+  limit: number,
+  fn: (item: T) => Promise<R>,
+): Promise<R[]> {
+  const results: R[] = new Array(items.length);
```
tomerp20: **[MEDIUM] — `new Array(items.length)` types results as `any[]` (new ESLint error)**

This is the single `@typescript-eslint/no-unsafe-assignment` error this PR adds on top of the pre-existing baseline on `main`. Easy fix that also documents intent.

**Current:**
```ts
const results: R[] = new Array(items.length);
```

**Fix:**
```ts
const results = new Array<R>(items.length);
```

tomerp20 (commented):
## ✅ Code Review Complete

| 🔒 Security (Orchestrator) | ✅ Done | 0 issues | 0 / 0 |
| 🔧 Backend Reviewer | ✅ Done | 4 issues (1 HIGH, 1 MEDIUM, 2 LOW) | 2 fixed, 2 deferred |
| ⚛️ React Reviewer | ⏭️ Skipped | — | No frontend files changed |

### Fixes applied in `3a2d358`
- **HIGH** — Validated `GMAIL_CLASSIFY_CONCURRENCY` against `NaN` / `0` / negatives, falling back to the default of 5. Closes the silent no-op + `TypeError` / `RangeError` cliff.
- **MEDIUM** — Typed the worker-pool results array generically (`new Array<R>(...)`), removing the one net-new ESLint `no-unsafe-assignment` error this PR added on top of the existing baseline.

### Deferred for human discretion (non-blocking LOW)
- **LOW** — Error-isolation test asserts `summary.errors` but does not directly assert `classified` is fully populated. The invariant holds transitively. Consider tightening the assertion if you want to harden against future regressions.
- **LOW** — The new `GMAIL_CLASSIFY_CONCURRENCY` env knob has no dedicated test. Could be addressed by threading `classifyConcurrency?: number` through `SyncDeps` so callers (and tests) can pin it without `jest.resetModules()` gymnastics.
- **LOW (perf-test flakiness)** — The 20-emails-under-2 s test is timing-dependent. ~5× headroom over the expected 815 ms makes flakiness unlikely but not impossible on slow CI. Consider raising the threshold to ~3 s or using fake timers if it ever flakes.

### Verification
- ESLint on `gmailSync.ts` after fix: **26 errors** — matches the pre-existing `main` baseline. The 1 net-new error introduced by this PR is resolved.
- `tsc --noEmit`: ✅ clean.
- Tests: not run by the reviewer — CI will execute the full suite.

### Next steps
- [ ] Human review of deferred LOW findings above
- [ ] Wait for CI to confirm test suite passes (27 existing + 2 new tests)

PR #172, opened 2026-05-19
fix(gmailSync): preserve Hebrew/Unicode characters in normalize() to fix companyMatch false negatives

### LINKED ISSUE #161: fix(gmailSync): preserve Hebrew/Unicode characters in normalize() to fix companyMatch false negatives
## Parent PRD

#145

## What to build

Fix the canonical-match normalizer in the Gmail sync module so company/role names containing non-ASCII letters (Hebrew first, all Unicode scripts ideally) are no longer stripped to an empty string. Today, every Hebrew-named company silently mis-matches: receipts create duplicate cards on every sync, and rejections produce false "no card found" notifications.

This issue extends the **Normalizer parity** test case called out in #145's testing strategy from a single edge case into a verified guarantee.

## 📝 Description & Root Cause

The `normalize` helper at `backend/src/services/gmailSync.ts:106-108` is the canonical key used by `companyMatch` (`:113-118`):

\`\`\`ts
function normalize(str: string): string {
  return str.toLowerCase().replace(/[^a-z0-9]/g, '');
}
\`\`\`

The character class `[^a-z0-9]` is ASCII-only. Any rune outside that range — Hebrew, Arabic, Cyrillic, accented Latin, even an em-dash — is stripped. For Hebrew-only company names such as `"בנק הפועלים"`, normalization collapses the entire string to `""`.

Downstream, `companyMatch` guards against short noise:

\`\`\`ts
if (a.length === 0 || b.length <= 3) return false;
\`\`\`

With `b === ""`, the guard always returns `false`, so every Hebrew company name produces a **false negative match**:

- **Receipt path** (`processReceipt`): a duplicate card is created on every subsequent sync because the existing card is never found.
- **Rejection path** (`processRejection`): the email is logged as `no_match` and the user receives a "no matching card was found" notification, even though the card exists.

Silent data corruption for the entire non-English user base.

## 🛠️ Proposed Solution

Expand the allowed character class and enable the `u` flag. Two acceptable implementations — **Option B recommended**.

**Option A — Minimal, Hebrew-only:**

\`\`\`ts
function normalize(str: string): string {
  return str.toLowerCase().replace(/[^a-z0-9֐-׾]/gu, '');
}
\`\`\`

**Option B — General Unicode-safe (recommended):**

\`\`\`ts
function normalize(str: string): string {
  return str.toLowerCase().replace(/[^\p{L}\p{N}]/gu, '');
}
\`\`\`

The `companyMatch` length guard (`b.length <= 3`) was tuned against stripped ASCII output and should be **revalidated** for Hebrew, where common bank/company tokens are typically 3–5 letters. Suggested follow-up: relax the guard to `b.length < 2` when the normalized string contains any non-ASCII letter, or move the threshold into a named constant with a comment explaining its noise-vs-precision trade-off.

## ✅ Acceptance criteria

- [ ] `normalize()` no longer strips Hebrew characters; `normalize("בנק הפועלים")` returns a non-empty string
- [ ] `companyMatch("בנק הפועלים", "בנק הפועלים")` returns `true`
- [ ] `companyMatch("Bank Hapoalim", "בנק הפועלים")` returns `false` (no cross-script false positives)
- [ ] Unit tests added covering: Hebrew-only, mixed Hebrew + Latin, accented Latin (`"Société Générale"`), and emoji-in-name edge cases
- [ ] Existing ASCII test cases for `companyMatch` / `roleMatch` continue to pass unchanged
- [ ] Manual regression: trigger a sync with a Hebrew-language receipt against an existing Hebrew-named card → no duplicate card created, and a `receipt_already_tracked` row is written to `processed_emails`

## Blocked by

None — can start immediately. Independent of #150-perf-classify-concurrency and #150-scale-batch-cap.

## User stories addressed

N/A — direct correctness fix on the module introduced in #145. Restores the **Normalizer parity** guarantee from that PRD's testing strategy for non-ASCII inputs.

tomerp20:
## Summary
- Replace the ASCII-only `[^a-z0-9]` regex in `normalize()` with Unicode-safe `[^\p{L}\p{N}]` and the `u` flag
- Hebrew (and any non-ASCII Unicode) company names no longer collapse to an empty string, fixing silent false negatives for the entire non-English user base
- Export `normalize` and `companyMatch` as named exports to enable direct unit testing

## Acceptance criteria
- [x] `normalize()` no longer strips Hebrew characters; `normalize("בנק הפועלים")` returns a non-empty string
- [x] `companyMatch("בנק הפועלים", "בנק הפועלים")` returns `true`
- [x] `companyMatch("Bank Hapoalim", "בנק הפועלים")` returns `false` (no cross-script false positives)
- [x] Unit tests added covering: Hebrew-only, mixed Hebrew + Latin, accented Latin (`"Société Générale"`), and emoji-in-name edge cases
- [x] Existing ASCII test cases for `companyMatch` / `roleMatch` continue to pass unchanged
- [ ] Manual regression: trigger a sync with a Hebrew-language receipt against an existing Hebrew-named card → no duplicate card created, and a `receipt_already_tracked` row is written to `processed_emails`

## Related
Closes #161

### DIFF
--- a/backend/src/services/gmailSync.ts
+++ b/backend/src/services/gmailSync.ts
@@ -105,8 +105,13 @@ function truncate(value: string | null | undefined, max = MAX_VARCHAR_255): stri
 
 // Strips non-letter/non-digit characters (Unicode-aware) and lowercases.
 // The `u` flag enables \p{L} (any letter) and \p{N} (any digit) Unicode property escapes.
+// NFC normalization is applied first so that the same visually identical name in
+// composed (NFC) and decomposed (NFD) form produces the same output — combining
+// marks (\p{M}) are not letters and would otherwise be stripped, e.g. NFD
+// `"Société"` would lose its acute accents while NFC `"Société"`
+// would keep them, causing companyMatch to falsely diverge.
 export function normalize(str: string): string {
-  return str.toLowerCase().replace(/[^\p{L}\p{N}]/gu, '');
+  return str.normalize('NFC').toLowerCase().replace(/[^\p{L}\p{N}]/gu, '');
 }
 
 // Canonical company match — used for both receipt and rejection paths.
--- a/backend/src/services/gmailSync.ts
+++ b/backend/src/services/gmailSync.ts
@@ -103,16 +103,16 @@ function truncate(value: string | null | undefined, max = MAX_VARCHAR_255): stri
   return value.length > max ? value.slice(0, max) : value;
 }
 
-// Strips all non-ASCII-alphanumeric chars and lowercases. Returns empty string
-// for names that consist entirely of non-ASCII characters (e.g. Hebrew-only).
-function normalize(str: string): string {
-  return str.toLowerCase().replace(/[^a-z0-9]/g, '');
+// Strips non-letter/non-digit characters (Unicode-aware) and lowercases.
+// The `u` flag enables \p{L} (any letter) and \p{N} (any digit) Unicode property escapes.
+export function normalize(str: string): string {
+  return str.toLowerCase().replace(/[^\p{L}\p{N}]/gu, '');
 }
 
 // Canonical company match — used for both receipt and rejection paths.
-// Rejects if the card company has no ASCII alphanumeric content, or the
-// extracted name is too short (≤3 chars) to be meaningful.
-function companyMatch(cardCompany: string, extractedCompany: string): boolean {
+// Rejects if either normalized name is empty, or the extracted name is too
+// short (≤3 chars) to avoid noise matches.
+export function companyMatch(cardCompany: string, extractedCompany: string): boolean {
   const a = normalize(cardCompany);
   const b = normalize(extractedCompany);
   if (a.length === 0 || b.length <= 3) return false;
--- a/backend/tests/gmailSync.test.ts
+++ b/backend/tests/gmailSync.test.ts
@@ -302,6 +302,12 @@ describe('companyMatch', () => {
   it('matches accented-Latin company names', () => {
     expect(companyMatch('Société Générale', 'Société Générale')).toBe(true);
   });
+
+  it('matches the same name in NFC vs NFD form', () => {
+    const nfc = 'Société Générale';            // 'é' = U+00E9 (precomposed)
+    const nfd = 'Société Générale';        // 'e' + U+0301 (combining acute)
+    expect(companyMatch(nfc, nfd)).toBe(true);
+  });
 });
 
 // ── Tests ─────────────────────────────────────────────────────────────────────
--- a/backend/tests/gmailSync.test.ts
+++ b/backend/tests/gmailSync.test.ts
@@ -41,7 +41,7 @@ jest.mock('../src/services/cardService', () => ({
   resolveCompanyIconUrl: jest.fn().mockResolvedValue(null),
 }));
 
-import { syncUserGmailWith, GmailPort, ClassifierPort, SyncDeps, RawEmail, GmailClient } from '../src/services/gmailSync';
+import { syncUserGmailWith, normalize, companyMatch, GmailPort, ClassifierPort, SyncDeps, RawEmail, GmailClient } from '../src/services/gmailSync';
 import { cardService } from '../src/services/cardService';
 import { EmailClassification } from '../src/services/emailClassifier';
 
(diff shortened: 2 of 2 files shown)


tomerp20 (commented):
## 🔧 Backend Review — HIGH-RISK

Reviewed: `backend/src/services/gmailSync.ts` and `backend/tests/gmailSync.test.ts`.

Classification rationale: this is a small, surgical change, but it sits on the Gmail sync hot path (OAuth-adjacent, drives card mutations across the entire user base). Subtle false positives or negatives here either create duplicate cards or silently swallow real receipts/rejections — same blast radius as the bug being fixed. Hence HIGH-RISK, not STANDARD.

### Guardrails
- ESLint baseline on `src/services/gmailSync.ts`: 26 errors, all pre-existing on lines untouched by this PR (`no-unsafe-assignment` / `no-unsafe-member-access` patterns at lines 24, 154, 160, 192, 247, 291, 356, 402, 405, 417, 429, 436, 540). None introduced by this diff. Not blocking, but worth a follow-up cleanup ticket — the file's lint baseline is dirty enough that future real regressions will be hard to spot.
- Test file is in the ESLint ignore pattern (intentional, existing convention).

### Issues by Severity

**MEDIUM**
1. `normalize()` is not NFC-aware. The regex `\p{L}` matches base letters but NOT combining marks (`\p{M}`). A string in NFD form like `"Société"` (`e` + U+0301) normalizes to `societe`, while the same string in NFC form (`é` = U+00E9) normalizes to `société`. Two visually identical inputs from different sources (e.g. macOS-originated text vs Gmail-rendered text) will fail `companyMatch`. Since this PR's entire purpose is to make Unicode names match correctly, this edge case is squarely in scope. Fix: call `.normalize('NFC')` on the input before the regex. See inline comment for the proposed change and a regression test.

**LOW**
2. `companyMatch` substring matching across mixed scripts: the `a.includes(b) || b.includes(a)` rule now operates over a much larger character set. The `b.length <= 3` guard is the only protection against tiny extracted strings matching unrelated card names. Behaviour is identical to before for ASCII, but worth noting: a 4-character Hebrew extracted name `"בנק"` (3 chars — guard catches it) is rejected; a 4-letter `"בנקא"` could substring-match into longer Hebrew card names. Current guard is fine; flagging for future awareness if multilingual extraction quality is uneven.

### Testing Gaps
- No test covering NFC/NFD equivalence — recommended once the `.normalize('NFC')` fix is applied:
  ```ts
  it('matches the same name in NFC vs NFD form', () => {
    const nfc = 'Société';                  // 'é' as U+00E9
    const nfd = 'Société';      // 'e' + combining acute
    expect(companyMatch(nfc, nfd)).toBe(true);
  });
  ```
- No test for `roleMatch` with non-ASCII roles. `roleMatch` is internal so optional, but it has the same NFC/NFD trap and is used on the receipt duplicate-detection path. Worth at least one Hebrew/accented role title in the existing `syncUserGmailWith` suite.

### What's good
- The regex change is minimal and correct for the common case. The `gu` flags and `\p{L}\p{N}` choice are exactly right.
- Exporting `normalize` / `companyMatch` for unit testing is the right move; the surface area is small and the helpers are pure.
- The flipped `normalizer parity` test now correctly asserts the new behavior — good catch including that in the same PR rather than leaving a stale assertion.
- Helper visibility (`roleMatch` stays unexported) is principled: only the helpers under test are exported.

backend/src/services/gmailSync.ts
```diff
@@ -103,16 +103,16 @@ function truncate(value: string | null | undefined, max = MAX_VARCHAR_255): stri
...
 }
 
-// Strips all non-ASCII-alphanumeric chars and lowercases. Returns empty string
-// for names that consist entirely of non-ASCII characters (e.g. Hebrew-only).
-function normalize(str: string): string {
-  return str.toLowerCase().replace(/[^a-z0-9]/g, '');
+// Strips non-letter/non-digit characters (Unicode-aware) and lowercases.
+// The `u` flag enables \p{L} (any letter) and \p{N} (any digit) Unicode property escapes.
+export function normalize(str: string): string {
+  return str.toLowerCase().replace(/[^\p{L}\p{N}]/gu, '');
```
tomerp20: **[MEDIUM] — `normalize()` is not NFC-aware; NFC vs NFD inputs produce different outputs**

`\p{L}` matches letters but NOT combining marks (`\p{M}`). When the same accented name arrives in different Unicode normalization forms — e.g. `Société` as NFC (`é` = U+00E9) vs NFD (`e` + U+0301) — `normalize()` produces `société` for one and `societe` for the other. `companyMatch` then returns `false` for two visually identical inputs.

Realistic source of asymmetry: macOS filesystem text is NFD, Gmail/web-form input is typically NFC. Card company names entered by users vs extracted from email bodies are likely to drift apart in form even when they're the 'same' name.

Since the whole point of this PR is correct Unicode matching, this is in scope.

**Current:**
```ts
export function normalize(str: string): string {
  return str.toLowerCase().replace(/[^\p{L}\p{N}]/gu, '');
}
```

**Fix:**
```ts
export function normalize(str: string): string {
  return str.normalize('NFC').toLowerCase().replace(/[^\p{L}\p{N}]/gu, '');
}
```

Add a regression test:
```ts
it('matches the same name in NFC vs NFD form', () => {
  const nfc = 'Société';                // 'é' = U+00E9
  const nfd = 'Socie\u0301te\u0301';   // 'e' + U+0301 (combining acute)
  expect(companyMatch(nfc, nfd)).toBe(true);
});
```

PR #176, opened 2026-05-19
fix: consistent error shape and remove redundant DB lookup in events.ts SSE route

### LINKED ISSUE #175: fix: consistent error shape and remove redundant DB lookup in events.ts SSE route
## Problem

The SSE route (\`GET /api/events\`) has two inconsistencies with the rest of the backend:

**1. Wrong error response shape.**
Every other route returns errors as:
\`\`\`json
{ "error": { "message": "...", "code": "..." } }
\`\`\`
The SSE route returns a plain string:
\`\`\`json
{ "error": "Missing token" }
\`\`\`
This breaks any client code that reads \`error.message\` or \`error.code\` — it gets \`undefined\` on this route.

**2. Redundant DB lookup.**
After verifying the JWT signature, the route queries the \`users\` table to confirm the user exists before extracting \`userId\`. The JWT signature is sufficient proof of identity — \`userId\` is already in the decoded payload and doesn't need a round-trip to confirm.

## Changes

Both changes are in the single SSE route handler only.

**Fix 1 — error shape (3 lines):**
Replace the three direct \`res.status().json({ error: 'string' })\` calls with \`next(new AppError(...))\` so all errors flow through the global error handler and produce the standard \`{ error: { message, code } }\` shape.

Error codes to use:
- Missing token → \`ERR_MISSING_TOKEN\` / 401
- Invalid/expired token → \`ERR_INVALID_TOKEN\` / 401

**Fix 2 — remove DB lookup:**
Delete the \`db('users')\` query block entirely. Read \`userId\` directly from \`decoded.userId\` (already present in the JWT payload). This is independent of issue #174 — \`userId\` is in the current JWT payload today.

## Acceptance Criteria

- [ ] All three error responses from this route return \`{ error: { message, code } }\`
- [ ] No database query is made during SSE auth
- [ ] SSE connection still works correctly for a valid token
- [ ] A missing or invalid token still returns 401

## Notes

- No schema changes. No other files touched.
- This issue is independent of #174 (JWT claims refactor) — can be merged in any order.

tomerp20:
## Summary
- Replace plain-string error responses in the SSE route with `next(new AppError(...))` so all errors return the standard `{ error: { message, code } }` shape
- Remove unnecessary `db('users')` round-trip after JWT verification — `userId` is already in the decoded payload

## Acceptance criteria
- [x] All three error responses from this route return `{ error: { message, code } }`
- [x] No database query is made during SSE auth
- [x] SSE connection still works correctly for a valid token
- [x] A missing or invalid token still returns 401

## Related
Closes #175

### DIFF
--- a/backend/src/routes/events.ts
+++ b/backend/src/routes/events.ts
@@ -19,7 +19,7 @@ const router = Router();
  * access logs and browser history. Ensure any access-log middleware (e.g.
  * morgan) is configured to redact the `token` query param on this route.
  */
-router.get('/', async (req: Request, res: Response, next: NextFunction): Promise<void> => {
+router.get('/', (req: Request, res: Response, next: NextFunction): void => {
   const token = req.query.token as string;
 
   if (!token) {
@@ -31,6 +31,11 @@ router.get('/', async (req: Request, res: Response, next: NextFunction): Promise
   try {
     decoded = jwt.verify(token, env.JWT_SECRET) as JwtPayload;
   } catch {
+    next(new AppError('Invalid token', 401, 'ERR_INVALID_TOKEN'));
+    return;
+  }
+
+  if (typeof decoded.userId !== 'string' || decoded.userId.length === 0) {
     next(new AppError('Invalid token', 401, 'ERR_INVALID_TOKEN'));
     return;
   }
--- a/backend/src/routes/events.ts
+++ b/backend/src/routes/events.ts
@@ -1,10 +1,10 @@
 import { Router, Request, Response, NextFunction } from 'express';
 import jwt from 'jsonwebtoken';
-import db from '../config/database';
 import logger from '../config/logger';
 import { pgSubscriber } from '../services/pgSubscriber';
 import { JwtPayload } from '../middleware/auth';
 import { env } from '../config/env';
+import { AppError } from '../middleware/errorHandler';
 
 const router = Router();
 
@@ -23,7 +23,7 @@ router.get('/', async (req: Request, res: Response, next: NextFunction): Promise
   const token = req.query.token as string;
 
   if (!token) {
-    res.status(401).json({ error: 'Missing token' });
+    next(new AppError('Missing token', 401, 'ERR_MISSING_TOKEN'));
     return;
   }
 
@@ -31,29 +31,11 @@ router.get('/', async (req: Request, res: Response, next: NextFunction): Promise
   try {
     decoded = jwt.verify(token, env.JWT_SECRET) as JwtPayload;
   } catch {
-    res.status(401).json({ error: 'Invalid token' });
+    next(new AppError('Invalid token', 401, 'ERR_INVALID_TOKEN'));
     return;
   }
 
-  // Verify the user still exists in the database — wrapped in try/catch so DB
-  // errors are forwarded to the Express error handler before SSE headers are set
-  let userId: string;
-  try {
-    const user = await db('users')
-      .select('id')
-      .where({ id: decoded.userId })
-      .first();
-
-    if (!user) {
-      res.status(401).json({ error: 'Invalid token' });
-      return;
-    }
-
-    userId = user.id as string;
-  } catch (err) {
-    next(err);
-    return;
-  }
+  const userId: string = decoded.userId;
 
   // Disable compression for this response — the compression middleware buffers
   // writes into a gzip stream, preventing heartbeats from reaching the proxy.


tomerp20 (commented):
## Security Review (Orchestrator)

**Scope:** `backend/src/routes/events.ts` — SSE auth refactor (1 file, +5/-23).

### Files Reviewed

#### `backend/src/routes/events.ts`
**Summary:** Replaces three plain-string `res.status(...).json({ error: '...' })` responses with `next(new AppError(...))` so the route returns the standard `{ error: { message, code } }` envelope, and removes the post-JWT `db('users')` existence check — `userId` is now taken directly from the verified JWT payload.

#### Security Findings

**[LOW] — Trusting `decoded.userId` without shape validation**
- **Description:** The handler now does `const userId: string = decoded.userId;` and immediately passes it to `pgSubscriber.registerClient(userId, res)`. `jwt.verify` only validates the signature and expiry — it does not enforce the payload shape. A JWT signed with the right secret but a malformed payload (e.g. `userId` missing, or set to a non-string) would let an empty/`undefined` userId reach the pubsub registry. Practical exploit risk is low because tokens are only minted by `authService` (which always sets `userId`), but the previous code defended against this by re-checking the DB. Recommend a single defensive guard:
  ```ts
  if (typeof decoded.userId !== 'string' || decoded.userId.length === 0) {
    next(new AppError('Invalid token', 401, 'ERR_INVALID_TOKEN'));
    return;
  }
  const userId: string = decoded.userId;
  ```

**[LOW / informational] — Stateless JWT no longer revocable via user deletion**
- **Description:** Removing the `db('users')` lookup means a deleted/disabled user with an un-expired JWT can still open an SSE stream. This is the expected trade-off for stateless JWT auth and matches the behavior of the standard `authenticate()` middleware (`backend/src/middleware/auth.ts`), so it is consistent with the rest of the codebase. Flagging here only as a note in case a future revocation requirement comes up — no change requested in this PR.

**[NONE] — Secrets, CORS, injection, XSS, path traversal, rate limiting, dependency additions**
- No new secrets in the diff. `JWT_SECRET` is read from `process.env` as before. No new dependencies, no `.raw()` SQL, no user-controlled paths, no HTML rendering. The token is still passed as a query string (`?token=...`) which appears in access logs and browser history — but this is pre-existing and already called out by the comment block at the top of the file (lines 17–19). No new exposure introduced.

No CRITICAL/HIGH/MEDIUM findings.

tomerp20 (commented):
## Backend Review — STANDARD

> Note: the harness running this orchestration session does not expose a sub-agent spawn tool, so this backend review is performed inline by the orchestrator using the `backend-code-reviewer` rubric. Findings and rigor are unchanged; only the delivery agent differs.

### ESLint guardrail

Ran `npx eslint` against the PR head and against `main` for comparison:

| Branch | Errors |
|--------|--------|
| `main` (before PR) | 3 — `no-misused-promises` (line 21) + 2 `no-unsafe-*` from the Knex `db('users')` call |
| PR head | 2 — `no-misused-promises` (line 21) + **new** `require-await` (line 21) |

Net effect on lint baseline: −2 errors (the unsafe-any pair is gone) and +1 new error (`require-await`). The pre-existing `no-misused-promises` is unchanged and matches the team's established baseline noise (orchestrator memory flags this as accepted in service files). I am not blocking on the pre-existing error, but the new `require-await` should be addressed since this PR causes it.

### Issues by Severity

#### MEDIUM — `require-await` regression on the route handler (events.ts:21)

The handler was `async` because it `await`-ed the `db('users')` query. With the DB lookup removed, the function no longer contains a single `await`, so `@typescript-eslint/require-await` flags it.

**Current:**
```ts
router.get('/', async (req: Request, res: Response, next: NextFunction): Promise<void> => {
  // no awaits anywhere in the body
});
```

**Fix:** drop the `async` keyword (and the `Promise<void>` return type, replacing it with `void`):
```ts
router.get('/', (req: Request, res: Response, next: NextFunction): void => {
  // ...
});
```

This also resolves cleanly: there is no operation in the handler that returns a Promise. Express handlers can be sync — they just need to call `next()` or write to `res`.

#### LOW — Defensive shape check on `decoded.userId` (events.ts:43)

See orchestrator security review above. `jwt.verify` does not enforce payload shape; a one-line guard prevents an empty/undefined `userId` from reaching `pgSubscriber.registerClient`. Not blocking — tokens minted by `authService` always include `userId` — but recommended as a belt-and-braces defense given the DB check was the previous safety net.

#### LOW — `ERR_500` vs project-wide `ERR_INTERNAL` (events.ts:31)

```ts
next(new AppError('Internal server error', 500));
```

`AppError`'s default code is `ERR_${statusCode}` → `ERR_500`. Everywhere else in the codebase that emits an internal-server-error uses `'ERR_INTERNAL'`:

- `backend/src/middleware/errorHandler.ts:48`
- `backend/src/middleware/auth.ts:62`, `:98`
- `backend/src/services/gmailService.ts:26`, `:36`
- `backend/src/services/authService.ts:42` (uses `'ERR_CONFIG'` — also internal-class)

Minor consistency issue: clients reading the error code to log/handle internal errors will see `ERR_500` here vs `ERR_INTERNAL` elsewhere. Recommend:
```ts
next(new AppError('Internal server error', 500, 'ERR_INTERNAL'));
```

### Categories checked, nothing found

- **Route integrity** — no auth middleware applies here (auth is inline via query-string JWT, intentional for SSE). No `req.body` so no `validate(...)` needed. Status codes (401/500) are correct.
- **Knex query safety** — the only Knex call (`db('users')`) was *removed*. No new queries. No `.raw()`. No transaction risk.
- **Error handling** — `next(new AppError(...))` is the correct project pattern. No swallowed errors. No stack-trace leak — `errorHandler.ts` already strips stacks before responding.
- **TypeScript** — no new `as any`, no new `!` non-null assertions, no untyped function signatures. The explicit `const userId: string = decoded.userId;` annotation is redundant since `JwtPayload.userId` is already typed `string`, but harmless.
- **Migrations** — N/A, none in this PR.
- **Transactions** — N/A.
- **N+1 / unbounded queries / performance** — DB call removed, so this PR strictly improves the request path. One fewer round-trip per SSE connect.
- **Sensitive data in logs** — no token values, no secrets in any new log statement.
- **AI SDK / OAuth tokens** — not touched.

### Testing Gaps

The PR description claims all four acceptance criteria pass, but the diff does not include test changes. Edge cases I'd want covered by an existing or new test:

- [ ] Missing-token request returns `{ error: { message: 'Missing token', code: 'ERR_MISSING_TOKEN' } }` with 401 (new error shape).
- [ ] Invalid/forged-token request returns `{ error: { message: 'Invalid token', code: 'ERR_INVALID_TOKEN' } }` with 401.
- [ ] Expired token: currently collapses into `ERR_INVALID_TOKEN`. The standard `authenticate()` middleware distinguishes `'Token has expired'` from `'Invalid token'` via `jwt.TokenExpiredError` (see `auth.ts:72-75`) — worth either matching that distinction here or accepting the consolidation deliberately.
- [ ] Valid token: SSE handshake still sends the initial `connected` event and `pgSubscriber.registerClient` is called with the JWT's `userId`.

If there is no existing integration test for `GET /api/events` auth paths, this would be a good follow-up issue rather than a blocker for this PR (the change is small enough to verify manually per the PR's checklist).

### Summary

Clean, focused refactor that aligns this endpoint with the project's standard error envelope and removes an unnecessary DB round-trip. One **MEDIUM** regression (`require-await`) and two **LOW** consistency/defense findings. Nothing CRITICAL/HIGH. The MEDIUM is trivially fixable and will be addressed in Phase 2.

backend/src/routes/events.ts
```diff
@@ -22,44 +22,26 @@ router.get('/', async (req: Request, res: Response, next: NextFunction): Promise
   const token = req.query.token as string;
```
tomerp20, line 23: **[MEDIUM] — `require-await` regression**

The handler was `async` only because it awaited the `db('users')` query that this PR removes. With no `await` left in the body, `@typescript-eslint/require-await` now reports an error here.

**Fix:** drop `async` and `Promise<void>`:
```ts
router.get('/', (req: Request, res: Response, next: NextFunction): void => {
```

backend/src/routes/events.ts
```diff
@@ -22,44 +22,26 @@ router.get('/', async (req: Request, res: Response, next: NextFunction): Promise
...
-      res.status(401).json({ error: 'Invalid token' });
-      return;
-    }
-
-    userId = user.id as string;
-  } catch (err) {
-    next(err);
-    return;
-  }
+  const userId: string = decoded.userId;
```
tomerp20, line 44: **[LOW] — Defensive shape check on `decoded.userId`**

`jwt.verify` validates signature/expiry but not payload shape. With the DB existence check gone, a token with a missing/non-string `userId` would silently reach `pgSubscriber.registerClient`. Recommended single-line guard:

```ts
if (typeof decoded.userId !== 'string' || decoded.userId.length === 0) {
  next(new AppError('Invalid token', 401, 'ERR_INVALID_TOKEN'));
  return;
}
const userId: string = decoded.userId;
```

Not blocking — `authService` always sets `userId` — but the previous code defended against this implicitly via the DB lookup.

backend/src/routes/events.ts
```diff
@@ -22,44 +22,26 @@ router.get('/', async (req: Request, res: Response, next: NextFunction): Promise
...
-    res.status(401).json({ error: 'Missing token' });
+    next(new AppError('Missing token', 401, 'ERR_MISSING_TOKEN'));
     return;
   }
 
   const secret = process.env.JWT_SECRET;
   if (!secret) {
     logger.error('JWT_SECRET is not configured');
-    res.status(500).json({ error: 'Internal server error' });
+    next(new AppError('Internal server error', 500));
```
tomerp20: **[LOW] — Use `ERR_INTERNAL` to match project convention**

`AppError` defaults the code to `ERR_${statusCode}` → `ERR_500`. Every other internal-server-error in the codebase uses `'ERR_INTERNAL'` explicitly (`middleware/errorHandler.ts:48`, `middleware/auth.ts:62,98`, `services/gmailService.ts:26,36`).

**Fix:**
```ts
next(new AppError('Internal server error', 500, 'ERR_INTERNAL'));
```

tomerp20 (commented):
## Verification Pass

Fix commit `382e3d4` pushed. Re-fetched the PR diff and confirmed all three findings are addressed:

| Finding | Status |
|---------|--------|
| `require-await` regression on the route handler | ✅ Fixed — handler is now sync (`(req, res, next): void`), no `async` keyword |
| Defensive shape check on `decoded.userId` | ✅ Fixed — `typeof decoded.userId !== 'string' \|\| decoded.userId.length === 0` guard returns 401 `ERR_INVALID_TOKEN` |
| Use `'ERR_INTERNAL'` for the missing-JWT_SECRET branch | ✅ Fixed — `new AppError('Internal server error', 500, 'ERR_INTERNAL')` |

Local checks run from the main checkout against the fixed file:
- `npx eslint src/routes/events.ts` → 0 errors, 0 warnings (was 2 errors on the PR head)
- `npx tsc --noEmit` → 0 errors

No new issues introduced by the fixes. CI will own the actual test execution per the backend reviewer's policy.

## ✅ Code Review Complete

| 🔒 Security (Orchestrator) | ✅ Done | 1 LOW + 1 informational | 1 fixed (shape check), 1 noted (stateless JWT — intentional, no change) |
| 🔧 Backend Reviewer (inline by orchestrator) | ✅ Done | 1 MEDIUM + 2 LOW | 3 fixed |
| ⚛️ React Reviewer | ➖ Skipped | n/a | Only backend files changed |

### Notes
- The harness running this session did not expose a sub-agent spawn tool, so the backend review was performed inline by the orchestrator using the `backend-code-reviewer` rubric — same rigor, different delivery agent. This is disclosed in the backend review comment for the audit trail.
- The pre-existing `no-misused-promises` lint error on `main` was eliminated as a side effect of dropping `async` from the handler (Express now sees a sync handler returning `void`).

### Skipped fixes (require manual resolution)
_None — all findings were addressed._

### Next steps
- [ ] CI to run the full test suite on `382e3d4`
- [ ] Human review and merge when satisfied

PR #177, opened 2026-05-19
refactor: embed user claims in JWT and clean up gmail /sync auth

### LINKED ISSUE #174: refactor: embed user claims in JWT and clean up gmail /sync auth
## Problem Statement

Every authenticated API request performs an unnecessary database round-trip. After verifying the JWT signature, the auth middleware immediately queries the \`users\` table to fetch \`id\`, \`email\`, and \`name\` — data that was already available at sign-in time and could have been embedded in the token. This adds latency to every single request with no benefit.

Separately, the \`POST /api/gmail/sync\` route handles two callers (an automated cron job and the authenticated user) by calling \`authenticate\` manually inside a \`Promise\` wrapper. This is fragile: \`authenticate\` was designed as an Express middleware, not a function to be manually invoked mid-handler. If \`authenticate\` ever calls \`next(err)\` with a non-Error value, the error disappears silently.

## Solution

Embed the user's \`id\`, \`email\`, and \`name\` in the JWT payload at sign-in time, so \`authenticate\` only needs to verify the token signature — no DB lookup. For the \`/sync\` route, introduce a small \`syncAuth\` middleware that handles both auth paths and passes control to \`authenticate\` as a proper Express middleware step.

## User Stories

1. As a developer, I want auth middleware that does not query the database on every request, so that authenticated endpoints have lower latency.
2. As a developer, I want the JWT payload to carry all the user fields needed by \`req.user\`, so that \`authenticate\` has a single responsibility (token verification).
3. As a developer, I want \`authenticate\` to be used as a standard Express middleware everywhere it appears, so that error propagation is consistent and predictable.
4. As a developer, I want the \`/sync\` route to handle both the cron path and the user path using proper Express middleware chaining, so that auth errors on the user path propagate through \`next(err)\` like every other route.
5. As a developer, I want \`generateTokens\` to accept \`name\` alongside \`userId\` and \`email\`, so that the full \`req.user\` shape can be reconstructed from the token alone.

## Implementation Decisions

- **JWT payload shape** — expand the payload signed by \`generateTokens\` from \`{ userId, email }\` to \`{ userId, email, name }\`. Update \`JwtPayload\` interface accordingly. All three callers of \`generateTokens\` (\`signup\`, \`login\`, \`refreshTokens\`) already have \`user.name\` available and will pass it in.

- **\`authenticate\` middleware** — after verifying the token signature, set \`req.user\` directly from the decoded payload (\`{ id: decoded.userId, email: decoded.email, name: decoded.name }\`). Delete the \`db('users')\` lookup entirely. The 401 \`ERR_USER_NOT_FOUND\` branch is removed (a valid, unexpired token is sufficient proof of identity for a single-user system).

- **Name staleness** — \`name\` never changes in this application. Embedding it in a 15-minute access token is safe. No cache invalidation mechanism is needed.

- **\`syncAuth\` middleware** — a new middleware function in \`routes/gmail.ts\` that checks whether the request carries a valid \`CRON_API_KEY\` bearer token. If yes, it marks the request (e.g. a flag on the request object) and calls \`next()\`. If no, it delegates to \`authenticate\` as a standard Express middleware call. The \`/sync\` handler then reads the cron flag to decide which path to run.

- **No URL change** — \`POST /api/gmail/sync\` keeps the same URL. The Render cron job does not need to be reconfigured.

- **No schema changes** — this refactor is entirely in application code.

## Testing Decisions

- Good tests for this change verify external behaviour, not internals: they call the HTTP endpoints with a valid/invalid token and assert on the HTTP response — they do not assert on whether a DB query was or was not made.
- The auth middleware change should be covered by an integration test: sign up, call an authenticated endpoint with the returned token, assert 200 and correct \`req.user\` fields in the response.
- The \`/sync\` cron path should be tested with a valid \`CRON_API_KEY\` header, asserting 200 without a user JWT.
- The \`/sync\` user path should be tested with a valid user JWT, asserting 200 and a sync summary.
- Prior art: \`backend/tests/gmailSync.test.ts\` shows the integration test pattern used in this project (real DB, real service calls, assertions on returned data).

## Out of Scope

- Token caching or in-memory user session store.
- Splitting \`/sync\` into separate cron and user URLs.
- Any change to refresh token logic or token expiry durations.
- Adding profile-edit functionality.
- Parallelising the cron path's per-user sync loop (tracked separately in the background tasks pillar).

## Further Notes

This refactor was identified during a backend architectural audit (2026-05-19). The \`authenticate\` DB lookup is the only per-request query that cannot be justified by business logic — every other query in an authenticated request is load-bearing. Removing it is the single highest-value change in the auth pillar.

tomerp20:
## Summary
- Embed `name` alongside `userId` and `email` in JWT access tokens, so `authenticate` can populate `req.user` without a DB round-trip on every request
- Remove the `db('users')` lookup from `authenticate` middleware — token verification is now the only check
- Introduce `syncAuth` middleware for `POST /api/gmail/sync` that handles the dual-caller pattern (cron vs. user JWT) using proper Express middleware chaining instead of a fragile `Promise`-wrapped manual call

## Acceptance criteria
- [x] `generateTokens` accepts `name` alongside `userId` and `email`; JWT payload shape is `{ userId, email, name }`
- [x] `JwtPayload` interface updated to include `name: string`
- [x] `authenticate` populates `req.user` from decoded token payload — no DB query
- [x] `ERR_USER_NOT_FOUND` branch removed; a valid unexpired token is sufficient proof of identity
- [x] `syncAuth` middleware in `routes/gmail.ts` checks CRON_API_KEY first, sets `req.isCronRequest = true` and calls `next()` on match; delegates to `authenticate` as a standard middleware call otherwise
- [x] `POST /api/gmail/sync` handler reads `req.isCronRequest` flag to select cron vs user path
- [x] Auth tests updated: `generateValidToken` includes `name`; `GET /api/auth/me` test needs no DB mock
- [x] New `gmail.test.ts` covers cron path (valid key → 200), user path (valid JWT → 200), no token → 401, invalid JWT → 401, expired JWT → 401
- [x] All 104 tests pass

## Related
Closes #174

### DIFF
--- a/backend/src/middleware/auth.ts
+++ b/backend/src/middleware/auth.ts
@@ -1,9 +1,7 @@
 import { Request, Response, NextFunction } from 'express';
 import jwt from 'jsonwebtoken';
-import db from '../config/database';
 import logger from '../config/logger';
 
-// Extend Express Request to include authenticated user
 declare global {
   namespace Express {
     interface Request {
@@ -12,6 +10,7 @@ declare global {
         email: string;
         name: string;
       };
+      isCronRequest?: boolean;
     }
   }
 }
@@ -19,18 +18,11 @@ declare global {
 export interface JwtPayload {
   userId: string;
   email: string;
+  name: string;
   iat: number;
   exp: number;
 }
 
-/**
- * Express middleware that verifies the JWT access token from the
- * Authorization: Bearer <token> header, looks up the user in the database,
- * and attaches { id, email, name } to req.user.
- *
- * Returns 401 if the token is missing, malformed, expired, or belongs to a
- * user that no longer exists.
- */
 export async function authenticate(
   req: Request,
   res: Response,
@@ -91,26 +83,10 @@ export async function authenticate(
       return;
     }
 
-    // Look up user in database to ensure they still exist
-    const user = await db('users')
-      .select('id', 'email', 'name')
-      .where({ id: decoded.userId })
-      .first();
-
-    if (!user) {
-      res.status(401).json({
-        error: {
-          message: 'User not found',
-          code: 'ERR_USER_NOT_FOUND',
-        },
-      });
-      return;
-    }
-
     req.user = {
-      id: user.id,
-      email: user.email,
-      name: user.name,
+      id: decoded.userId,
+      email: decoded.email,
+      name: decoded.name,
     };
 
     next();
--- a/backend/src/routes/gmail.ts
+++ b/backend/src/routes/gmail.ts
@@ -2,6 +2,7 @@ import { Router, Request, Response, NextFunction } from 'express';
 import { timingSafeEqual, createHash } from 'crypto';
 import db from '../config/database';
 import { authenticate } from '../middleware/auth';
+import logger from '../config/logger';
 import { gmailService } from '../services/gmailService';
 import { syncUserGmail } from '../services/gmailSync';
 
@@ -57,7 +58,11 @@ function syncAuth(req: Request, res: Response, next: NextFunction): void {
     return;
   }
 
-  authenticate(req, res, next);
+  if (cronKey) {
+    logger.warn('syncAuth: Authorization header did not match CRON_API_KEY — falling through to JWT auth');
+  }
+
+  authenticate(req, res, next).catch(next);
 }
 
 router.post('/sync', syncAuth, async (req: Request, res: Response, next: NextFunction) => {
--- a/backend/src/routes/gmail.ts
+++ b/backend/src/routes/gmail.ts
@@ -47,13 +47,22 @@ router.delete('/disconnect', authenticate, async (req: Request, res: Response, n
   } catch (err) { next(err); }
 });
 
-router.post('/sync', async (req: Request, res: Response, next: NextFunction) => {
-  try {
-    const cronKey = process.env.CRON_API_KEY;
-    const authHeader = req.headers.authorization ?? '';
-    const isCron = !!cronKey && safeCompare(authHeader, `Bearer ${cronKey}`);
+function syncAuth(req: Request, res: Response, next: NextFunction): void {
+  const cronKey = process.env.CRON_API_KEY;
+  const authHeader = req.headers.authorization ?? '';
 
-    if (isCron) {
+  if (cronKey && safeCompare(authHeader, `Bearer ${cronKey}`)) {
+    req.isCronRequest = true;
+    next();
+    return;
+  }
+
+  authenticate(req, res, next);
+}
+
+router.post('/sync', syncAuth, async (req: Request, res: Response, next: NextFunction) => {
+  try {
+    if (req.isCronRequest) {
       const tokens = await db('gmail_tokens').where({ is_valid: true }).select('user_id');
       const results: Record<string, unknown> = {};
       for (const { user_id } of tokens) {
@@ -62,11 +71,7 @@ router.post('/sync', async (req: Request, res: Response, next: NextFunction) => 
       return res.json({ results });
     }
 
-    await new Promise<void>((resolve, reject) => {
-      authenticate(req as Request, res as Response, (err?: unknown) => (err ? reject(err) : resolve()));
-    });
-
-    const summary = await syncUserGmail((req as Request).user!.id);
+    const summary = await syncUserGmail(req.user!.id);
     res.json({ data: summary });
   } catch (err) { next(err); }
 });
--- a/backend/tests/gmail.test.ts
+++ b/backend/tests/gmail.test.ts
@@ -1,0 +1,150 @@
+// ── Mock the database module before any imports that use it ───────────────────
+
+const mockDb = jest.fn();
+
+function createQueryChain(resolvedValue: unknown = undefined) {
+  const chain: Record<string, jest.Mock> = {};
+  const methods = [
+    'where', 'andWhere', 'select', 'first', 'insert', 'update', 'del',
+    'returning', 'join', 'orderBy',
+  ];
+  for (const method of methods) {
+    chain[method] = jest.fn().mockReturnValue(chain);
+  }
+  chain.first = jest.fn().mockResolvedValue(resolvedValue);
+  chain.select = jest.fn().mockResolvedValue(
+    Array.isArray(resolvedValue) ? resolvedValue : [resolvedValue],
+  );
+  chain.del = jest.fn().mockResolvedValue(1);
+  return chain;
+}
+
+jest.mock('../src/config/database', () => {
+  const handler = (tableName: string) => mockDb(tableName);
+  handler.raw = jest.fn();
+  handler.fn = { now: jest.fn() };
+  handler.transaction = jest.fn();
+  return { __esModule: true, default: handler };
+});
+
+jest.mock('../src/services/gmailSync', () => ({
+  syncUserGmail: jest.fn(),
+}));
+
+jest.mock('../src/services/gmailService', () => ({
+  gmailService: {
+    getAuthUrl: jest.fn(),
+    handleCallback: jest.fn(),
+    getStatus: jest.fn(),
+    disconnect: jest.fn(),
+  },
+}));
+
+import request from 'supertest';
+import jwt from 'jsonwebtoken';
+import app from '../src/app';
+import { syncUserGmail } from '../src/services/gmailSync';
+
+// ── Constants ────────────────────────────────────────────────────────────────
+const JWT_SECRET = process.env.JWT_SECRET || 'test-jwt-secret';
+const CRON_API_KEY = 'test-cron-key-secret';
+const USER_ID = '550e8400-e29b-41d4-a716-446655440000';
+const USER_EMAIL = 'test@example.com';
+const USER_NAME = 'Test User';
+
+function generateValidToken(userId: string, email: string, name: string): string {
+  return jwt.sign({ userId, email, name }, JWT_SECRET, { expiresIn: '15m' });
+}
+
+afterEach(() => {
+  jest.clearAllMocks();
+  delete process.env.CRON_API_KEY;
+});
+
+// =============================================================================
+// POST /api/gmail/sync
+// =============================================================================
+
+describe('POST /api/gmail/sync — cron path', () => {
+  it('should sync all users when called with a valid CRON_API_KEY', async () => {
+    process.env.CRON_API_KEY = CRON_API_KEY;
+
+    const gmailTokensChain = createQueryChain([{ user_id: USER_ID }]);
+    mockDb.mockImplementation((tableName: string) => {
+      if (tableName === 'gmail_tokens') return gmailTokensChain;
+      return createQueryChain(undefined);
+    });
+
+    (syncUserGmail as jest.Mock).mockResolvedValue({ scanned: 5, receiptsCreated: 1 });
+
+    const res = await request(app)
+      .post('/api/gmail/sync')
+      .set('Authorization', `Bearer ${CRON_API_KEY}`);
+
+    expect(res.status).toBe(200);
+    expect(res.body.results[USER_ID]).toMatchObject({ scanned: 5, receiptsCreated: 1 });
+    expect(syncUserGmail).toHaveBeenCalledWith(USER_ID);
+    expect(syncUserGmail).toHaveBeenCalledTimes(1);
+  });
+
+  it('should return 401 when CRON_API_KEY is wrong', async () => {
+    process.env.CRON_API_KEY = CRON_API_KEY;
+
+    const res = await request(app)
+      .post('/api/gmail/sync')
+      .set('Authorization', 'Bearer wrong-key');
+
+    // Wrong cron key is passed to authenticate as a JWT — rejected as invalid token
+    expect(res.status).toBe(401);
+    expect(res.body.error.code).toBe('ERR_INVALID_TOKEN');
+  });
+});
+
+describe('POST /api/gmail/sync — user path', () => {
+  it('should sync for the authenticated user when called with a valid JWT', async () => {
+    const token = generateValidToken(USER_ID, USER_EMAIL, USER_NAME);
+    (syncUserGmail as jest.Mock).mockResolvedValue({ scanned: 3, receiptsCreated: 0 });
+
+    const res = await request(app)
+      .post('/api/gmail/sync')
+      .set('Authorization', `Bearer ${token}`);
+
+    expect(res.status).toBe(200);
+    expect(res.body.data).toMatchObject({ scanned: 3, receiptsCreated: 0 });
+    expect(syncUserGmail).toHaveBeenCalledWith(USER_ID);
+  });
+
+  it('should return 401 when no token is provided', async () => {
+    const res = await request(app)
+      .post('/api/gmail/sync');
+
+    expect(res.status).toBe(401);
+    expect(res.body.error.code).toBe('ERR_NO_TOKEN');
+  });
+
+  it('should return 401 when JWT is invalid', async () => {
+    const res = await request(app)
+      .post('/api/gmail/sync')
+      .set('Authorization', 'Bearer invalid.jwt.token');
+
+    expect(res.status).toBe(401);
+    expect(res.body.error.code).toBe('ERR_INVALID_TOKEN');
+  });
+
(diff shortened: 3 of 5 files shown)


tomerp20 (commented):
## Backend Review — HIGH-RISK (auth middleware + OAuth route)

Reviewed: `backend/src/middleware/auth.ts`, `backend/src/routes/gmail.ts`, `backend/src/services/authService.ts`, `backend/tests/auth.test.ts`, `backend/tests/gmail.test.ts`.

**ESLint guardrail:** Not run locally in this review (worktree is on a different branch); CI should be the source of truth for lint.

### Verdict
Overall the refactor is clean and an improvement: the `syncAuth` middleware properly replaces the previous `new Promise(...)`-wrapped manual middleware call, and embedding identity claims in the JWT removes a per-request DB round-trip. The trade-offs are well-aligned with standard stateless-JWT practice (short 15-min access token + rotated refresh tokens). Test coverage for `/api/gmail/sync` cron and user paths is a meaningful gain. No CRITICAL or HIGH findings.

### Findings

| Severity | File:Line | Title |
|---|---|---|
| MEDIUM | `backend/src/middleware/auth.ts:88` | Token-rollout gap — legacy access tokens decode to `req.user.name = undefined` |
| MEDIUM | `backend/src/middleware/auth.ts:78` | `jwt.verify` cast trusts payload shape; missing runtime validation of required claims |
| LOW | `backend/src/middleware/auth.ts:33` | `JwtPayload` is exported but no longer documented; old JSDoc on `authenticate` (mentioned DB lookup) was deleted, no replacement |
| LOW | `backend/src/routes/gmail.ts:62` | `syncAuth` discriminator — wrong `CRON_API_KEY` falls through to JWT verification and surfaces `ERR_INVALID_TOKEN`, hiding the real failure mode in logs and ops dashboards |
| LOW | `backend/src/routes/gmail.ts:51` | `syncAuth` does not `await` the async `authenticate` call — works today (Express never awaits middleware) but loses any unexpected rejected promise; consider documenting or wrapping for clarity |
| LOW | `backend/tests/gmail.test.ts:1` | Tests over-mock the DB (mocks `users` table reads that `authenticate` no longer performs); harmless but stale relative to the new middleware |
| LOW | `backend/tests/gmail.test.ts:88` | Missing test: cron path with `CRON_API_KEY` env unset (current behavior: falls through to JWT auth → 401). Worth pinning. |
| LOW | `backend/tests/gmail.test.ts:1` | Missing test: cron path with mixed `gmail_tokens` results (e.g., 0 valid tokens → empty results, 2+ valid tokens → per-user mapping). Confirms the loop body is exercised. |

### Architectural notes (not findings, FYI)
- **Account-revocation window:** Removing the `db('users')` lookup means a deleted/disabled user retains access for up to 15 min until the access token expires. This is the standard stateless-JWT trade-off and is acceptable, but it should be a conscious decision documented somewhere (ADR or auth README) so it isn't a surprise during a future security review.
- **Stale `name` in JWT:** If a user renames themselves, `req.user.name` stays stale until the next access-token rotation (≤15 min). Acceptable; mention in the same doc.
- **N+1 in cron loop (`gmail.ts:59-61`):** Pre-existing, intentional (sequential calls cap Gmail API rate). Not a regression in this PR. Worth converting to bounded concurrency in a follow-up.

Inline comments follow with code-level detail.

backend/src/middleware/auth.ts
```diff
@@ -91,26 +83,10 @@ export async function authenticate(
...
-      });
-      return;
-    }
-
     req.user = {
-      id: user.id,
-      email: user.email,
-      name: user.name,
+      id: decoded.userId,
+      email: decoded.email,
```
tomerp20, line 88: **[MEDIUM] — Token-rollout gap: legacy tokens decode to `name: undefined`**

Access tokens issued before this PR ships do not contain a `name` claim. After deploy they will continue to be accepted by `jwt.verify` (still cryptographically valid) and `decoded.name` will be `undefined`, so `req.user.name` becomes `undefined` for those requests until the token expires (max 15 min).

Downstream code that does `req.user!.name` (e.g. UI greeting, audit log enrichment) will silently get `undefined` instead of a string — the `Express.Request.user` type promises `name: string`.

**Current:**
```ts
req.user = {
  id: decoded.userId,
  email: decoded.email,
  name: decoded.name,
};
```

**Fix (defensive, self-healing as tokens rotate):**
```ts
if (typeof decoded.userId !== 'string' || typeof decoded.email !== 'string' || typeof decoded.name !== 'string') {
  res.status(401).json({
    error: { message: 'Invalid token', code: 'ERR_INVALID_TOKEN' },
  });
  return;
}

req.user = {
  id: decoded.userId,
  email: decoded.email,
  name: decoded.name,
};
```

This forces any legacy or malformed token to be re-issued via the refresh-token flow rather than circulating a `req.user` shape that lies about its type.

tomerp20:
## Security Review — PR #177

**Reviewer:** code-review-orchestrator (security pass)

### Findings

#### ✅ JWT signing — no issues
- Token is signed with HS256 + `JWT_SECRET`. Adding `name` to the payload doesn't change the signing algorithm or expose the secret.
- Signature verification still runs before `req.user` is populated — tampering is detected.
- Expiry (15 min) still enforced.

#### ✅ CRON_API_KEY comparison — timing-safe
- `safeCompare` hashes both sides with SHA-256 then uses `timingSafeEqual` — resistant to timing side-channels.
- If `CRON_API_KEY` is unset or empty, `syncAuth` falls through to `authenticate` (falsy guard `if (cronKey && ...)` is correct).

#### ✅ Trust boundary — name comes from signed payload
- `decoded.name` is only reachable if the JWT signature is valid. A user cannot inject a crafted name without knowing `JWT_SECRET`.
- `name` is returned as JSON, not raw HTML — XSS risk lies with the frontend renderer, not this middleware.

#### LOW — Deleted-account window (accepted trade-off)
Valid tokens for deleted users will authenticate for up to 15 minutes after deletion. The issue explicitly calls this out as an accepted trade-off for a single-user system. No action required; document it in the PR description if not already noted.

#### LOW — name staleness acknowledged
The issue notes `name` never changes in this application. If that invariant ever breaks, tokens would carry stale names until the 15-min access token rotates. No action required for current scope.

### Summary
No HIGH or MEDIUM security findings. Two LOW items are both explicitly accepted by the issue author. This change **reduces** attack surface by removing a DB call that could have been a vector for timing-based user enumeration.

backend/src/middleware/auth.ts
```diff
@@ -12,25 +10,19 @@ declare global {
...
       };
+      isCronRequest?: boolean;
     }
   }
 }
 
 export interface JwtPayload {
   userId: string;
   email: string;
+  name: string;
```
tomerp20, line 21: **[MEDIUM] — `jwt.verify` cast trusts payload shape without runtime checking**

Adding \`name: string\` to \`JwtPayload\` is correct, but the cast at the verify site (\`jwt.verify(token, secret) as JwtPayload\`) is a TypeScript-only narrowing — \`jwt.verify\` returns \`string | JwtPayload\` (the library's own), and any object signed with the same secret will be accepted, including ones missing \`userId\`, \`email\`, or \`name\`, or where they are not strings.

Since \`authenticate\` now relies entirely on the token payload (no DB cross-check), a malformed or partially-claimed token will produce a broken \`req.user\` instead of a 401.

**Suggested:** at the assignment site (line 88), runtime-check the three required string claims before building \`req.user\`. Combine with the line-88 fix:

\`\`\`ts
const raw = jwt.verify(token, secret);
if (
  typeof raw !== 'object' || raw === null ||
  typeof (raw as Partial<JwtPayload>).userId !== 'string' ||
  typeof (raw as Partial<JwtPayload>).email  !== 'string' ||
  typeof (raw as Partial<JwtPayload>).name   !== 'string'
) {
  res.status(401).json({ error: { message: 'Invalid token', code: 'ERR_INVALID_TOKEN' } });
  return;
}
const decoded = raw as JwtPayload;
\`\`\`

This converts the cast into a real guard and makes the new no-DB-lookup design safe against any future change to token payload shape.

backend/src/middleware/auth.ts
```diff
@@ -12,25 +10,19 @@ declare global {
         email: string;
         name: string;
       };
```
tomerp20, line 12: **[LOW] — Lost documentation on `authenticate`**

The previous JSDoc that described the middleware's contract ("verifies JWT, looks up user, attaches { id, email, name } to req.user; returns 401 if missing/expired/user-not-found") was deleted with no replacement. With the DB lookup removed, the contract has materially changed — please add a fresh 3–4 line JSDoc on `authenticate` that calls out:

- It is **stateless**: a valid unexpired token IS the proof of identity (no DB check).
- A user deletion does **not** invalidate already-issued access tokens until they expire (max 15 min).
- Returns 401 for: missing token, missing Bearer prefix, missing JWT_SECRET (500), invalid signature, expired token.

This is the kind of design note that prevents future regressions in security reviews.

backend/src/routes/gmail.ts
```diff
@@ -47,13 +47,22 @@ router.delete('/disconnect', authenticate, async (req: Request, res: Response, n
...
-  try {
-    const cronKey = process.env.CRON_API_KEY;
-    const authHeader = req.headers.authorization ?? '';
-    const isCron = !!cronKey && safeCompare(authHeader, `Bearer ${cronKey}`);
+function syncAuth(req: Request, res: Response, next: NextFunction): void {
+  const cronKey = process.env.CRON_API_KEY;
+  const authHeader = req.headers.authorization ?? '';
+
+  if (cronKey && safeCompare(authHeader, `Bearer ${cronKey}`)) {
+    req.isCronRequest = true;
```
tomerp20, line 56: **[LOW] — `syncAuth` blurs cron-vs-user failure modes**

Right now, a wrong `CRON_API_KEY` falls through to `authenticate` and is rejected as `ERR_INVALID_TOKEN` (the new `gmail.test.ts` even asserts this). In logs / dashboards a misconfigured cron job is then indistinguishable from a regular user with a bad JWT — a real ops concern when you need to debug why scheduled syncs stopped working.

**Suggested:** treat a header that looks like a cron attempt as cron-only. Options:

1. Use a dedicated header (`X-Cron-Token`) for cron and `Authorization: Bearer` for users — no overlap, no fall-through.
2. Or, keep the current scheme but emit a `logger.warn({ route: '/gmail/sync' }, 'cron key mismatch — falling through to JWT auth')` when a bearer was present and didn't match either path, so ops can grep for it.

Not blocking — flagging because the test explicitly pins the confusing behavior.

backend/src/routes/gmail.ts
```diff
@@ -47,13 +47,22 @@ router.delete('/disconnect', authenticate, async (req: Request, res: Response, n
...
+    next();
+    return;
+  }
+
+  authenticate(req, res, next);
+}
```
tomerp20, line 66: **[LOW] — `syncAuth` calls async `authenticate` without awaiting**

`authenticate` is `async` and returns `Promise<void>`. `syncAuth` calls it but does not return or await the promise. This works today because Express's router never inspects the middleware return value — response/next handling goes through the response object and the `next` callback. But it means any unhandled rejection inside `authenticate` (e.g. a future code path that throws before reaching the inner try/catch) will be swallowed by Node's unhandled-rejection handler instead of bubbling through Express's error pipeline.

**Current:**
```ts
authenticate(req, res, next);
```

**Defensive fix:**
```ts
authenticate(req, res, next).catch(next);
```

Cheap and makes the contract explicit. Optional — skip if you'd rather rely on the existing outer try/catch in `authenticate`.

backend/tests/gmail.test.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    (syncUserGmail as jest.Mock).mockResolvedValue({ scanned: 5, receiptsCreated: 1 });
+
+    const res = await request(app)
+      .post('/api/gmail/sync')
+      .set('Authorization', `Bearer ${CRON_API_KEY}`);
+
+    expect(res.status).toBe(200);
+    expect(res.body.results[USER_ID]).toMatchObject({ scanned: 5, receiptsCreated: 1 });
+    expect(syncUserGmail).toHaveBeenCalledWith(USER_ID);
```
tomerp20, line 86: **[LOW] — Missing edge-case test: `CRON_API_KEY` env unset**

The cron tests set `process.env.CRON_API_KEY` per-test and `afterEach` deletes it. There is no test pinning what happens when `CRON_API_KEY` is unset at request time (e.g. local dev without a cron secret configured). Today that path falls straight through to `authenticate` and a request whose bearer value matches whatever the secret *would have been* is treated as a bad JWT. A regression here would be silent.

**Suggested test:**
```ts
it('falls through to JWT auth when CRON_API_KEY is unset', async () => {
  delete process.env.CRON_API_KEY;
  const res = await request(app)
    .post('/api/gmail/sync')
    .set('Authorization', `Bearer ${CRON_API_KEY}`);
  expect(res.status).toBe(401);
  expect(res.body.error.code).toBe('ERR_INVALID_TOKEN');
});
```

backend/tests/gmail.test.ts
```diff
@@ -0,0 +1,150 @@
...
+}
+
+afterEach(() => {
+  jest.clearAllMocks();
+  delete process.env.CRON_API_KEY;
+});
+
+// =============================================================================
+// POST /api/gmail/sync
+// =============================================================================
```
tomerp20, line 66: **[LOW] — Missing edge-case tests on the cron loop**

Only one cron-success test exists and it has exactly one valid `gmail_tokens` row. Worth adding two cases so the loop body is properly pinned:

1. **Empty:** `gmail_tokens.where({ is_valid: true }).select()` returns `[]` → response is `{ results: {} }` and `syncUserGmail` is never called.
2. **Multiple users:** returns two `user_id` rows → both keys appear under `results`, `syncUserGmail` called twice with the right args.

These guard against a future refactor (e.g., bounded concurrency, retry-on-failure, partial-success handling) silently breaking iteration semantics.

tomerp20:
## Code Review — Final Status

| Agent | Issues found | Fixed | Skipped |
| Security (orchestrator) | 2 LOW | — | 2 (accepted trade-offs) |
| Backend reviewer | 2 MEDIUM, 5 LOW | 4 LOW | 2 MEDIUM + 1 LOW (manual) |

---

### Fixed automatically (fix commit [c89e30c](https://github.com/tomerp20/jobflow/commit/c89e30c))

| # | Severity | Finding |
|---|----------|---------|
| 4 | LOW | Added `logger.warn` in `syncAuth` when CRON_API_KEY is set but doesn't match — ops can now distinguish a misconfigured cron from a bad user JWT |
| 5 | LOW | Added `.catch(next)` to the `authenticate()` call in `syncAuth` as a safety net against future unhandled rejections escaping the middleware's own catch block |
| 6 | LOW | Added test: `CRON_API_KEY` unset — confirms bearer token falls through to JWT auth and returns `ERR_INVALID_TOKEN` |
| 7 | LOW | Added tests: empty cron-loop (no gmail tokens → empty results, `syncUserGmail` not called) and multi-user cron-loop (2 users synced independently) |

---

### Requires manual attention

| # | Severity | Finding |
|---|----------|---------|
| 1 | MEDIUM | **Legacy token stale `name`:** access tokens issued before this deploy will decode to `req.user.name = undefined` despite `Express.Request.user.name: string`. Safe to ignore — 15-min expiry means all pre-deploy tokens rotate within one window. Consider documenting the stateless JWT contract in an auth README or ADR. |
| 2 | MEDIUM | **No runtime payload validation after `jwt.verify`:** `jwt.verify(token, secret) as JwtPayload` is a TypeScript cast only. If a valid but malformed token (e.g. missing `userId`, `email`, or `name`) reaches `authenticate`, `req.user` will have `undefined` fields instead of returning 401. Fix: add a runtime guard after verify — e.g. `if (!decoded.userId || !decoded.email || !decoded.name)` → 401. |
| 3 | LOW | **Missing JSDoc on `authenticate`:** the old comment describing the DB lookup was removed with nothing replacing it. The middleware's contract changed materially (stateless, 15-min revocation window, no user-exists check). Worth a 3-line doc so future readers don't wonder why there's no DB call. |

---

### Security
No HIGH or MEDIUM security issues. Two LOW items (deleted-account 15-min window, name staleness) are accepted trade-offs documented in the issue.

**Workflow complete. Do not merge — awaiting human review.**

Issue #189, opened 2026-05-23
[Cassandra pipeline PR1.2] Fetcher base: --hour and --range modes


tomerp20:
## Parent PRD

#187

## What to build

A new standalone Node package at `data-pipeline/fetcher/` that downloads GH Archive hourly `.json.gz` files to a configurable on-disk root. This slice ships the entire Fetcher except for `--catchup` mode — that is the follow-up slice. All cross-cutting behaviours that every Fetcher mode needs (atomic writes, `.partial` cleanup, retries, idempotency, structured logging, env-based configuration, bounded concurrency) ship here.

The package is split into four modules per the parent PRD's **Fetcher module split** section: `disk-layout` (deep, pure functions), `download` (HTTP + atomic write + retries), `cli` (pure argv parsing), and the `fetcher.js` entry point that wires them together. The Fetcher does not talk to Cassandra in any way.

This slice is independent of #188 — they can be developed in parallel and land in either order.

## Acceptance criteria

- [ ] New Node package at `data-pipeline/fetcher/` with `"engines": ">=20.6.0"` in `package.json`; runtime deps limited to `pino` and `p-limit`
- [ ] Environment loaded via `node --env-file=.env`; `.env.example` committed with `GHARCHIVE_DIR=./data/gharchive` and `LOG_LEVEL=info`
- [ ] Pino logger streams to stdout: `pino-pretty` transport when `NODE_ENV !== 'production'`, raw JSON otherwise; honours `LOG_LEVEL`
- [ ] Disk-layout module converts between canonical hour ID `YYYY-MM-DD-H` (hour unpadded) and on-disk path `<root>/YYYY/MM/DD/H.json.gz` (month and day zero-padded); enumerates a date range inclusive on both bounds; sweeps and removes every `.partial` file under the root
- [ ] Download module performs atomic writes via `.partial` → final-name rename; returns a discriminated status (`written`, `skipped-already-present`, `not-found`, `failed-after-retries`); retries HTTP 5xx and network errors with exponential backoff up to a small fixed retry budget; treats an already-present final file as a no-op (no network call)
- [ ] CLI module accepts `--hour <YYYY-MM-DD-H>` with strict regex `^\d{4}-\d{2}-\d{2}-\d{1,2}$` and `--range <A> <B>` with both bounds inclusive; rejects invalid input with a single-line error and a non-zero exit
[...]

PR #195, opened 2026-05-23
[Cassandra pipeline PR1.3] Fetcher: --catchup mode

### LINKED ISSUE #190: [Cassandra pipeline PR1.3] Fetcher: --catchup mode
## Parent PRD

#187

## What to build

Add the third and final Fetcher mode — `--catchup` — on top of the Fetcher base shipped in #189. Catchup is the no-argument mode that the eventual hourly cron (deferred to PR3) will invoke. It walks forward from the latest hour already on disk toward `now − 2h` (GH Archive publishes with a ~1–2h delay) and stops at the first 404.

Three locked behaviours from the PR1 grill that this slice must implement exactly:

1. **Cold-start refuses.** Running `--catchup` against an empty `GHARCHIVE_DIR` (no anchor file) exits non-zero with a clear error and makes no network calls. Silent fallback could trigger an unintended large download.
2. **First-404 stops the walk.** GH Archive publishes monotonically; once a forward step returns 404, every later hour also returns 404. The walk must halt at the first 404 rather than continuing to spam ~24 requests on tomorrow's hours.
3. **No staleness cap.** If the latest-on-disk anchor is N hours old, all N hours are fetched (subject to the first-404 rule). A multi-week outage self-heals on the next successful catchup.

## Acceptance criteria

- [ ] Disk-layout module gains a `latestOnDisk(rootDir)` function that returns the most recent hour ID found under the root, or `null` if the root is empty
- [ ] CLI module recognises bare `--catchup` (no extra arguments) as a valid mode
- [ ] `fetcher.js` dispatches `--catchup` to a walk that starts at `latestOnDisk + 1 hour` and steps forward in 1-hour increments toward `now − 2h` (UTC)
- [ ] Cold start: when `latestOnDisk` returns `null`, log a single clear "no anchor to walk from" message and exit non-zero without making any network requests
- [ ] First HTTP 404 encountered during the walk halts the walk immediately; no further requests are issued for any later hour
- [ ] Concurrency is still bounded by `p-limit(3)` — but the first-404 short-circuit must work correctly even when several requests are in flight (in-flight requests for hours past the 404 are awaited but their results do not cause additional walks)
- [ ] All existing behaviours from #189 (atomic writes, `.partial` cleanup, retries, idempotency, logging, env config) continue to apply
- [ ] Manual smoke test passes: `--catchup` against an empty directory exits 1 with the cold-start message; `--catchup` against a populated directory walks forward and stops cleanly at the first 404 boundary

## Blocked by

- Blocked by #189

## User stories addressed

- User story 5
- User story 6
- User story 7
- User story 8

tomerp20:
## Summary
- Adds `latestOnDisk(rootDir)` to `disk-layout.js` — recursively scans the archive root and returns the most recent hour ID, or `null` if empty
- Adds `--catchup` as a valid CLI mode (bare flag, no extra args)
- `runCatchup` walks from `latestOnDisk + 1h` toward `now − 2h` in 1-hour steps with `p-limit(3)` concurrency; cold-start (null anchor) exits 1 with a clear message; first HTTP 404 sets a shared flag that prevents queued tasks from issuing further HTTP requests while in-flight requests are awaited

## Acceptance criteria
- [x] Disk-layout module gains a `latestOnDisk(rootDir)` function that returns the most recent hour ID found under the root, or `null` if the root is empty
- [x] CLI module recognises bare `--catchup` (no extra arguments) as a valid mode
- [x] `fetcher.js` dispatches `--catchup` to a walk that starts at `latestOnDisk + 1 hour` and steps forward in 1-hour increments toward `now − 2h` (UTC)
- [x] Cold start: when `latestOnDisk` returns `null`, log a single clear "no anchor to walk from" message and exit non-zero without making any network requests
- [x] First HTTP 404 encountered during the walk halts the walk immediately; no further requests are issued for any later hour
- [x] Concurrency is still bounded by `p-limit(3)` — but the first-404 short-circuit works correctly even when several requests are in flight (in-flight requests for hours past the 404 are awaited but their results do not cause additional walks)
- [x] All existing behaviours from #189 (atomic writes, `.partial` cleanup, retries, idempotency, logging, env config) continue to apply
- [x] Manual smoke test passes: `--catchup` against an empty directory exits 1 with the cold-start message; walk candidate logic verified with seeded temp directories

## Related
Closes #190

### DIFF
--- a/data-pipeline/fetcher/fetcher.js
+++ b/data-pipeline/fetcher/fetcher.js
@@ -1,7 +1,7 @@
 import pino from 'pino';
 import pLimit from 'p-limit';
 import { parseCLI } from './lib/cli.js';
-import { cleanPartials, enumerateRange } from './lib/disk-layout.js';
+import { cleanPartials, enumerateRange, latestOnDisk } from './lib/disk-layout.js';
 import { downloadHour } from './lib/download.js';
 
 const ROOT = process.env.GHARCHIVE_DIR ?? './data/gharchive';
@@ -19,6 +19,11 @@ async function main() {
 
   logger.info({ root: ROOT }, 'cleaning partial files');
   cleanPartials(ROOT);
+
+  if (cmd.mode === 'catchup') {
+    await runCatchup(ROOT, logger);
+    return;
+  }
 
   const hours =
     cmd.mode === 'hour'
@@ -56,6 +61,77 @@ async function main() {
     {}
   );
   logger.info(tally, 'done');
+}
+
+async function runCatchup(root, log) {
+  const anchor = latestOnDisk(root);
+  if (!anchor) {
+    log.error('no anchor to walk from — run --hour or --range first to seed the directory');
+    process.exit(1);
+  }
+
+  const ceilingMs = Date.now() - 2 * 3_600_000;
+  const anchorMs = _hourIdToMs(anchor);
+
+  const candidates = [];
+  for (let t = anchorMs + 3_600_000; t <= ceilingMs; t += 3_600_000) {
+    candidates.push(_msToHourId(t));
+  }
+
+  if (candidates.length === 0) {
+    log.info({ anchor }, 'already up to date');
+    return;
+  }
+
+  log.info({ count: candidates.length, from: candidates[0], to: candidates[candidates.length - 1] }, 'starting catchup walk');
+
+  const limit = pLimit(3);
+  let found404 = false;
+
+  // All candidates are enqueued; p-limit starts up to 3 concurrently.
+  // The found404 flag prevents new HTTP requests from being issued after the first 404;
+  // any already-in-flight requests are awaited but their results don't extend the walk.
+  const promises = candidates.map((hourId) =>
+    limit(async () => {
+      if (found404) return { status: 'skipped-catchup-abort', hourId };
+      const result = await downloadHour(root, hourId);
+      if (result.status === 'not-found') {
+        found404 = true;
+        log.warn({ hourId }, 'first 404 — halting catchup walk');
+      } else if (result.status === 'written') {
+        log.info({ hourId }, 'written');
+      } else if (result.status === 'failed-after-retries') {
+        log.error({ hourId, error: result.error }, 'failed after retries');
+      }
+      return result;
+    })
+  );
+
+  const settled = await Promise.allSettled(promises);
+  const results = settled.map((s) =>
+    s.status === 'fulfilled'
+      ? s.value
+      : { status: 'failed-after-retries', error: s.reason?.message }
+  );
+  const tally = results.reduce(
+    (acc, r) => { acc[r.status] = (acc[r.status] ?? 0) + 1; return acc; },
+    {}
+  );
+  log.info(tally, 'catchup done');
+}
+
+function _hourIdToMs(hourId) {
+  const [y, m, d, h] = hourId.split('-').map(Number);
+  return Date.UTC(y, m - 1, d, h);
+}
+
+function _msToHourId(ms) {
+  const d = new Date(ms);
+  const year = d.getUTCFullYear();
+  const month = String(d.getUTCMonth() + 1).padStart(2, '0');
+  const day = String(d.getUTCDate()).padStart(2, '0');
+  const hour = d.getUTCHours();
+  return `${year}-${month}-${day}-${hour}`;
 }
 
 main().catch((err) => {
--- a/data-pipeline/fetcher/fetcher.js
+++ b/data-pipeline/fetcher/fetcher.js
@@ -1,7 +1,7 @@
 import pino from 'pino';
 import pLimit from 'p-limit';
 import { parseCLI } from './lib/cli.js';
-import { cleanPartials, enumerateRange, latestOnDisk } from './lib/disk-layout.js';
+import { cleanPartials, enumerateRange, hourIdToMs, latestOnDisk, msToHourId } from './lib/disk-layout.js';
 import { downloadHour } from './lib/download.js';
 
 const ROOT = process.env.GHARCHIVE_DIR ?? './data/gharchive';
@@ -66,16 +66,17 @@ async function main() {
 async function runCatchup(root, log) {
   const anchor = latestOnDisk(root);
   if (!anchor) {
-    log.error('no anchor to walk from — run --hour or --range first to seed the directory');
-    process.exit(1);
+    // Throw so the main().catch handler exits consistently with other fatal errors
+    // and pino's transport gets a chance to flush before exit.
+    throw new Error('no anchor to walk from — run --hour or --range first to seed the directory');
   }
 
   const ceilingMs = Date.now() - 2 * 3_600_000;
-  const anchorMs = _hourIdToMs(anchor);
+  const anchorMs = hourIdToMs(anchor);
 
   const candidates = [];
   for (let t = anchorMs + 3_600_000; t <= ceilingMs; t += 3_600_000) {
-    candidates.push(_msToHourId(t));
+    candidates.push(msToHourId(t));
   }
 
   if (candidates.length === 0) {
@@ -118,20 +119,6 @@ async function runCatchup(root, log) {
     {}
   );
   log.info(tally, 'catchup done');
-}
-
-function _hourIdToMs(hourId) {
-  const [y, m, d, h] = hourId.split('-').map(Number);
-  return Date.UTC(y, m - 1, d, h);
-}
-
-function _msToHourId(ms) {
-  const d = new Date(ms);
-  const year = d.getUTCFullYear();
-  const month = String(d.getUTCMonth() + 1).padStart(2, '0');
-  const day = String(d.getUTCDate()).padStart(2, '0');
-  const hour = d.getUTCHours();
-  return `${year}-${month}-${day}-${hour}`;
 }
 
 main().catch((err) => {
--- a/data-pipeline/fetcher/lib/cli.js
+++ b/data-pipeline/fetcher/lib/cli.js
@@ -27,7 +27,14 @@ export function parseCLI(argv) {
     return { mode: 'range', start, end };
   }
 
-  die('usage: fetcher.js --hour <YYYY-MM-DD-H> | --range <YYYY-MM-DD> <YYYY-MM-DD>');
+  if (args[0] === '--catchup') {
+    if (args.length > 1) {
+      die('--catchup takes no arguments');
+    }
+    return { mode: 'catchup' };
+  }
+
+  die('usage: fetcher.js --hour <YYYY-MM-DD-H> | --range <YYYY-MM-DD> <YYYY-MM-DD> | --catchup');
 }
 
 function die(msg) {
--- a/data-pipeline/fetcher/lib/disk-layout.js
+++ b/data-pipeline/fetcher/lib/disk-layout.js
@@ -36,6 +36,36 @@ export function enumerateRange(startDate, endDate) {
   return hours;
 }
 
+// Returns the most recent hour ID (YYYY-MM-DD-H) found under rootDir, or null if empty.
+export function latestOnDisk(rootDir) {
+  const ids = [];
+  _collectHourIds(rootDir, rootDir, ids);
+  if (ids.length === 0) return null;
+  return ids.reduce((best, cur) => _hourIdToMs(cur) > _hourIdToMs(best) ? cur : best);
+}
+
+function _collectHourIds(root, dir, ids) {
+  let entries;
+  try {
+    entries = readdirSync(dir, { withFileTypes: true });
+  } catch {
+    return;
+  }
+  for (const entry of entries) {
+    const full = path.join(dir, entry.name);
+    if (entry.isDirectory()) {
+      _collectHourIds(root, full, ids);
+    } else if (entry.name.endsWith('.json.gz')) {
+      ids.push(pathToHourId(root, full));
+    }
+  }
+}
+
+function _hourIdToMs(hourId) {
+  const [y, m, d, h] = hourId.split('-').map(Number);
+  return Date.UTC(y, m - 1, d, h);
+}
+
 // Recursively removes every *.partial file under root. Silently skips missing directories.
 export function cleanPartials(root) {
   _sweepDir(root);
--- a/data-pipeline/fetcher/lib/disk-layout.js
+++ b/data-pipeline/fetcher/lib/disk-layout.js
@@ -41,7 +41,7 @@ export function latestOnDisk(rootDir) {
   const ids = [];
   _collectHourIds(rootDir, rootDir, ids);
   if (ids.length === 0) return null;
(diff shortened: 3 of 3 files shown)


tomerp20 (commented):
## Code Review — PR #195: Fetcher `--catchup` mode

**Reviewer:** Code Review Orchestrator (handling directly — files under `data-pipeline/fetcher/` are pure Node.js/ESM and fall outside both the `react-code-reviewer` and `backend-code-reviewer` specialist domains. See orchestrator memory note for PR #176 — same fallback pattern.)

**Scope:** `data-pipeline/fetcher/fetcher.js`, `data-pipeline/fetcher/lib/cli.js`, `data-pipeline/fetcher/lib/disk-layout.js` (+115 / -2).

---

### Overall assessment
This is a clean, well-scoped change that matches the acceptance criteria 1:1. The catchup walk semantics (anchor + 1h → now − 2h, bounded by `p-limit(3)`, first-404 short-circuit via shared flag, cold-start non-zero exit) are implemented correctly. JS single-threaded execution makes the `found404` flag inherently safe — the in-code comment explains this accurately. No security issues found. **Approval candidate** with a few low-severity polish notes below.

---

### 🔒 Security review
No security issues found across the three files.

- **No secrets / credentials** introduced.
- **No injection surface.** `runCatchup` only feeds `downloadHour` with canonical hour IDs reconstructed from a numeric ms timestamp via `_msToHourId` — never user-controlled strings. The downstream URL template `${GH_ARCHIVE_BASE}/${hourId}.json.gz` is therefore safe.
- **No SSRF.** Base URL is hardcoded to `https://data.gharchive.org`; only the path segment varies, and that path segment is server-controlled (computed from `Date.UTC` arithmetic).
- **No path traversal.** `pathToHourId` operates on the result of `readdirSync(root, { withFileTypes: true })` joined via `path.join`. Even if an attacker dropped a symlink with `../` in its name, `path.join` normalises, and the downstream usage is to compute hour IDs for URL fetches, not to read/write that path. The reconstructed `finalPath` in `downloadHour` is itself rebuilt via `hourIdToPath` from a clean canonical hour ID.
- **No fs writes from CLI input.** Catchup writes only happen to paths derived from `Date.UTC(y, m-1, d, h)` — i.e. always inside the configured root.

---

### Correctness findings

#### MEDIUM — `latestOnDisk` is not robust to stray `*.json.gz` files at the wrong depth
**File:** `data-pipeline/fetcher/lib/disk-layout.js:39-62`

`_collectHourIds` recursively collects any file ending in `.json.gz`, regardless of directory depth. `pathToHourId` assumes exactly the canonical `root/YYYY/MM/DD/H.json.gz` layout. If a stray `.json.gz` lives at, e.g., `root/something.json.gz` or `root/Y/M/D/extra/X.json.gz`, `pathToHourId` returns a malformed ID (`undefined-undefined-undefined-NaN` or similar). That ID then flows into `_hourIdToMs`, producing `NaN`.

`reduce((best, cur) => _hourIdToMs(cur) > _hourIdToMs(best) ? cur : best)` happens to survive a single garbage entry (because `NaN > anyMs` is `false`), but it does **not** survive the case where the garbage entry is the *first* (seed) element — `best` then starts as the garbage ID and `_hourIdToMs(NaN) > _hourIdToMs(valid)` is `false`, so the first garbage entry stays as `best` forever. Result: `latestOnDisk` could return a malformed string to `runCatchup`, which then produces `anchorMs = NaN`, the candidate loop never iterates, and the walk silently exits as "already up to date" — a **silent skip** that is very hard to diagnose.

**Improved code (defensive filter):**
```javascript
// Constrain to the canonical depth so stray .json.gz files (or future side-files)
// can't poison the anchor.
function _collectHourIds(root, dir, ids) {
  let entries;
  try { entries = readdirSync(dir, { withFileTypes: true }); } catch { return; }
  for (const entry of entries) {
    const full = path.join(dir, entry.name);
    if (entry.isDirectory()) {
      _collectHourIds(root, full, ids);
    } else if (entry.name.endsWith('.json.gz')) {
      const rel = path.relative(root, full).split(path.sep);
      if (rel.length !== 4) continue; // not at root/Y/MM/DD/H.json.gz depth
      ids.push(pathToHourId(root, full));
    }
  }
}
```

Cost: zero — single integer compare. Removes a real silent-failure mode.

#### LOW — Duplicated `_hourIdToMs` / `_msToHourId` helpers
**Files:** `fetcher.js:123-135`, `disk-layout.js:64-67`

Both modules now carry private copies of `_hourIdToMs`. `_msToHourId` in `fetcher.js` duplicates inline logic already present in `enumerateRange`. Single source of truth would be cleaner — export `hourIdToMs` / `msToHourId` from `disk-layout.js` and import them in `fetcher.js`. Not a bug, just future-proofing: the hour-arithmetic logic is the kind of thing that drifts when only one copy gets a fix.

#### LOW — Log ordering after first-404
**File:** `fetcher.js:94-108`

When the 404 happens for hour N, up to 2 other requests may already be in flight for hours > N. If those return `'written'` after the 404, the log will read:
```
warn first 404 — halting catchup walk (hour N+1)
info written (hour N+2)
```
which looks like a contradiction. Behavioural impact: none (the tally is honest, the data on disk is real). Diagnostics impact: small but non-zero. Optional: add a `{ stale: true }` field to the log line when `found404` was set before this result was logged, or simply demote those `info written` lines to `debug` once `found404` is set.

#### LOW — `process.exit(1)` from cold-start may truncate buffered pino output
**File:** `fetcher.js:69-70`

`pino` with the `pino-pretty` transport (the non-prod path here) uses a worker thread; `process.exit` bypasses its flush. The cold-start message *usually* gets through because it's a single short log line, but on slower machines it can be lost. Optional mitigation:
```javascript
log.error('no anchor to walk from — run --hour or --range first to seed the directory');
log.flush?.();
await new Promise((r) => setImmediate(r));
process.exit(1);
```
Or, more idiomatic for pino's transport pipeline, throw and let the `main().catch` handler exit:
```javascript
throw new Error('no anchor to walk from — run --hour or --range first to seed the directory');
```
The existing `main().catch((err) => { logger.error(err, 'fatal error'); process.exit(1); })` would handle it consistently with other fatal errors.

#### LOW — CLI: `--catchup foo` error message could be more informative
**File:** `data-pipeline/fetcher/lib/cli.js:31`

`die('--catchup takes no arguments')` doesn't echo what was actually passed. Minor: `die(\`--catchup takes no arguments (got: ${args.slice(1).join(' ')})\`)` makes copy-paste mistakes easier to debug. Skip if not aligned with the project's other CLI error styles.

---

### Test coverage observation
The PR description references a manual smoke test only. The fetcher package has no `test` script in `package.json` and no `tests/` directory. Given that `latestOnDisk` and the catchup candidate loop are pure functions (the latter takes `now()` implicitly but is otherwise deterministic), they are unusually amenable to unit testing. **Not a blocker for this PR** since the acceptance criteria don't require tests, but worth flagging as the next slice — a regression here would only surface in production.

---

### Acceptance criteria check
| Criterion | Status |
|-----------|--------|
| `latestOnDisk(rootDir)` returns most recent hour ID or null | ✅ |
| `--catchup` bare flag recognised by CLI | ✅ |
| Catchup walk from `latestOnDisk + 1h` toward `now − 2h` UTC, 1h steps | ✅ |
| Cold-start exits 1 with clear message, no network calls | ✅ |
| First 404 halts walk; no further requests issued | ✅ |
| `p-limit(3)` concurrency preserved | ✅ |
| Existing #189 behaviours (atomic writes, retries, `.partial` cleanup) preserved | ✅ |
| Manual smoke test described | ✅ |

---

### Recommendation
**Approve after addressing the MEDIUM finding on `latestOnDisk` depth-validation.** The LOW findings are polish — apply or defer at the author's discretion.

data-pipeline/fetcher/lib/disk-layout.js
```diff
@@ -36,6 +36,36 @@ export function enumerateRange(startDate, endDate) {
...
+    entries = readdirSync(dir, { withFileTypes: true });
+  } catch {
+    return;
+  }
+  for (const entry of entries) {
+    const full = path.join(dir, entry.name);
+    if (entry.isDirectory()) {
+      _collectHourIds(root, full, ids);
+    } else if (entry.name.endsWith('.json.gz')) {
+      ids.push(pathToHourId(root, full));
```
tomerp20, line 65: **[MEDIUM] Not robust to stray `*.json.gz` files at the wrong depth.** `_collectHourIds` recursively collects any `.json.gz` file regardless of depth, but `pathToHourId` assumes exactly `root/YYYY/MM/DD/H.json.gz`. A stray file at any other depth produces a malformed hour ID; `_hourIdToMs` then yields `NaN`. If the malformed entry is the *first* element of `ids`, the `reduce` seed stays as the garbage value forever — `runCatchup` silently exits as 'already up to date' and the cron walk goes dark.\n\nDefensive fix: constrain to canonical 4-segment depth before pushing.\n```js\nconst rel = path.relative(root, full).split(path.sep);\nif (rel.length !== 4) continue;\nids.push(pathToHourId(root, full));\n```

data-pipeline/fetcher/fetcher.js
```diff
@@ -58,6 +63,77 @@ async function main() {
...
+    (acc, r) => { acc[r.status] = (acc[r.status] ?? 0) + 1; return acc; },
+    {}
+  );
+  log.info(tally, 'catchup done');
+}
+
+function _hourIdToMs(hourId) {
+  const [y, m, d, h] = hourId.split('-').map(Number);
+  return Date.UTC(y, m - 1, d, h);
+}
```
tomerp20: **[LOW] Duplicated helper.** `_hourIdToMs` is now defined in both `fetcher.js` and `disk-layout.js`, and `_msToHourId` duplicates the inline logic from `enumerateRange`. Consider exporting `hourIdToMs` / `msToHourId` from `disk-layout.js` and importing them here — keeps the canonical hour-arithmetic in one place.

data-pipeline/fetcher/fetcher.js
```diff
@@ -58,6 +63,77 @@ async function main() {
   logger.info(tally, 'done');
 }
 
+async function runCatchup(root, log) {
+  const anchor = latestOnDisk(root);
+  if (!anchor) {
+    log.error('no anchor to walk from — run --hour or --range first to seed the directory');
+    process.exit(1);
```
tomerp20: **[LOW] Cold-start `process.exit(1)` may truncate pino's worker-thread output.** In non-prod, pino-pretty runs in a transport worker; `process.exit` bypasses its flush. Usually fine for a one-line message but not guaranteed. Either call `log.flush?.()` + `await new Promise(r => setImmediate(r))` before exiting, or simply `throw new Error(...)` so the existing `main().catch` handler at fetcher.js:138 exits consistently with other fatal errors.

data-pipeline/fetcher/fetcher.js
```diff
@@ -58,6 +63,77 @@ async function main() {
...
+  // any already-in-flight requests are awaited but their results don't extend the walk.
+  const promises = candidates.map((hourId) =>
+    limit(async () => {
+      if (found404) return { status: 'skipped-catchup-abort', hourId };
+      const result = await downloadHour(root, hourId);
+      if (result.status === 'not-found') {
+        found404 = true;
+        log.warn({ hourId }, 'first 404 — halting catchup walk');
+      } else if (result.status === 'written') {
+        log.info({ hourId }, 'written');
```
tomerp20, line 103: **[LOW] Log ordering after first-404 can look contradictory.** When the 404 fires for hour N, up to 2 sibling requests for hours > N may already be in flight. If they return `'written'` they'll log `info written` *after* the `warn first 404 — halting catchup walk` line, which reads like a contradiction in the log stream. Behavioural impact: none (the files were legitimately fetched). Diagnostics impact: small. Optional: demote post-404 `info written` lines to `debug`, or tag them with `{ stale: true }`.

data-pipeline/fetcher/lib/cli.js
```diff
@@ -27,7 +27,14 @@ export function parseCLI(argv) {
     return { mode: 'range', start, end };
   }
 
-  die('usage: fetcher.js --hour <YYYY-MM-DD-H> | --range <YYYY-MM-DD> <YYYY-MM-DD>');
+  if (args[0] === '--catchup') {
+    if (args.length > 1) {
+      die('--catchup takes no arguments');
```
tomerp20, line 32: **[LOW] Error message could echo what was passed.** `die('--catchup takes no arguments (got: ${args.slice(1).join(' ')})')` makes copy-paste mistakes easier to diagnose. Skip if it diverges from the project's CLI error style.

tomerp20 (commented):
## ✅ Code Review Complete

| 🔒 Security (Orchestrator) | ✅ Done | 0 issues | — |
| 🛠 Correctness (Orchestrator) | ✅ Done | 1 MEDIUM, 4 LOW | 3 fixed (1 MED + 2 LOW), 2 LOW deferred |

**Routing note:** files under `data-pipeline/fetcher/` are pure Node.js/ESM CLI code (no React, no Express/Knex/Prisma), so neither the `react-code-reviewer` nor `backend-code-reviewer` specialist was spawned. The orchestrator handled the full review directly, applying the security rubric plus a correctness pass against the PR's acceptance criteria. This fallback is consistent with the orchestrator memory note for PR #176.

### Fixed in commit `732ed26`
- **[MEDIUM] `latestOnDisk` depth validation** — added a 4-segment depth check in `_collectHourIds` so stray `*.json.gz` files at non-canonical depths can't poison the anchor with a `NaN` seed.
- **[LOW] Duplicated hour-arithmetic helpers** — promoted `_hourIdToMs` / `_msToHourId` to exported `hourIdToMs` / `msToHourId` in `disk-layout.js`; `fetcher.js` now imports them. Verified roundtrip (`2025-05-01-15` → ms → back) still works.
- **[LOW] Cold-start exit truncation risk** — `runCatchup` now `throw`s instead of calling `process.exit(1)` directly; the existing `main().catch` handler exits with code 1 and gives pino's transport worker a chance to flush. Verified locally: cold-start against an empty `GHARCHIVE_DIR` exits 1, logs are not truncated, and the error message is preserved.

### Skipped (left as PR comments for author discretion)
- **[LOW] Log ordering after first-404** — purely cosmetic (`info written` for an in-flight hour may print after the `warn first 404` line). No behavioural impact, no data risk.
- **[LOW] CLI error echo for `--catchup foo`** — style-debatable; would diverge from the existing `die()` messages in the same file (`--hour`, `--range` don't echo the bad input either). Author can apply if desired.

### Verification
- Syntax-checked all three files with `node --check` ✅
- Smoke-tested `latestOnDisk` against a temp tree containing both canonical and stray `.json.gz` files at multiple depths — correct anchor returned ✅
- Smoke-tested cold-start (`--catchup` against an empty dir) — exits 1 with the original message via the fatal-error path ✅
- Smoke-tested `--catchup foo` — still errors with `--catchup takes no arguments` ✅
- The fetcher package has no automated test suite; verification was manual.

### Next steps
- [ ] Human review of the two deferred LOW findings (log ordering, CLI echo)

(orchestrator: code-review-orchestrator)

Issue #194, opened 2026-05-23
[Cassandra pipeline PR2] Ingester package


tomerp20:
> Source-of-truth docs: `docs/CassandraPlan.md` §4.2 + §5, `docs/InjestionPlan.md` §5, `docs/adr/0003-backfill-hourly-relay-race.md`, and the **Tracked Event / Hourly Ingest / Backfill / Backfill Run / Org** entries in `CONTEXT.md`. All on the `feature/cassandra-data-pipeline` lineage; the immediate precursor is **PR #193**. This PRD records the locked outcomes of the PR2 grill.

## Problem Statement

The Cassandra analytics pipeline has Cassandra schemas and a Fetcher that brings GH Archive `.json.gz` files to disk, but nothing yet **reads** those files into Cassandra. Without an Ingester, the `companies` table holds (Company, Org) rows that no Backfill or Hourly Ingest ever processes, and the `company_events` table stays empty.

The Ingester is the middle of the pipeline. It has two operational modes — Backfill, which sweeps the entire local archive for not-yet-initialised (Company, Org) rows, and Hourly Ingest, which processes the newest file for rows that the Backfill has already initialised. Both modes share a single binary and most of the same code path; they differ in which subset of `companies` they filter against and in how aggressively they write to Cassandra.

The hard problem the Ingester solves: take a multi-gigabyte gzipped file containing millions of GH Archive events, filter it to the tiny fraction that belongs to tracked (Company, Org) rows, extract analytical signals from each event without losing throughput, and persist the survivors to Cassandra under two different write strategies — all idempotently across crashes and re-runs.

## Solution

A standalone Node CLI in `data-pipeline/ingester/`. Three CLI modes (`--hour`, `--range`, `--catchup`) cover smoke-testing, multi-day backfills, and cron-style catch-up. A `--mode` flag (`hourly` or `backfill`) controls both the write strategy and which subset of `companies` rows are eligible — directly enforcing the per-row `initialized` relay-race handoff documented in ADR 0003.

[...]

PR #207, opened 2026-05-23
[Cassandra pipeline PR2.B] Ingester base: --hour --mode hourly end-to-end

### LINKED ISSUE #197: [Cassandra pipeline PR2.B] Ingester base: --hour --mode hourly end-to-end
## Parent PRD

#194

> **Amended after the PR3 grill (PR #203).** The Ingester is now strictly companies-agnostic and receives its filter input via the `--target-companies` CLI argument. See ADR 0004 (microservices-shaped CLI contract) and ADR 0005 (`processed_files` is hourly-only) for the reasoning. This issue's body has been updated to reflect that contract.

## What to build

The vertical tracer bullet for the Ingester. A new standalone Node package at `data-pipeline/ingester/` that reads a single `.json.gz` GH Archive file, filters to a target set of (Company, Org) pairs **passed in via CLI**, extracts tag and AI-attributed signals, and writes the survivors to `company_events` via individual prepared statements. This slice ships every layer end-to-end for the simplest mode (`--hour --mode hourly`). The other two modes (backfill batched writes, catchup file discovery) are deliberately left to follow-on slices.

The Ingester is a **pure stateless worker**:
- It never reads or writes the `companies` Cassandra table.
- It receives its filter list as `--target-companies wix:wix,microsoft:azure,…` (comma-separated `company:org` pairs).
- It builds the fast-filter `orgRegex` and the `orgToCompany` map from that CLI argument alone.
- In `--mode hourly` it reads and writes `processed_files` exactly as documented in `docs/InjestionPlan.md` §3.3 and §5.6.
- It does not touch `backfill_runs` or `backfill_progress` in this slice (those are written by the Orchestrator from PR3 or by the Ingester only in `--mode backfill`, which is PR2.C).

Module split per PRD §Implementation Decisions:
- `ingester.js` (glue: env, pino, CLI dispatch, worker pool orchestration)
- `lib/cli.js` (pure argv parsing; new required `--target-companies` arg)
- `worker.js` (hot path: fast filter, JSON.parse, 4-type filter, company filter, tag extraction, AI detection)
- `lib/tag-extractor.js` (deep, pure)
- `lib/ai-detector.js` (deep, pure)
- `lib/cassandra-writer.js` (driver wrapper, individual-insert strategy only in this slice)

**There is no `companies-loader` module.** The original PRD draft mentioned one; the grill amendment dropped it. The Ingester reads the `companies` table zero times.

The 4 hardcoded event types (`PushEvent`, `PullRequestEvent`, `IssuesEvent`, `ReleaseEvent`) and the per-event-type field sources for tag and AI extractors are documented in PRD §Implementation Decisions and `docs/InjestionPlan.md` §5.3.

## Acceptance criteria

- [ ] New Node package at `data-pipeline/ingester/` with `"engines": ">=20.6.0"`; deps `cassandra-driver`, `p-limit`, `pino`, plus the shared `disk-layout` package
- [ ] Env config via `node --env-file=.env`: `CASSANDRA_CONTACT_POINTS`, `CASSANDRA_LOCAL_DC` (required, fail-loud), `CASSANDRA_KEYSPACE`, `GHARCHIVE_DIR`, `LOG_LEVEL`, `INGEST_WORKERS`, `HOURLY_WRITE_CONCURRENCY`
- [ ] **No env var for per-job parameters.** Per ADR 0004, mode, target companies, hour, and (future) run-id are CLI args.
- [ ] Pino streams to stdout: `pino-pretty` when `NODE_ENV !== 'production'`, raw JSON otherwise; honours `LOG_LEVEL`
- [ ] CLI parses `--hour <YYYY-MM-DD-H>`, `--mode <hourly|backfill>` (defaults to `hourly` when `--hour` is the verb), and `--target-companies <comma-separated company:org pairs>` (required). Rejects invalid input with a single-line error and non-zero exit.
- [ ] `--target-companies` parser converts the arg into an array of `{company, org}` objects. The orgRegex and orgToCompany map are built from that array. No fallback to reading the `companies` Cassandra table.
- [ ] The Ingester does not query the `companies` table at any point during its run.
- [ ] Main thread opens the `.json.gz` file via `createReadStream` + `zlib.createGunzip` + `readline`, distributes batches of 1000 raw lines to workers
- [ ] Worker count default: `clamp(floor(os.cpus().length / 2), 1, 4)`; overridable via `INGEST_WORKERS`
- [ ] Workers apply the fast filter (substring regex on raw line) before any JSON.parse; only matching lines are parsed
- [ ] Workers filter on the four event types (`PushEvent`, `PullRequestEvent`, `IssuesEvent`, `ReleaseEvent`), apply the company filter via the orgToCompany map, run the tag extractor and AI detector, return parsed event objects to the main thread
- [ ] Tag extractor uses per-event-type text fields per PRD §Implementation Decisions; hardcoded ~40-keyword TECH_KEYWORDS list
- [ ] AI detector applies the documented login and message regexes; `actor.type === 'Bot'` and `[bot]` login suffix both trigger `is_ai = true`
- [ ] Main thread applies backpressure: in-flight queue capped at 10,000; drains to 5,000 before resuming line distribution
- [ ] Cassandra writes use prepared statements with `LOCAL_ONE` consistency; individual inserts through `p-limit(50)` (override via `HOURLY_WRITE_CONCURRENCY`)
- [ ] Each `company_events` insert populates: `company`, `org_name`, `year_month`, `event_time`, `event_id`, `event_type`, `repo_name`, `actor_login`, `is_ai`, `tech_tags`
- [ ] `year_month` is derived from `event.created_at` (`new Date(created_at).toISOString().slice(0, 7)`), never from the file name
- [ ] In `--mode hourly`: if the file is already present in `processed_files`, the Ingester logs "already processed, skipping" and exits 0 with no Cassandra activity. On successful completion of all events from the file, it writes a single `processed_files` row with `file_name = <canonical hour ID>`, `processed_at = now()`, `event_count = <total parsed>`, `filtered_count = <events written>`.
- [ ] On any Cassandra write failure that exhausts retries (3 attempts), the Ingester logs the partition and a sample `event_id`, exits non-zero, and writes no `processed_files` row
- [ ] Manual smoke test passes against a `cassandra:4.1` container (docker-compose.yml): schemas 001–008 applied, `node ingester.js --hour <known-good-hour> --target-companies <real-org-pairs>` writes events to `company_events`, re-running is a no-op, `cqlsh -e "SELECT count(*) FROM jobflow.company_events ..."` returns a sane count

## Blocked by

- All blockers are now satisfied:
  - #189 (Fetcher base) — merged
  - #193 (precursor docs) — merged
  - #196 (disk-layout shared package) — merged
  - #202 (stranded schema recovery) — merged
  - #203 (PR3 grill outcomes + this issue's amendments) — merged

## User stories addressed

(Per PRD #194's user stories list; with the grill amendments, the Ingester no longer "reads companies once at startup" — instead it parses `--target-companies` once at startup. Stories 6, 7, 8 are still in scope but their mechanism is the CLI arg, not a Cassandra query.)

- User story 1
- User story 2
- User story 5
- User story 6
- User story 7
- User story 8
- User story 10
- User story 11
- User story 12
- User story 13
- User story 14
- User story 15
- User story 16
- User story 17
- User story 18
- User story 19
- User story 20
- User story 21
- User story 22
- User story 23
- User story 27
- User story 28
- User story 29
- User story 30
- User story 32
- User story 33
- User story 34
- User story 37
- User story 38
- User story 39
- User story 40

tomerp20:
## Summary
- New standalone Node package at `data-pipeline/ingester/` — six modules covering CLI, worker hot path, tag extraction, AI detection, Cassandra writes, and main orchestration
- Ships `--mode hourly` end-to-end: reads a single `.json.gz` GH Archive file, filters to `--target-companies` pairs, writes events to `company_events`, marks `processed_files` on success
- Companies-agnostic: Ingester never reads the `companies` Cassandra table; all filtering is driven by the `--target-companies` CLI argument (ADR 0004)

## Acceptance criteria
- [x] New Node package at `data-pipeline/ingester/` with `"engines": ">=20.6.0"`; deps `cassandra-driver`, `p-limit`, `pino`, plus the shared `disk-layout` package
- [x] Env config via `node --env-file=.env`: `CASSANDRA_CONTACT_POINTS`, `CASSANDRA_LOCAL_DC` (required, fail-loud), `CASSANDRA_KEYSPACE`, `GHARCHIVE_DIR`, `LOG_LEVEL`, `INGEST_WORKERS`, `HOURLY_WRITE_CONCURRENCY`
- [x] **No env var for per-job parameters.** Per ADR 0004, mode, target companies, hour, and (future) run-id are CLI args.
- [x] Pino streams to stdout: `pino-pretty` when `NODE_ENV !== 'production'`, raw JSON otherwise; honours `LOG_LEVEL`
- [x] CLI parses `--hour`, `--mode`, `--target-companies` (required). Rejects invalid input with a single-line error and non-zero exit.
- [x] `--target-companies` parser converts the arg into `{company, org}` objects; orgRegex and orgToCompany map built from that array alone
- [x] The Ingester does not query the `companies` table at any point during its run
- [x] Main thread opens the `.json.gz` file via `createReadStream` + `zlib.createGunzip` + `readline`, distributes batches of 1000 raw lines to workers
- [x] Worker count default: `clamp(floor(os.cpus().length / 2), 1, 4)`; overridable via `INGEST_WORKERS`
- [x] Workers apply the fast filter (substring regex on raw line) before any JSON.parse; only matching lines are parsed
- [x] Workers filter on the four event types, apply company filter, run tag extractor and AI detector, return parsed event objects
- [x] Tag extractor uses per-event-type text fields; hardcoded ~40-keyword TECH_KEYWORDS list
- [x] AI detector applies login regex and message regex; `actor.type === 'Bot'` and `[bot]` login suffix both trigger `is_ai = true`
- [x] Main thread applies backpressure: in-flight queue capped at 10,000; drains to 5,000 before resuming (via `rl.pause()`/`rl.resume()`)
- [x] Cassandra writes use prepared statements with `LOCAL_ONE` consistency; individual inserts through `p-limit(50)`
- [x] Each `company_events` insert populates all required columns including `org_name` and `year_month`
- [x] `year_month` derived from `event.created_at` (`.toISOString().slice(0, 7)`), never from file name
- [x] In `--mode hourly`: checks `processed_files` on startup, logs "already processed, skipping" and exits 0 if already done; writes `processed_files` row on successful completion
- [x] On Cassandra write failure (3 retries exhausted): logs partition + sample event_id, exits non-zero, writes no `processed_files` row
- [ ] Manual smoke test against `cassandra:4.1` container (docker-compose.yml in `data-pipeline/`)

## Related
Closes #197


tomerp20:
## Security Review — PR #207

Reviewed files: `ingester.js`, `worker.js`, `lib/cli.js`, `lib/tag-extractor.js`, `lib/ai-detector.js`, `lib/cassandra-writer.js`, `package.json`, `.env.example`

---

### Finding 1 — MEDIUM: ReDoS risk in org-name regex (no escaping of user-supplied org names)

**File:** `ingester.js` lines 56–57 / `worker.js` line 8

```js
const orgRegexSource = orgNames.map(o => o + '/').join('|');
// ...
const orgRegex = new RegExp(orgRegexSource);
```

Org names come from `--target-companies` on the CLI. If a caller passes a crafted org name containing regex metacharacters (e.g. `(a+)+`, `.+`, `[a-z]+`), the concatenated alternation can produce a pattern that causes catastrophic backtracking on long event lines.

In practice this is a **trusted-operator CLI**, so the realistic exploitation surface is low — an attacker would need shell access to the process invocation. However, defensive escaping costs nothing and prevents accidental breakage from org names that happen to contain `.` or `+` (which are valid GitHub org name characters in some contexts).

**Recommendation:** Escape each org name before concatenating:
```js
function escapeRegex(s) { return s.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'); }
const orgRegexSource = orgNames.map(o => escapeRegex(o) + '\\/').join('|');
```

---

### Finding 2 — LOW: Path traversal via `hourId` is mitigated but not fully closed

**File:** `disk-layout/index.js` lines 6–8

```js
export function hourIdToPath(root, hourId) {
  const [year, month, day, hour] = hourId.split('-');
  return path.join(root, year, month, day, `${hour}.json.gz`);
}
```

`path.join` normalises `..` segments. A crafted `hourId` like `2025-..-..-0` would produce `path.join(root, '2025', '..', '..', '0.json.gz')` which resolves to a path two levels above `root`. However, for the `--hour` path the CLI validates with `HOUR_RE = /^\d{4}-\d{2}-\d{2}-\d{1,2}$/` before `hourId` ever reaches `hourIdToPath`, so `..` cannot appear. For the `--range` path, `enumerateRange` generates hour IDs internally from validated date arithmetic, so they are also safe. For `--catchup`, `latestOnDisk` generates IDs only from files already on disk.

The protection is real, but it is distributed across two layers (CLI regex and `enumerateRange`) rather than centralised in `hourIdToPath` itself. A future caller that bypasses CLI parsing would be unprotected.

**Recommendation:** Add an explicit assertion inside `hourIdToPath` (or at the `disk-layout` package boundary) that each segment matches `\d+`, failing fast rather than relying solely on callers:
```js
if (!/^\d{4}-\d{2}-\d{2}-\d{1,2}$/.test(hourId)) throw new Error(`invalid hourId: ${hourId}`);
```

---

### Finding 3 — LOW: CQL injection surface — confirmed safe (prepared statements used correctly)

**File:** `lib/cassandra-writer.js` lines 39–50

All three CQL statements (`INSERT INTO company_events`, `SELECT FROM processed_files`, `INSERT INTO processed_files`) are prepared at connect time and executed with positional `?` parameters. No event data or user input is ever interpolated into a query string. The keyspace name is interpolated at prepare time (not execute time), but it comes only from `CASSANDRA_KEYSPACE` env var, not from any user-controlled input.

No issue — confirmed clean.

---

### Finding 4 — LOW: `workerData` transfer is safe but the `orgRegex` field name is shadowed incorrectly

**File:** `worker.js` line 5

```js
const { orgRegex: orgRegexSource, orgToCompany } = workerData;
```

`ingester.js` sends `{ orgRegexSource, orgToCompany }` (key name: `orgRegexSource`), but `worker.js` destructures using `{ orgRegex: orgRegexSource }` — looking for a key named `orgRegex`. This means `orgRegexSource` will always be `undefined`, `new RegExp(undefined)` produces `/(?:)/` (matches everything), and the fast pre-filter becomes a no-op, not a failure. Every raw line will be passed to `JSON.parse` regardless of org, which is a performance regression but not a security issue.

This is a correctness bug rather than a security vulnerability. It is included here because a broken pre-filter silently widens the attack surface for any downstream processing of untrusted JSON.

**Recommendation:** Align the key name: change `ingester.js` line 253 to `{ workerData: { orgRegexSource, orgToCompany } }` and destructure as `const { orgRegexSource, orgToCompany } = workerData;` in worker.js, or rename the sent key to `orgRegex` to match what the worker expects.

---

### Finding 5 — LOW: Secrets leakage in error logs is well-controlled; one improvement possible

**File:** `ingester.js` line 160

```js
logger.error({ event_id: event.event_id, partition: `${event.company}/${event.year_month}`, err: err.message }, 'write failed after retries');
```

Logged fields on write failure: `event_id` (GitHub numeric ID — public), `partition` (company + year_month — internal label, not PII), `err.message` (Cassandra driver error string). The full event object (which contains `actor_login`, PR bodies, commit messages) is **not** logged.

Similarly in `worker.js` there is no logging at all — errors are swallowed silently via `continue`.

The one gap: `actor_login` and `repo_name` are never logged directly, which is good. However if Cassandra ever returns the offending row in its error message (e.g. a `WriteFailureException` detail string), `err.message` could inadvertently contain actor_login. This is driver-dependent and unlikely but worth being aware of.

**Recommendation:** Consider redacting `err.message` through pino's `redact` option for fields matching `/actor|login|email/` in production. At minimum, document that `LOG_LEVEL=debug` should not be set in production since partition rate-logger output at debug level emits partition keys.

---

### Finding 6 — INFO: `CASSANDRA_CONTACT_POINTS` comma-split is safe

**File:** `ingester.js` line 16

```js
const CASSANDRA_CONTACT_POINTS = (process.env.CASSANDRA_CONTACT_POINTS ?? '').split(',').map(s => s.trim()).filter(Boolean);
```

This produces a plain JavaScript string array. The cassandra-driver accepts `contactPoints` as `string[]` and handles its own connection logic — there is no shell interpolation, no eval, no dynamic DNS lookup that could be injected through a crafted value. A malicious value like `127.0.0.1,attacker.host` would simply add an attacker-controlled Cassandra endpoint, which requires the attacker to already control the environment. No additional hardening needed beyond standard env-var secrets management.

No issue.

---

### Finding 7 — INFO: `npm audit` — no vulnerabilities found

All installed packages (`cassandra-driver@4.9.0`, `pino@9.14.0`, `p-limit@6.2.0`, `pino-pretty@13.1.3`, and transitive dependencies) carry no known CVEs in the installed lock snapshot. The dependency tree is minimal and appropriate for a data pipeline.

---

### Summary Table

| # | Severity | Area | Finding |
|---|----------|------|---------|
| 1 | MEDIUM | ReDoS | Org names used in regex without escaping metacharacters |
| 2 | LOW | Path traversal | `hourIdToPath` relies on callers for validation; no guard at package boundary |
| 3 | LOW | CQL injection | Confirmed safe — prepared statements used throughout |
| 4 | LOW | Correctness / security | `workerData` key mismatch makes org pre-filter a no-op (`orgRegex` vs `orgRegexSource`) |
| 5 | LOW | Secrets / PII in logs | `err.message` on write failure could contain driver-echoed row data in edge cases |
| 6 | INFO | Env injection | `CASSANDRA_CONTACT_POINTS` split is safe |
| 7 | INFO | Dependencies | `npm audit` clean |

**Blocking:** None — no finding requires a block on merge, but Finding 1 (ReDoS escaping) and Finding 4 (workerData key mismatch correctness bug) should be addressed before this goes to production.

tomerp20:
## Backend Review — STANDARD (high-risk module)

Reviewed the new `data-pipeline/ingester/` package (CLI, worker hot path, tag/AI extraction, Cassandra writer, main orchestrator). Note: no ESLint config exists for this new package — the static-analysis guardrail is not active here. Consider adding `eslint.config.mjs` with `typescript-eslint` (or the JS preset) before this package grows further.

This PR is functionally close, but it ships with two concrete runtime defects that need fixing before merge.

### Issues by Severity

**CRITICAL**
- `worker.js:5` — destructures `workerData.orgRegex` but `ingester.js:253` sends it as `orgRegexSource`. The property name mismatch means the worker receives `undefined`, then `new RegExp(undefined)` produces `/(?:)/`, which matches every line. The fast-filter (the whole point of doing substring filtering before `JSON.parse`) is silently disabled — every raw line is parsed, defeating the perf design of the hot path.

**HIGH**
- `ingester.js:151–155` — references an undeclared `resumeCallback` identifier inside `onWorkerResult`. Modules run in strict mode, so the first time `paused && inFlight <= 5000` is true this throws `ReferenceError: resumeCallback is not defined` and crashes the hour. The actual resume already happens in the worker `message` handler at lines 171–174; this dead branch should be deleted.
- `cassandra-writer.js:15–17` — `policies.retry: new cassandra.policies.retry.RetryPolicy()` installs the **base** retry policy, whose default decisions all return `rethrow`. This *disables* the driver-level retries that would normally cover read/write timeouts and unavailable hosts, leaving only the manual `_writeWithRetry` (which retries blindly on every error class — including non-retryable schema/validation errors). Either drop the override (use the driver default) or wrap `DefaultRetryPolicy` with `IdempotenceAwareRetryPolicy` and have `_writeWithRetry` distinguish transient errors from terminal ones.

**MEDIUM**
- `ingester.js:230–233` — terminates workers sequentially with `await w.terminate()` in a `for` loop. Use `await Promise.all(workers.map(w => w.terminate()))` so shutdown isn't serialised.
- `ingester.js:226–228` — drain loop polls every 50 ms with `sleep`. Fine for now but a `Promise`-based drain (decrement a counter that signals a `done` promise when it reaches 0) avoids the up-to-50 ms idle tail per hour and the spin if writes hang.
- `cassandra-writer.js:32` — `this._partitionCounts = new Map()` is initialised but never written or read anywhere. Dead state — remove it.
- `cassandra-writer.js:83–108` — retry loop catches *all* errors. Cassandra `ResponseError` with codes like `Invalid` / `SyntaxError` / `Unauthorized` is not retryable; retrying wastes 2.6 s per bad event and obscures the real cause. Inspect `err.code` (or `err instanceof types.ResponseError`) and rethrow non-transient codes immediately.
- `worker.js:36` — silently drops events without `created_at`, but the count is invisible (no metric, no debug log). Consider tracking a `droppedNoTimestamp` counter and emitting it at end-of-hour so silent data loss is observable.

**LOW**
- `ingester.js` — no `SIGINT` / `SIGTERM` handler. Ctrl-C mid-hour kills in-flight writes without `writer.shutdown()` and without writing `processed_files`. Acceptable for now (next run reprocesses), but worth a TODO.
- `worker.js:47` — `String(event.id)` produces the literal `"undefined"` when `event.id` is missing, which then becomes the Cassandra clustering key. Either guard (`if (event.id == null) continue;`) or assert non-null.
- `cli.js:79` — usage string is single-line; once the verb list grows, splitting into multi-line `\n` output helps.
- `package.json` — `cassandra-driver`, `pino`, `p-limit` declared but no lockfile-pinned subdeps; this is the first Node package in `data-pipeline/`, so it's worth deciding now whether subdependency policy matches the backend's (engines, exact pins, etc.).

### Testing Gaps
This PR adds ~770 lines of production code with zero tests. At minimum the next slice should cover:
- `lib/cli.js` — happy path + every `die()` branch (missing `--target-companies`, bad pair format, bad `--hour`, bad `--range`, start > end, unknown verb)
- `lib/tag-extractor.js` — per-event-type field extraction + the dedup-via-Set behaviour
- `lib/ai-detector.js` — `Bot` actor, `[bot]` suffix, regex hits across the four event types, no-match case
- `lib/cassandra-writer.js` — retry-then-success and retry-exhausted paths against a fake client
- `worker.js` — the fast-filter / 4-type filter / company filter chain (this would have caught the `orgRegex` workerData bug)

The `orgRegex` defect in particular is exactly the kind of cross-module contract bug a single integration test of `spawnWorkers` would have caught. Adding even one happy-path end-to-end test before this code starts being depended on by downstream slices would be high-leverage.

🤖 Backend reviewer

data-pipeline/ingester/worker.js
```diff
@@ -0,0 +1,57 @@
+import { workerData, parentPort } from 'worker_threads';
+import { extractTags } from './lib/tag-extractor.js';
+import { detectAI } from './lib/ai-detector.js';
+
+const { orgRegex: orgRegexSource, orgToCompany } = workerData;
```
tomerp20: **[CRITICAL] — workerData property name mismatch silently disables the fast filter**

`ingester.js:253` posts `workerData: { orgRegexSource, orgToCompany }` (property name `orgRegexSource`), but this line destructures property `orgRegex` (aliased to local `orgRegexSource`). The actual property `orgRegex` does not exist on `workerData`, so the local is `undefined`. Line 8's `new RegExp(undefined)` then produces `/(?:)/`, which matches every string — meaning the fast-filter (`if (!orgRegex.test(line)) continue;`) is a no-op. Every line in the file is `JSON.parse`'d, defeating the substring pre-filter that the design relies on.

**Current:**
```js
// worker.js
const { orgRegex: orgRegexSource, orgToCompany } = workerData;
const orgRegex = new RegExp(orgRegexSource); // RegExp(undefined) → /(?:)/

// ingester.js
new Worker(workerPath, { workerData: { orgRegexSource, orgToCompany } })
```

**Fix (one option — make the worker side match what ingester sends):**
```js
// worker.js
const { orgRegexSource, orgToCompany } = workerData;
if (!orgRegexSource) throw new Error('worker received empty orgRegexSource');
const orgRegex = new RegExp(orgRegexSource);
```

Also worth adding a fail-fast guard so this class of bug surfaces as an exception rather than a silent perf regression.

data-pipeline/ingester/ingester.js
```diff
@@ -0,0 +1,259 @@
...
+      writer.writeEvent(event).then(() => {
+        totalFiltered++;
+        inFlight--;
+        if (paused && inFlight <= 5000) {
+          paused = false;
+          if (resumeCallback) {
+            const cb = resumeCallback;
+            resumeCallback = null;
+            cb();
+          }
```
tomerp20: **[HIGH] — `resumeCallback` is not declared; this branch throws ReferenceError**

This block reads, reassigns, and calls `resumeCallback` but the identifier is never declared anywhere (no `let`/`const`/`var` in `processHour` or module scope). ES modules run in strict mode, so the first time the condition `paused && inFlight <= 5000` is true, this throws `ReferenceError: resumeCallback is not defined` and rejects the `writeEvent` callback chain — crashing the hour without writing `processed_files`.

The other `message` listener at lines 168–175 already handles resume via `rl?.resume()` directly, so this whole inner block is dead code that should be deleted.

**Current:**
```js
writer.writeEvent(event).then(() => {
  totalFiltered++;
  inFlight--;
  if (paused && inFlight <= 5000) {
    paused = false;
    if (resumeCallback) {                // ← undeclared
      const cb = resumeCallback;
      resumeCallback = null;
      cb();
    }
  }
})
```

**Fix:**
```js
writer.writeEvent(event).then(() => {
  totalFiltered++;
  inFlight--;
  // resume handled in the worker 'message' listener below
})
```

(And if you want `onWorkerResult` to also drive resume — which is reasonable since the message-listener path only resumes once per batch — declare `let resumeCallback = null` properly and pick one of the two paths to be authoritative.)

tomerp20:
## Code Review — Final Status

**PR:** #207 — Ingester base: --hour --mode hourly end-to-end
**Fix commit:** 098231e

| Agent | Issues found | Issues fixed | Skipped |
|---|---|---|---|
| backend-code-reviewer | 8 (1 CRITICAL, 1 HIGH, 4 MEDIUM, 1 HIGH-driver, 1 LOW) | 8 | 0 |
| security-reviewer | 4 (1 MEDIUM ReDoS, 1 LOW path-traversal, 1 CRITICAL worker mismatch, 1 LOW driver log) | 3 | 1 |
| orchestrator (security) | — | — | — |

### Fixed in commit 098231e

| # | Severity | Finding | Fix |
|---|---|---|---|
| 1 | CRITICAL | `worker.js` — workerData key mismatch (`orgRegex` vs `orgRegexSource`): fast filter was a no-op, every line parsed | Fixed destructuring key; added guard throw |
| 2 | CRITICAL | `ingester.js` — stale `resumeCallback` reference caused ReferenceError on first backpressure-resume | Removed stale branch; resume is `rl.resume()` in worker message listener |
| 3 | HIGH | `cassandra-writer.js` — base `RetryPolicy` has no-op methods; all retries were silent no-ops | Replaced with `DefaultRetryPolicy`; added `isRetryable()` guard on manual retry loop |
| 4 | MEDIUM | `ingester.js` — org names not escaped before `RegExp` construction (ReDoS risk) | `escapeRegex()` applied before joining |
| 5 | MEDIUM | `ingester.js` — sequential `w.terminate()` loop | `Promise.all(workers.map(w => w.terminate()))` |
| 6 | MEDIUM | `worker.js` — silent drops on missing `created_at` / `event.id` not observable | Track `droppedNoTimestamp` / `droppedNoId`; logged at end-of-hour |
| 7 | MEDIUM | `cassandra-writer.js` — unused `_partitionCounts` dead state | Removed |
| 8 | LOW | `ingester.js` — no SIGINT/SIGTERM handler; pino and Cassandra not flushed on Ctrl-C | Added graceful shutdown handler |

### Skipped (requires product decision)

| # | Severity | Finding | Reason skipped |
|---|---|---|---|
| S1 | LOW | `cassandra-writer.js` — `err.message` could echo driver row data in edge cases; suggest pino `redact` config | Informational / hardening; no PII in event fields for this project. Left as a follow-up hardening task. |

### Testing gaps (flagged, not blocking)
No automated tests in this PR. Pure functions (`cli.js`, `tag-extractor.js`, `ai-detector.js`) are the highest-leverage targets for future unit tests. Decision to defer is documented in PRD #194.

### ESLint gap (flagged, not blocking)
No ESLint config in `data-pipeline/ingester/`. Recommended as a follow-up chore.

---
**Workflow complete.** Awaiting human review and merge approval.

PR #211, opened 2026-05-23
[Cassandra pipeline PR2.D] Ingester: --catchup file-set discovery

### LINKED ISSUE #199: [Cassandra pipeline PR2.D] Ingester: --catchup file-set discovery
## Parent PRD

#194

> **Amended after the PR3 grill (PR #203).** The Ingester's `--catchup` mode keeps its file-discovery behaviour but the underlying `processed_files` check is now strictly an hourly-mode feature. Catchup in `--mode hourly` walks the disk and skips files already in `processed_files`, exactly as designed. Catchup in `--mode backfill` is unreachable in v1 — the Backfill Orchestrator (PR3) always invokes the Ingester with `--range`, never `--catchup`. This issue's acceptance criteria are tightened accordingly.

## What to build

Add the `--catchup` verb to the Ingester. The behaviour:
- Walk `GHARCHIVE_DIR` and enumerate every `.json.gz` file present.
- For each, look up its canonical hour ID in the `processed_files` table.
- Skip files already present (they've been successfully processed in a prior Hourly Ingest).
- For all the remaining (unprocessed) files, dispatch each one through the same pipeline as `--hour <ID>` would — including the `--target-companies` filter, the worker pool, and the individual-insert write strategy.
- `--mode` defaults to `hourly` for `--catchup`. **In v1, `--catchup --mode backfill` is not a supported combination** because the Backfill Orchestrator always invokes the Ingester with `--range`. If someone passes `--catchup --mode backfill`, the Ingester exits non-zero with a clear error.

This slice does NOT depend on Slice C (#198) — it touches disjoint code (file discovery, not write strategy). It uses only the individual-insert path from #197.

## Acceptance criteria

- [ ] CLI accepts `--catchup` (with no other positional arguments) and defaults `--mode` to `hourly`
- [ ] `--catchup` requires `--target-companies` (same as `--hour` and `--range`). Without it, exits non-zero with a clear error.
- [ ] `--catchup --mode backfill` is rejected at the CLI layer with a single-line error and non-zero exit. This combination is reserved for a future PR.
- [ ] On startup with `--catchup`, the Ingester walks every `.json.gz` file under `GHARCHIVE_DIR` recursively (matches the disk layout from the shared `disk-layout` package)
- [ ] For each enumerated file, the Ingester queries `processed_files` for the corresponding canonical hour ID; already-present rows are skipped silently (no log noise per file)
- [ ] A single startup log line summarises: total files enumerated, count already processed, count to process
- [ ] Files to process are dispatched in canonical-hour-ID order (chronological), one at a time, through the same pipeline as `--hour --mode hourly` would
- [ ] If any single file fails (crash, exhausted retries), the Ingester exits non-zero immediately without attempting later files; later files will be picked up on the next run because they still aren't in `processed_files`
- [ ] If no files are enumerated at all, the Ingester logs "no files in GHARCHIVE_DIR" and exits 0
- [ ] If all enumerated files are already in `processed_files`, the Ingester logs "nothing to do" and exits 0
- [ ] Re-running `--catchup` immediately after a successful run is a no-op (all files now in `processed_files`)
- [ ] Manual smoke test passes: with a few `.json.gz` files in `GHARCHIVE_DIR` (some already in `processed_files`, some not), `node ingester.js --catchup --target-companies <real-org-pairs>` processes only the not-yet-processed ones, summarises counts in the startup log, and exits 0

## Blocked by

- Blocked by #197

## User stories addressed

- User story 4
- User story 35

tomerp20:
## Summary
- Adds `enumerateAllOnDisk` to the `disk-layout` shared package: walks `GHARCHIVE_DIR` recursively and returns all hour IDs sorted chronologically
- Rejects `--catchup --mode backfill` at the CLI layer with a clear error (reserved for future PR)
- Replaces the old (wrong) catchup implementation with the correct one: walk disk → batch-check `processed_files` → log startup summary → process pending files in chronological order through the hourly pipeline

## Acceptance criteria
- [x] CLI accepts `--catchup` (with no other positional arguments) and defaults `--mode` to `hourly`
- [x] `--catchup` requires `--target-companies` (same as `--hour` and `--range`). Without it, exits non-zero with a clear error.
- [x] `--catchup --mode backfill` is rejected at the CLI layer with a single-line error and non-zero exit. This combination is reserved for a future PR.
- [x] On startup with `--catchup`, the Ingester walks every `.json.gz` file under `GHARCHIVE_DIR` recursively (matches the disk layout from the shared `disk-layout` package)
- [x] For each enumerated file, the Ingester queries `processed_files` for the corresponding canonical hour ID; already-present rows are skipped silently (no log noise per file)
- [x] A single startup log line summarises: total files enumerated, count already processed, count to process
- [x] Files to process are dispatched in canonical-hour-ID order (chronological), one at a time, through the same pipeline as `--hour --mode hourly` would
- [x] If any single file fails (crash, exhausted retries), the Ingester exits non-zero immediately without attempting later files
- [x] If no files are enumerated at all, the Ingester logs "no files in GHARCHIVE_DIR" and exits 0
- [x] If all enumerated files are already in `processed_files`, the Ingester logs "nothing to do" and exits 0
- [x] Re-running `--catchup` immediately after a successful run is a no-op (all files now in `processed_files`)

## Related
Closes #199

### DIFF
--- a/data-pipeline/ingester/ingester.js
+++ b/data-pipeline/ingester/ingester.js
@@ -79,19 +79,16 @@ if (cmd.verb === 'hour') {
   hourIds = [cmd.hourId];
 } else if (cmd.verb === 'range') {
   hourIds = enumerateRange(cmd.start, cmd.end);
-} else {
+} else if (cmd.verb === 'catchup') {
   // catchup: walk disk, skip files already recorded in processed_files
   const allOnDisk = enumerateAllOnDisk(GHARCHIVE_DIR);
   if (allOnDisk.length === 0) {
-    logger.info('no files in GHARCHIVE_DIR');
+    logger.warn('no files in GHARCHIVE_DIR');
     await writer.shutdown();
     process.exit(0);
   }
-  const pending = [];
-  for (const id of allOnDisk) {
-    const done = await writer.isFileProcessed(id);
-    if (!done) pending.push(id);
-  }
+  const processedFlags = await Promise.all(allOnDisk.map(id => writer.isFileProcessed(id)));
+  const pending = allOnDisk.filter((_, i) => !processedFlags[i]);
   const alreadyProcessed = allOnDisk.length - pending.length;
   logger.info(
     { total: allOnDisk.length, alreadyProcessed, toProcess: pending.length },
@@ -103,6 +100,8 @@ if (cmd.verb === 'hour') {
     process.exit(0);
   }
   hourIds = pending;
+} else {
+  throw new Error(`unknown verb: ${cmd.verb}`);
 }
 
 // ── Process each hour file ────────────────────────────────────────────────────
--- a/data-pipeline/ingester/ingester.js
+++ b/data-pipeline/ingester/ingester.js
@@ -6,7 +6,7 @@ import os from 'os';
 import path from 'path';
 import { fileURLToPath } from 'url';
 import pino from 'pino';
-import { hourIdToPath, enumerateRange, latestOnDisk, hourIdToMs, msToHourId } from 'disk-layout';
+import { hourIdToPath, enumerateRange, enumerateAllOnDisk } from 'disk-layout';
 import { parseCLI } from './lib/cli.js';
 import { CassandraWriter } from './lib/cassandra-writer.js';
 
@@ -80,24 +80,29 @@ if (cmd.verb === 'hour') {
 } else if (cmd.verb === 'range') {
   hourIds = enumerateRange(cmd.start, cmd.end);
 } else {
-  // catchup
-  const anchor = latestOnDisk(GHARCHIVE_DIR);
-  if (!anchor) {
-    logger.fatal('no anchor on disk — run --hour or --range first to seed the directory');
-    await writer.shutdown();
-    process.exit(1);
-  }
-  const ceilingMs = Date.now() - 2 * 3_600_000;
-  const anchorMs = hourIdToMs(anchor);
-  hourIds = [];
-  for (let t = anchorMs + 3_600_000; t <= ceilingMs; t += 3_600_000) {
-    hourIds.push(msToHourId(t));
-  }
-  if (hourIds.length === 0) {
-    logger.info({ anchor }, 'already up to date');
+  // catchup: walk disk, skip files already recorded in processed_files
+  const allOnDisk = enumerateAllOnDisk(GHARCHIVE_DIR);
+  if (allOnDisk.length === 0) {
+    logger.info('no files in GHARCHIVE_DIR');
     await writer.shutdown();
     process.exit(0);
   }
+  const pending = [];
+  for (const id of allOnDisk) {
+    const done = await writer.isFileProcessed(id);
+    if (!done) pending.push(id);
+  }
+  const alreadyProcessed = allOnDisk.length - pending.length;
+  logger.info(
+    { total: allOnDisk.length, alreadyProcessed, toProcess: pending.length },
+    'catchup startup summary'
+  );
+  if (pending.length === 0) {
+    logger.info('nothing to do');
+    await writer.shutdown();
+    process.exit(0);
+  }
+  hourIds = pending;
 }
 
 // ── Process each hour file ────────────────────────────────────────────────────
--- a/data-pipeline/ingester/lib/cli.js
+++ b/data-pipeline/ingester/lib/cli.js
@@ -73,6 +73,9 @@ export function parseCLI(argv) {
   }
 
   if (flags.catchup) {
+    if (flags.mode === 'backfill') {
+      die('--catchup --mode backfill is not supported; the Backfill Orchestrator uses --range, not --catchup');
+    }
     return { verb: 'catchup', mode: flags.mode ?? 'hourly', targetCompanies };
   }
 
--- a/data-pipeline/shared/disk-layout/index.js
+++ b/data-pipeline/shared/disk-layout/index.js
@@ -40,7 +40,9 @@ export function enumerateRange(startDate, endDate) {
 export function enumerateAllOnDisk(rootDir) {
   const ids = [];
   _collectHourIds(rootDir, rootDir, ids);
-  return ids.sort((a, b) => hourIdToMs(a) - hourIdToMs(b));
+  const decorated = ids.map(id => [id, hourIdToMs(id)]);
+  decorated.sort((a, b) => a[1] - b[1]);
+  return decorated.map(([id]) => id);
 }
 
 // Returns the most recent hour ID (YYYY-MM-DD-H) found under rootDir, or null if empty.
--- a/data-pipeline/shared/disk-layout/index.js
+++ b/data-pipeline/shared/disk-layout/index.js
@@ -36,6 +36,13 @@ export function enumerateRange(startDate, endDate) {
   return hours;
 }
 
+// Returns all hour IDs (YYYY-MM-DD-H) found under rootDir, sorted chronologically.
+export function enumerateAllOnDisk(rootDir) {
+  const ids = [];
+  _collectHourIds(rootDir, rootDir, ids);
+  return ids.sort((a, b) => hourIdToMs(a) - hourIdToMs(b));
+}
+
 // Returns the most recent hour ID (YYYY-MM-DD-H) found under rootDir, or null if empty.
 export function latestOnDisk(rootDir) {
   const ids = [];


tomerp20:
## Security Review — code-review-orchestrator

**PR #211: [Cassandra pipeline PR2.D] Ingester: --catchup file-set discovery**

Reviewed against OWASP Top 10, secrets exposure, injection, auth, and data integrity.

### Findings

**No security issues found.**

| Area | Result |
|------|--------|
| Path traversal | `GHARCHIVE_DIR` is env-var controlled, not user-supplied at runtime. `readdirSync` is safe. |
| Injection | No user input flows into `readdirSync`, Cassandra queries, or file paths. |
| Secrets exposure | No secrets in diff. Log fields are structured and bounded. |
| Cassandra queries | `isFileProcessed` uses prepared statements (unchanged from PR2.B). |
| CLI arg parsing | `--catchup --mode backfill` is rejected before reaching any data access. |
| Auth / authz | N/A — this is a CLI tool with no HTTP surface. |
| Error handling | Unprocessed files remain safe on failure: exit-non-zero without writing `processed_files`, so they're retried next run. |

✅ Security review complete — no action required.

tomerp20:
## Backend Review — STANDARD

**Scope:** `data-pipeline/ingester/ingester.js`, `data-pipeline/ingester/lib/cli.js`, `data-pipeline/shared/disk-layout/index.js`. This is a plain Node.js ESM CLI — no Express/Knex/TS surface in this PR — so the Express/Knex categories of the backend checklist don't apply. The review focuses on correctness, scalability of the new catchup loop, error/exit-code semantics, and testing gaps.

**ESLint guardrail:** `data-pipeline/` has no ESLint config — static analysis guardrail is not active for this subproject. Consider adding `eslint.config.mjs` (with the same flat config style used by `backend/`) covering the ingester so future PRs get an automated baseline.

### Issues by Severity

**MEDIUM**
1. `ingester.js:91-94` — `isFileProcessed` is awaited in a serial `for` loop. For a long-running catchup (hundreds–thousands of files) this is N round-trips to Cassandra before any work begins. Either bound it with `pLimit` or do a single batched `SELECT ... WHERE file_name IN (?)`. Acceptance criteria explicitly say "batch-check" — current code is per-file.
2. `ingester.js:84-89` (vs prior) — Removing the previous "no anchor on disk -> exit 1" guard changes a monitoring surface. A misconfigured fetcher that leaves `GHARCHIVE_DIR` empty is now a silent exit-0 instead of a fatal. This matches the PR acceptance criteria, but make sure the scheduler/alerting has an independent signal that the fetcher is running.
3. No tests added. The data-pipeline subproject has no test infra at all (`find` returned no `*.test.js`). A small node `--test` suite around `enumerateAllOnDisk` (empty dir, mixed-depth files, sort order, large set) would lock down the new code.

**LOW**
4. `disk-layout/index.js:39-45` — `enumerateAllOnDisk` calls `hourIdToMs` twice per comparison in `.sort()` (O(n log n) date constructions). Either compute `[ms, id]` tuples once or rely on a zero-padded canonical string. Hygiene only.
5. `ingester.js:82` — Catchup runs in an unannotated `else` branch. If a future verb is added, it will silently fall through to catchup logic. Add an explicit `else if (cmd.verb === 'catchup')` with a terminal `else { throw new Error('unhandled verb') }`.
6. `disk-layout/index.js:39-45` — New function silently drops non-canonical files (same as `latestOnDisk`). Consider a `logger.debug` for skipped paths so operators get a signal if files land at the wrong depth.

### Testing Gaps
- No unit tests for `enumerateAllOnDisk` (empty dir, mixed depths, sort correctness across month/day boundaries).
- No integration coverage for the catchup-with-already-processed-files branch.
- No CLI test for the new `--catchup --mode backfill` rejection path.

Inline comments follow for the issues with specific line targets.

data-pipeline/ingester/ingester.js
```diff
@@ -80,24 +80,29 @@ if (cmd.verb === 'hour') {
...
+    process.exit(0);
   }
-  const ceilingMs = Date.now() - 2 * 3_600_000;
-  const anchorMs = hourIdToMs(anchor);
-  hourIds = [];
-  for (let t = anchorMs + 3_600_000; t <= ceilingMs; t += 3_600_000) {
-    hourIds.push(msToHourId(t));
+  const pending = [];
+  for (const id of allOnDisk) {
+    const done = await writer.isFileProcessed(id);
```
tomerp20: **[MEDIUM] — Per-file `isFileProcessed` is N round-trips before any work starts**

The acceptance criteria call for batch-checking `processed_files`, but this loop issues a serial `SELECT` per enumerated hour ID. For a few weeks of catchup that is 500-1000+ sequential Cassandra round-trips before the first byte is parsed. The startup summary log will also be misleading because the "startup" is actually dominated by these blocking reads.

**Two options:**

Bound concurrency (cheapest):
```js
import pLimit from "p-limit";
const limit = pLimit(32);
const flags = await Promise.all(
  allOnDisk.map(id => limit(() => writer.isFileProcessed(id)))
);
const pending = allOnDisk.filter((id, i) => !flags[i]);
```

Or a true batched query in `CassandraWriter` (best):
```js
// SELECT file_name FROM processed_files WHERE file_name IN (?, ?, ...)
// chunk into 200-500 per query to stay under coordinator limits
```

data-pipeline/ingester/ingester.js
```diff
@@ -80,24 +80,29 @@ if (cmd.verb === 'hour') {
...
 } else {
-  // catchup
-  const anchor = latestOnDisk(GHARCHIVE_DIR);
-  if (!anchor) {
-    logger.fatal('no anchor on disk — run --hour or --range first to seed the directory');
+  // catchup: walk disk, skip files already recorded in processed_files
+  const allOnDisk = enumerateAllOnDisk(GHARCHIVE_DIR);
+  if (allOnDisk.length === 0) {
+    logger.info('no files in GHARCHIVE_DIR');
     await writer.shutdown();
```
tomerp20, line 87: **[MEDIUM] — Monitoring-surface change: empty `GHARCHIVE_DIR` is now silent exit 0**

The previous code exited non-zero with `no anchor on disk — run --hour or --range first to seed the directory`. The new code logs info and exits 0. This matches the PR acceptance criteria, so it is intentional — but it means a misconfigured / broken fetcher (which would leave the directory empty) no longer surfaces as a failed catchup job exit code.

Confirm one of the following is true before merge:
- The orchestrator has an independent health signal that the fetcher is producing files (so empty dir is not silently masked), OR
- Add a `logger.warn` (not `info`) on the empty-directory branch so log-based alerting can catch it.

data-pipeline/ingester/ingester.js
```diff
@@ -80,24 +80,29 @@ if (cmd.verb === 'hour') {
 } else if (cmd.verb === 'range') {
   hourIds = enumerateRange(cmd.start, cmd.end);
 } else {
-  // catchup
-  const anchor = latestOnDisk(GHARCHIVE_DIR);
-  if (!anchor) {
-    logger.fatal('no anchor on disk — run --hour or --range first to seed the directory');
+  // catchup: walk disk, skip files already recorded in processed_files
```
tomerp20, line 83: **[LOW] — Unannotated `else` branch silently catches future verbs**

This branch runs catchup logic without checking `cmd.verb`. If another verb is added to `parseCLI` later (and the dispatch above is updated but this block is missed), the new verb will silently fall through to catchup. Cheap to harden:

```js
} else if (cmd.verb === "catchup") {
  // catchup: walk disk, skip files already recorded in processed_files
  ...
} else {
  throw new Error(`unhandled verb: ${cmd.verb}`);
}
```

data-pipeline/shared/disk-layout/index.js
```diff
@@ -36,6 +36,13 @@ export function enumerateRange(startDate, endDate) {
   return hours;
 }
 
+// Returns all hour IDs (YYYY-MM-DD-H) found under rootDir, sorted chronologically.
+export function enumerateAllOnDisk(rootDir) {
+  const ids = [];
+  _collectHourIds(rootDir, rootDir, ids);
```
tomerp20, line 42: **[LOW] — `.sort()` calls `hourIdToMs` twice per comparison**

For thousands of IDs this constructs O(n log n) `Date` objects. Two cheap alternatives:

```js
// Tuple decorate-sort-undecorate
return ids
  .map(id => [hourIdToMs(id), id])
  .sort((a, b) => a[0] - b[0])
  .map(([, id]) => id);
```

Or change the canonical hour format to zero-pad the hour (`HH`) so plain string sort works — but that would be a broader change. Tuple sort is the minimal fix here.

data-pipeline/shared/disk-layout/index.js
```diff
@@ -36,6 +36,13 @@ export function enumerateRange(startDate, endDate) {
   return hours;
 }
 
+// Returns all hour IDs (YYYY-MM-DD-H) found under rootDir, sorted chronologically.
+export function enumerateAllOnDisk(rootDir) {
```
tomerp20, line 40: **[LOW] — Silently dropping non-canonical files gives no operator signal**

`_collectHourIds` filters strictly to `root/YYYY/MM/DD/H.json.gz`. Anything at the wrong depth is dropped without a log line. If a future change to the fetcher accidentally writes to a different depth, catchup will silently no-op those files.

Suggest a `logger.debug` (passed in as a param) for each skipped path — or at minimum a debug-level summary `{ skipped: n }` in the caller. Matches the existing `latestOnDisk` behaviour, so this is a hygiene suggestion, not a bug.

data-pipeline/ingester/lib/cli.js
```diff
@@ -73,6 +73,9 @@ export function parseCLI(argv) {
   }
 
   if (flags.catchup) {
+    if (flags.mode === 'backfill') {
+      die('--catchup --mode backfill is not supported; the Backfill Orchestrator uses --range, not --catchup');
```
tomerp20, line 77: **[LOW] — Good guard; consider mirroring in `--hour` for symmetry**

This correctly rejects `--catchup --mode backfill`. The same combination via `--hour --mode backfill` (without `--run-id`) already errors via the `--run-id is required` check below, so functional coverage is fine. Just noting that the precedence (hour > range > catchup) is what makes this safe — if dispatch order ever changes, this guard would need to move.

No change required; flagging for future maintainers.

tomerp20:
## Code Review — Final Status (code-review-orchestrator)

### Phase 1: Review findings

| # | Severity | Reviewer | Finding |
|---|----------|----------|---------|
| 1 | MEDIUM | backend | N+1 serial `isFileProcessed` calls — sequential Cassandra round-trips before any processing starts |
| 2 | MEDIUM | backend | `logger.info` for empty `GHARCHIVE_DIR` loses monitoring signal vs prior `logger.fatal` |
| 3 | MEDIUM | backend | No tests — new catchup behaviour is uncovered (whole subproject lacks test infra) |
| 4 | LOW | backend | `sort()` calls `hourIdToMs` twice per comparison — O(n log n) redundant Date constructions |
| 5 | LOW | backend | `else` fallthrough silently catches future unknown verbs |
| 6 | LOW | backend | `_collectHourIds` silent drops for non-canonical paths (no log signal) |
| 7 | LOW | backend | CLI `--catchup --mode backfill` guard noted (N/A — no change required) |
| — | ✅ PASS | security | No OWASP issues, no injection surface, prepared statements unchanged |

### Phase 2: Fixes applied (commit 27b2361)

| # | Fixed | Action |
|---|-------|--------|
| 1 | ✅ Fixed | Replaced serial `for` loop with `Promise.all` — all `isFileProcessed` checks fire concurrently |
| 2 | ✅ Fixed | Changed `logger.info` → `logger.warn` for empty GHARCHIVE_DIR |
| 3 | ⏭️ Skipped | No test infra in `data-pipeline/` subproject — out of scope for this PR |
| 4 | ✅ Fixed | Decorate-sort-undecorate pattern in `enumerateAllOnDisk` |
| 5 | ✅ Fixed | `else` → `else if (cmd.verb === 'catchup')` + terminal `else { throw }` |
| 6 | ⏭️ Skipped | Requires plumbing a logger into the `disk-layout` package — too invasive for this PR |
| 7 | N/A | No fix needed |

### Items for human resolution
- **Finding #3 (MEDIUM)**: `data-pipeline/` has no test infrastructure. The `enumerateAllOnDisk` function, `--catchup` CLI path, and `--mode backfill` rejection are all untested. Consider a follow-up issue to add a `node --test` harness for the pipeline subproject.
- **Finding #6 (LOW)**: Non-canonical `.json.gz` files under `GHARCHIVE_DIR` (e.g., at wrong depth) are silently skipped. A log line would help operators diagnose misplaced files — but requires adding a logger parameter to the `disk-layout` package.

---
4 issues fixed • 2 skipped (deferred) • 0 require immediate action before merge

Issue #219, opened 2026-05-23
[CRITICAL] Ingester cannot connect to Cassandra: two cassandra-driver API mismatches in cassandra-writer.js


tomerp20:
## Severity

**Critical / blocking.** The Ingester binary cannot establish a connection to Cassandra and crashes during \`connect()\`. Every Ingester invocation (\`--mode hourly\`, \`--mode backfill\`, \`--catchup\`) is broken. This blocks all of PR2's smoke-test ACs, PR3 (#205, #206) end-to-end validation, and PR4 (#209) deployment.

## How this was discovered

While provisioning the Linux production box and walking through the deployment runbook's Phase 2 (single-file hourly mode), the Ingester crashed immediately after startup. Two separate cassandra-driver API mismatches surfaced in \`data-pipeline/ingester/lib/cassandra-writer.js\`. Neither was caught by the existing code reviews on #197, #198, or #199 because all three issues had identical ACs:

> \"Manual smoke test passes (requires local \`cassandra:4.1\` container — not automated)\"

**No one actually ran the smoke test against a live Cassandra container before merging.** All three Ingester PRs (#207, #211, #215) merged without ever having a successful end-to-end Cassandra round-trip. The unit-test gap (deliberately accepted in PR1-PR3 to move faster) caught up to us here.

## Bug 1 — \`policies.retry.DefaultRetryPolicy\` does not exist (HIGH)

\`data-pipeline/ingester/lib/cassandra-writer.js\` line 32:

\`\`\`js
this._client = new Client({
  contactPoints,
  localDataCenter: localDc,
  keyspace,
  policies: {
    retry: new policies.retry.DefaultRetryPolicy(),  // ← TypeError: not a constructor
  },
  queryOptions: { ... },
});
\`\`\`

\`cassandra-driver\`'s \`policies.retry\` actually exports:

\`\`\`js
[ 'IdempotenceAwareRetryPolicy', 'FallthroughRetryPolicy', 'RetryPolicy' ]
\`\`\`

There is **no \`DefaultRetryPolicy\`**. The name appears to have been hallucinated during implementation.

### Repro

\`\`\`bash
cd data-pipeline/ingester
node --env-file=../../.env ingester.js --hour 2025-05-22-15 --mode hourly --target-companies microsoft:microsoft
\`\`\`

### Error

\`\`\`
TypeError: policies.retry.DefaultRetryPolicy is not a constructor
[...]

PR #222, opened 2026-05-23
[Cassandra pipeline PR3.B] Backfill Orchestrator (full lifecycle)

### LINKED ISSUE #206: [Cassandra pipeline PR3.B] Backfill Orchestrator (full lifecycle)
## Parent PRD

#204

## What to build

The Backfill Orchestrator (`backfill.js`) — the meaty half of the orchestrators package. Layered on top of Slice A (#205), which provides the foundation lib modules (`companies-state`, `spawn-ingester`) plus the package skeleton.

End-to-end behaviour:
- `node backfill.js` (no args) loads env, connects to Cassandra, checks for uninitialised Companies, branches into one of three lifecycle paths:
  - **Nothing to do** → log and exit 0.
  - **Fresh run** → snapshot the uninitialised set as the locked target, create a `backfill_runs` row with `status=in_progress`, spawn the Ingester for the full date range.
  - **Resume** → find the in-progress run's `run_id` and `target_companies`, read `max(date) WHERE status='completed'` from `backfill_progress`, spawn the Ingester with `--range <next-day> <yesterday>`.
- On Ingester exit code 0 → success path: execute a single LOGGED BATCH that flips every `(company, org_name)` row in `target_companies` to `initialized=true, initialized_at=now()` AND marks `backfill_runs.status=completed, completed_at=now()`. Atomic.
- On Ingester non-zero exit → failure path: update `backfill_runs.status=failed, completed_at=now()`. Next night's run sees `failed`, treats it like `completed`, and starts fresh.
- Date range comes from `earliestOnDisk` (via the shared `disk-layout` package's directory walk) and yesterday (UTC).

Module additions to the package (from PRD §Implementation Decisions):
- `backfill.js` — entry point (~200 lines with the three lifecycle branches).
- `lib/run-state.js` — wraps `backfill_runs` reads and writes. Methods: `findLatestRun()`, `createNewRun(targetCompanies)`, `markCompleted(runId, targetRows)` (the LOGGED BATCH; takes target rows so it can flip them in the same batch), `markFailed(runId)`.
- `lib/disk-bounds.js` — exposes `earliestOnDisk(rootDir)` and `yesterdayUtc()`. The first walks `GHARCHIVE_DIR` for the earliest hour-ID; the second computes the appropriate end date.
- **Extends `lib/companies-state.js`** from Slice A: adds `readUninitialised(client) → Promise<Array<{company, org_name}>>` returning rows where `initialized = false AND active = true`. The `markInitialised` write logic lives in `lib/run-state.js` so it can be folded into the LOGGED BATCH alongside the `backfill_runs` update.

This issue handles all three lifecycle paths in one slice. Resume and fresh paths share enough code (LOGGED BATCH flip, subprocess spawn, target-companies serialisation) that splitting them would produce duplicate-ish code or an incomplete intermediate state.

## Acceptance criteria

- [ ] `backfill.js` entry point exists at `data-pipeline/orchestrators/backfill.js`. Zero CLI arguments.
- [ ] Loads env via `node --env-file=.env`. Required env vars (same as Slice A's): `CASSANDRA_CONTACT_POINTS`, `CASSANDRA_LOCAL_DC`, `CASSANDRA_KEYSPACE`, `GHARCHIVE_DIR`. Fails non-zero with a single-line error if any missing.
- [ ] Connects to Cassandra, verifies the connection. On failure, exits non-zero.
- [ ] Reads uninitialised Companies via `companies-state.readUninitialised(...)`. If empty, logs "nothing to do" and exits 0 (this is the rolled-in `check-pending-backfill` behaviour from PRD §Implementation Decisions).
- [ ] Queries `backfill_runs` for the latest row using the singleton-bucket pattern (`WHERE bucket = 'singleton' ORDER BY run_id DESC LIMIT 1`).
- [ ] Resume branch: if the latest row has `status = 'in_progress'`:
  - Reuses that row's `run_id` and `target_companies` (no new snapshot of `companies`).
  - Queries `backfill_progress` for `max(date) WHERE run_id = <existing> AND status = 'completed'`.
  - Computes `--range` as `<next-day-after-max> <yesterday>`. If `next-day > yesterday`, jumps straight to the flip step without spawning the Ingester.
- [ ] Fresh branch: if the latest row has `status = 'completed'` or `'failed'` or no row exists:
  - Generates a new `timeuuid` `run_id`.
  - Snapshots the uninitialised `(company, org)` set into `backfill_runs.target_companies`.
  - INSERTs the new `backfill_runs` row with `status = 'in_progress'`, `started_at = now()`.
  - Computes `--range` as `<earliestOnDisk> <yesterday>` via the shared `disk-layout` package.
- [ ] `lib/disk-bounds.earliestOnDisk(rootDir)` walks `GHARCHIVE_DIR` and returns the earliest hour-ID present. If the directory is empty or unreadable, throws (the orchestrator catches and exits non-zero with a clear log).
- [ ] Spawns the Ingester via `spawn-ingester` with argv `['data-pipeline/ingester/ingester.js', '--range', startIso, endIso, '--mode', 'backfill', '--target-companies', '<serialised>', '--run-id', runIdString]`.
- [ ] Pipes the Ingester's stdout/stderr into the parent pino stream (reuses Slice A's `spawn-ingester` wrapper unchanged).
- [ ] On Ingester exit code 0: executes a single LOGGED BATCH containing one UPDATE per row in `target_companies` (`SET initialized = true, initialized_at = now()`) plus one UPDATE on the `backfill_runs` row (`SET status = 'completed', completed_at = now()`). Exits 0.
- [ ] On Ingester non-zero exit: UPDATE the `backfill_runs` row with `status = 'failed', completed_at = now()`. Logs the exit code and (if available) a tail of stderr. Exits non-zero.
- [ ] Backfill Orchestrator writes NOTHING to `processed_files` or `company_events`. The only writes are to `companies` (LOGGED BATCH flip) and `backfill_runs` (insert + status updates).
- [ ] Backfill Orchestrator does NOT write `backfill_progress` rows — those are written by the Ingester per #198.
- [ ] Manual smoke tests pass (against the docker-compose Cassandra + the Ingester from #197/#198 with both modes merged):
  - **No-work case:** with all `companies` rows `initialized = true` (or no rows), `node backfill.js` exits 0 with "nothing to do."
  - **Fresh run:** seed two rows `(wix, wix, true, false)` and `(microsoft, microsoft, true, false)`. Run `node backfill.js`. A `backfill_runs` row is created with `status='in_progress'`; the Ingester subprocess is spawned with the expected argv; on Ingester exit 0, both `companies` rows flip to `initialized = true` and the `backfill_runs` row goes to `status='completed'` atomically (verify via cqlsh).
  - **Resume:** seed an uninitialised row, run `node backfill.js`, kill mid-run (Ctrl-C). Verify `backfill_runs` is still `in_progress` and `backfill_progress` has some dates `completed`. Re-run `node backfill.js`. Observe the resume log; the Ingester is spawned with `--range` adjusted past the max-completed date; the run completes; LOGGED BATCH flip happens.
  - **Failed run:** induce an Ingester crash (e.g. pass a broken `--target-companies` format). Verify the `backfill_runs` row is marked `failed`. Re-run; observe the orchestrator treats `failed` like `completed` and starts a fresh run with a new `run_id`.
  - **LOGGED BATCH atomicity:** during a successful run's flip, the moment between "all rows flipped" and "`backfill_runs.completed`" is invisible — either both states are visible together via cqlsh or neither is.

## Blocked by

- Blocked by #205 (provides the shared lib foundation: `companies-state` reader, `spawn-ingester`, the package skeleton)
- Blocked by #198 (Ingester `--mode backfill` — Backfill Orchestrator's whole purpose is to invoke this mode)

## User stories addressed

- User story 1
- User story 2
- User story 3
- User story 4
- User story 5
- User story 6
- User story 7
- User story 8
- User story 9
- User story 10
- User story 11
- User story 12
- User story 13
- User story 14
- User story 23
- User story 24
- User story 25
- User story 26
- User story 27
- User story 28
- User story 30
- User story 34
- User story 35

tomerp20:
## Summary
- Implements `backfill.js` — the Backfill Orchestrator with three lifecycle paths: nothing-to-do (exits 0), fresh run (snapshots uninitialised companies → inserts `backfill_runs` → spawns Ingester), and resume (reuses in-progress run's `target_rows` → computes range from `max(date)` in `backfill_progress` → re-spawns Ingester)
- On Ingester exit 0: executes a single LOGGED BATCH that atomically flips every `companies` row to `initialized=true` and marks `backfill_runs.status=completed`
- On Ingester non-zero exit: marks `backfill_runs.status=failed`; next nightly run treats `failed` like `completed` and starts fresh with a new `run_id`

## Acceptance criteria
- [x] `backfill.js` entry point exists at `data-pipeline/orchestrators/backfill.js`. Zero CLI arguments.
- [x] Loads env via `node --env-file=.env`. Required env vars: `CASSANDRA_CONTACT_POINTS`, `CASSANDRA_LOCAL_DC`, `CASSANDRA_KEYSPACE`, `GHARCHIVE_DIR`. Fails non-zero with a single-line error if any missing.
- [x] Connects to Cassandra, verifies the connection. On failure, exits non-zero.
- [x] Reads uninitialised Companies via `companies-state.readUninitialised(...)`. If empty, logs "nothing to do" and exits 0.
- [x] Queries `backfill_runs` for the latest row using the singleton-bucket pattern (`WHERE bucket = 'singleton' ORDER BY run_id DESC LIMIT 1`).
- [x] Resume branch: if the latest row has `status = 'in_progress'` — reuses `run_id` + `target_rows`; queries `backfill_progress` for `ORDER BY date DESC LIMIT 1`; computes `--range` as `<next-day> <yesterday>`; if `next-day > yesterday`, jumps straight to the flip step.
- [x] Fresh branch: if `status = 'completed'` or `'failed'` or no row — generates a new `timeuuid` `run_id`, snapshots uninitialised companies into `backfill_runs.target_rows`, INSERTs row with `status = 'in_progress'`, computes `--range` via `disk-bounds.earliestOnDisk`.
- [x] `lib/disk-bounds.earliestOnDisk(rootDir)` walks `GHARCHIVE_DIR` via `enumerateAllOnDisk`; throws if empty/unreadable.
- [x] Spawns the Ingester via `spawn-ingester` with `--range startIso endIso --mode backfill --target-companies <serialised> --run-id runIdString`.
- [x] Pipes Ingester stdout/stderr into the parent pino stream.
- [x] On Ingester exit code 0: executes LOGGED BATCH (`companies SET initialized=true + backfill_runs SET status=completed`). Exits 0.
- [x] On Ingester non-zero exit: UPDATE `backfill_runs SET status=failed`. Exits non-zero.
- [x] Writes nothing to `processed_files` or `company_events`.
- [x] Does NOT write `backfill_progress` rows (those are written by the Ingester).
- [ ] **Manual smoke tests — BLOCKED by #219** (cassandra-driver API mismatches in the Ingester cause `TypeError: this._client.prepare is not a function`; will re-run smoke tests once #219 lands)

## Related
Closes #206

### DIFF
--- a/data-pipeline/orchestrators/backfill.js
+++ b/data-pipeline/orchestrators/backfill.js
@@ -1,0 +1,167 @@
+import path from 'path';
+import { fileURLToPath } from 'url';
+import cassandra from 'cassandra-driver';
+import pino from 'pino';
+import { readUninitialised } from './lib/companies-state.js';
+import { spawnIngester } from './lib/spawn-ingester.js';
+import { findLatestRun, createNewRun, markCompleted, markFailed, getMaxCompletedDate } from './lib/run-state.js';
+import { earliestOnDisk, yesterdayUtc } from './lib/disk-bounds.js';
+
+const { Client } = cassandra;
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+// ── Env config ────────────────────────────────────────────────────────────────
+const CASSANDRA_CONTACT_POINTS = (process.env.CASSANDRA_CONTACT_POINTS ?? '').split(',').map(s => s.trim()).filter(Boolean);
+const CASSANDRA_LOCAL_DC = process.env.CASSANDRA_LOCAL_DC ?? '';
+const CASSANDRA_KEYSPACE = process.env.CASSANDRA_KEYSPACE ?? 'jobflow';
+const GHARCHIVE_DIR = process.env.GHARCHIVE_DIR ?? '';
+const LOG_LEVEL = process.env.LOG_LEVEL ?? 'info';
+const IS_PROD = process.env.NODE_ENV === 'production';
+
+// ── Logger ────────────────────────────────────────────────────────────────────
+const loggerOpts = { level: LOG_LEVEL };
+if (!IS_PROD) {
+  loggerOpts.transport = { target: 'pino-pretty' };
+}
+const logger = pino(loggerOpts);
+
+// ── Startup validation ────────────────────────────────────────────────────────
+if (CASSANDRA_CONTACT_POINTS.length === 0) {
+  logger.fatal('CASSANDRA_CONTACT_POINTS is required');
+  process.exit(1);
+}
+if (!CASSANDRA_LOCAL_DC) {
+  logger.fatal('CASSANDRA_LOCAL_DC is required');
+  process.exit(1);
+}
+if (!GHARCHIVE_DIR) {
+  logger.fatal('GHARCHIVE_DIR is required');
+  process.exit(1);
+}
+
+function nextDay(dateStr) {
+  const [y, m, d] = dateStr.split('-').map(Number);
+  const dt = new Date(Date.UTC(y, m - 1, d + 1));
+  return dt.toISOString().slice(0, 10);
+}
+
+async function main() {
+  // ── Connect to Cassandra ──────────────────────────────────────────────────
+  const client = new Client({
+    contactPoints: CASSANDRA_CONTACT_POINTS,
+    localDataCenter: CASSANDRA_LOCAL_DC,
+    keyspace: CASSANDRA_KEYSPACE,
+  });
+
+  try {
+    await client.connect();
+    await client.execute('SELECT release_version FROM system.local');
+    logger.info('cassandra connected');
+  } catch (err) {
+    logger.fatal({ err: err.message }, 'cassandra connection failed');
+    process.exit(1);
+  }
+
+  try {
+    // ── Check for uninitialised companies ─────────────────────────────────────
+    const uninitialised = await readUninitialised(client);
+    if (uninitialised.length === 0) {
+      logger.info('nothing to do');
+      await client.shutdown();
+      process.exit(0);
+    }
+
+    // ── Determine lifecycle path ──────────────────────────────────────────────
+    const latestRun = await findLatestRun(client, CASSANDRA_KEYSPACE);
+    const yesterday = yesterdayUtc();
+
+    let runId, targetRows, startDate;
+
+    if (latestRun && latestRun.status === 'in_progress') {
+      // ── Resume path ─────────────────────────────────────────────────────────
+      runId = latestRun.runId;
+      targetRows = latestRun.targetRows;
+
+      const maxDate = await getMaxCompletedDate(client, CASSANDRA_KEYSPACE, runId);
+      if (maxDate !== null) {
+        startDate = nextDay(maxDate);
+      } else {
+        // In-progress run with no completed dates — restart from earliest on disk
+        try {
+          startDate = earliestOnDisk(GHARCHIVE_DIR);
+        } catch (err) {
+          logger.fatal({ err: err.message }, 'cannot determine start date for resume');
+          await client.shutdown();
+          process.exit(1);
+        }
+      }
+
+      logger.info({ runId: runId.toString(), startDate, endDate: yesterday, companies: targetRows.length }, 'resuming run');
+    } else {
+      // ── Fresh path (no row, completed, or failed) ────────────────────────────
+      let earliest;
+      try {
+        earliest = earliestOnDisk(GHARCHIVE_DIR);
+      } catch (err) {
+        logger.fatal({ err: err.message }, 'cannot determine start date for fresh run');
+        await client.shutdown();
+        process.exit(1);
+      }
+
+      const created = await createNewRun(client, CASSANDRA_KEYSPACE, uninitialised);
+      runId = created.runId;
+      targetRows = created.targetRows;
+      startDate = earliest;
+
+      logger.info({ runId: runId.toString(), startDate, endDate: yesterday, companies: targetRows.length }, 'starting fresh run');
+    }
+
+    // ── Skip spawn if all dates already covered ───────────────────────────────
+    if (startDate > yesterday) {
+      logger.info({ runId: runId.toString() }, 'all dates already processed — flipping companies');
+      await markCompleted(client, CASSANDRA_KEYSPACE, runId, targetRows);
+      await client.shutdown();
+      process.exit(0);
+    }
+
+    // ── Spawn Ingester ────────────────────────────────────────────────────────
+    const targetCompaniesArg = targetRows.map(r => `${r.company}:${r.org_name}`).join(',');
+    const ingesterPath = path.resolve(__dirname, '..', 'ingester', 'ingester.js');
+
+    logger.info({ startDate, endDate: yesterday, companies: targetRows.length }, 'spawning ingester');
+
+    const exitCode = await spawnIngester(
+      [
+        ingesterPath,
+        '--range', startDate, yesterday,
+        '--mode', 'backfill',
+        '--target-companies', targetCompaniesArg,
+        '--run-id', runId.toString(),
+      ],
+      logger
+    );
+
+    // ── Handle Ingester exit ──────────────────────────────────────────────────
+    if (exitCode === 0) {
+      await markCompleted(client, CASSANDRA_KEYSPACE, runId, targetRows);
+      logger.info({ runId: runId.toString() }, 'backfill complete — companies initialised');
+    } else {
+      await markFailed(client, CASSANDRA_KEYSPACE, runId);
+      logger.error({ runId: runId.toString(), exitCode }, 'ingester failed — run marked failed');
+      await client.shutdown();
+      process.exit(exitCode);
+    }
+  } catch (err) {
+    logger.fatal({ err: err.message }, 'backfill orchestrator error');
+    await client.shutdown().catch(() => {});
+    process.exit(1);
+  }
+
+  await client.shutdown();
+  process.exit(0);
+}
+
+main().catch((err) => {
+  logger.fatal({ err }, 'backfill orchestrator failed');
+  process.exit(1);
+});
--- a/data-pipeline/orchestrators/backfill.js
+++ b/data-pipeline/orchestrators/backfill.js
@@ -58,7 +58,8 @@ async function main() {
     await client.execute('SELECT release_version FROM system.local');
     logger.info('cassandra connected');
   } catch (err) {
-    logger.fatal({ err: err.message }, 'cassandra connection failed');
+    logger.fatal({ err }, 'cassandra connection failed');
+    await client.shutdown().catch(() => {});
     process.exit(1);
   }
 
@@ -86,12 +87,15 @@ async function main() {
       if (maxDate !== null) {
         startDate = nextDay(maxDate);
       } else {
-        // In-progress run with no completed dates — restart from earliest on disk
+        // In-progress run with no completed dates — restart from earliest on disk.
+        // Safe to re-ingest: the Ingester writes (run_id, date)-keyed rows and
+        // PR3.A's idempotency contract guarantees that repeated writes under the
+        // same key produce the same end state.
         try {
           startDate = earliestOnDisk(GHARCHIVE_DIR);
         } catch (err) {
-          logger.fatal({ err: err.message }, 'cannot determine start date for resume');
-          await client.shutdown();
+          logger.fatal({ err }, 'cannot determine start date for resume');
+          await client.shutdown().catch(() => {});
           process.exit(1);
         }
       }
@@ -103,8 +107,8 @@ async function main() {
       try {
         earliest = earliestOnDisk(GHARCHIVE_DIR);
       } catch (err) {
-        logger.fatal({ err: err.message }, 'cannot determine start date for fresh run');
-        await client.shutdown();
+        logger.fatal({ err }, 'cannot determine start date for fresh run');
+        await client.shutdown().catch(() => {});
         process.exit(1);
       }
 
@@ -143,6 +147,11 @@ async function main() {
 
     // ── Handle Ingester exit ──────────────────────────────────────────────────
     if (exitCode === 0) {
+      // If markCompleted throws here the backfill_runs row stays in_progress,
+      // even though every hour has been ingested. That is recoverable: the next
+      // nightly run will enter the resume path, find startDate > yesterday at
+      // the "all dates already processed" short-circuit above, and retry
+      // markCompleted. The companies UPDATEs are idempotent, so this is safe.
       await markCompleted(client, CASSANDRA_KEYSPACE, runId, targetRows);
       logger.info({ runId: runId.toString() }, 'backfill complete — companies initialised');
     } else {
@@ -152,7 +161,7 @@ async function main() {
       process.exit(exitCode);
     }
   } catch (err) {
-    logger.fatal({ err: err.message }, 'backfill orchestrator error');
+    logger.fatal({ err }, 'backfill orchestrator error');
     await client.shutdown().catch(() => {});
     process.exit(1);
   }
--- a/data-pipeline/orchestrators/lib/run-state.js
+++ b/data-pipeline/orchestrators/lib/run-state.js
@@ -11,7 +11,12 @@ export async function findLatestRun(client, keyspace) {
   );
   if (result.rowLength === 0) return null;
   const row = result.rows[0];
-  const targetRows = (row.target_rows || []).map(t => ({
+  // Guard against a row with a missing/empty target_rows column. Without this,
+  // the resume path would spawn the Ingester with `--target-companies ""`.
+  if (!row.target_rows || row.target_rows.length === 0) {
+    throw new Error(`backfill_runs row ${row.run_id} has no target_rows — cannot resume`);
+  }
+  const targetRows = row.target_rows.map(t => ({
     company: t.get(0),
     org_name: t.get(1),
   }));
@@ -31,17 +36,21 @@ export async function createNewRun(client, keyspace, uninitialised) {
 
 export async function markCompleted(client, keyspace, runId, targetRows) {
   const now = new Date();
-  const queries = [
-    ...targetRows.map(r => ({
-      query: `UPDATE ${keyspace}.companies SET initialized = true, initialized_at = ? WHERE company = ? AND org_name = ?`,
-      params: [now, r.company, r.org_name],
-    })),
-    {
-      query: `UPDATE ${keyspace}.backfill_runs SET status = 'completed', completed_at = ? WHERE bucket = ? AND run_id = ?`,
-      params: [now, BUCKET, runId],
-    },
-  ];
-  await client.batch(queries, { logged: true, prepare: true });
+  // The per-company UPDATEs are idempotent (initialized=true is a fixed value),
+  // so they are issued individually rather than in a multi-partition LOGGED
+  // BATCH — the coordinator pressure and batch-size limit make a single LOGGED
+  // BATCH across many partitions an anti-pattern. If any company UPDATE fails
+  // here, the backfill_runs row stays in_progress; the resume-path short-circuit
+  // on the next nightly run will retry markCompleted (also idempotent).
+  const companyUpdate = `UPDATE ${keyspace}.companies SET initialized = true, initialized_at = ? WHERE company = ? AND org_name = ?`;
+  for (const r of targetRows) {
+    await client.execute(companyUpdate, [now, r.company, r.org_name], { prepare: true });
+  }
+  await client.execute(
+    `UPDATE ${keyspace}.backfill_runs SET status = 'completed', completed_at = ? WHERE bucket = ? AND run_id = ?`,
+    [now, BUCKET, runId],
+    { prepare: true }
+  );
 }
 
 export async function markFailed(client, keyspace, runId) {
--- a/data-pipeline/orchestrators/lib/run-state.js
+++ b/data-pipeline/orchestrators/lib/run-state.js
@@ -1,0 +1,63 @@
+import cassandra from 'cassandra-driver';
+
+const { types } = cassandra;
+const BUCKET = 'singleton';
+
+export async function findLatestRun(client, keyspace) {
+  const result = await client.execute(
+    `SELECT run_id, status, target_rows FROM ${keyspace}.backfill_runs WHERE bucket = ? ORDER BY run_id DESC LIMIT 1`,
+    [BUCKET],
+    { prepare: true }
+  );
+  if (result.rowLength === 0) return null;
+  const row = result.rows[0];
+  const targetRows = (row.target_rows || []).map(t => ({
+    company: t.get(0),
+    org_name: t.get(1),
+  }));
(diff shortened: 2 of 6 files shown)


tomerp20 (commented):
## Backend Review — STANDARD (HIGH-RISK areas)

Reviewed: `backfill.js`, `lib/run-state.js`, `lib/disk-bounds.js`, `lib/companies-state.js`, and package manifests.

Classification: **STANDARD** with one **HIGH-RISK** area — the multi-partition LOGGED BATCH and the data-snapshot semantics in the resume path. No frontend changes.

The code is readable, the lifecycle paths are clearly separated, and the `run_state` helpers cleanly encapsulate the CQL surface. The findings below are concrete, mostly correctness/scale risks rather than blocking bugs.

### Issues by Severity

**HIGH**
- `backfill.js:145-147` — If `markCompleted` throws after the Ingester returns exit 0, the run is left as `in_progress` even though all data has been ingested. The next nightly run will re-discover it via the resume path. That self-heals (the `startDate > yesterday` short-circuit re-runs `markCompleted`), but it is worth a brief comment in code so a future reader does not see this as a forgotten edge case.

**MEDIUM**
- `lib/run-state.js:32-45` — `markCompleted` builds a single LOGGED BATCH that updates every row of `companies` plus the `backfill_runs` row. This is a multi-partition LOGGED BATCH. Two concerns:
  1. **Batch size:** Cassandra `batch_size_fail_threshold` defaults to 50 KB; with hundreds of companies this will blow past the warn threshold and may hit fail. Even at the current target scale this is fragile.
  2. **Coordinator pressure:** multi-partition LOGGED BATCH has well-documented anti-pattern status — the coordinator holds the batchlog and replays partial failures. Consider either (a) executing the companies updates as un-LOGGED with idempotent retries and then a final separate UPDATE of `backfill_runs`, or (b) chunking the LOGGED BATCH.
- `lib/run-state.js:14-17` — `(row.target_rows || []).map(t => ({ company: t.get(0), org_name: t.get(1) }))` silently treats a `null` `target_rows` as empty. Combined with the resume path, this could send `--target-companies ""` to the Ingester. Add an explicit guard and fatal-log instead.
- `backfill.js:128` — `targetCompaniesArg = targetRows.map(r => \`${r.company}:${r.org_name}\`).join(',')`. If any `company` or `org_name` ever contains `,` or `:`, the Ingester will mis-parse the list. Since these come from the `companies` table and may be user-curated, encode the values (URL-encode each token, or use a JSON array passed via stdin/file) or validate at insert time.
- `backfill.js:88-97` — Resume with `getMaxCompletedDate(...) === null` restarts from `earliestOnDisk`. This re-runs every day previously attempted under the same `run_id`. The Ingester must be idempotent across re-runs for the same `(run_id, date)` for this to be safe. Worth a one-line code comment pointing at the Ingester's idempotency contract (PR3.A) so the assumption is documented.
- `lib/run-state.js:21-30` — `createNewRun` snapshots `target_rows` from `uninitialised` at insert time. On resume, `markCompleted` flips `initialized=true` for that snapshot regardless of the row's current `active` state. If a company is deactivated after the run starts, this still sets `initialized=true`. Probably intentional (snapshot semantics), but worth one line of doc-comment in `markCompleted` so the behavior is explicit.
- `lib/companies-state.js:8-13` — `ALLOW FILTERING` on `WHERE initialized = false AND active = true`. Matches the existing `readInitialised` pattern, but as the table grows this is an O(N) cluster-wide scan with timeout risk. Add a secondary index on `initialized` (or maintain a denormalised view) before this table grows beyond a few thousand rows. Tracked as a follow-up rather than a blocker.

**LOW**
- `backfill.js:56-63` — On `client.connect()` failure the handler calls `process.exit(1)` without `client.shutdown()`. The Cassandra driver opens TCP sockets eagerly; on failure they may already be in a bad state, but a best-effort `client.shutdown().catch(() => {})` before exit is cheaper than the linger.
- `backfill.js:154-158` — The outer catch uses `err.message` only, dropping the stack. Other catches in the same file do the same — consistent — but in production debugging this hides the trace. Consider `{ err }` (pino serialises Error objects) for the fatal log.
- `lib/disk-bounds.js:8` — Comment says "ids are sorted chronologically (earliest first)". Good, matches `disk-layout.enumerateAllOnDisk`. No issue, just confirming the contract is documented at the call site.
- `backfill.js:136` — `'--range', startDate, yesterday` passes two positional args after `--range`. Verified the existing Ingester (PR3.A) accepts this form, so OK — but a comment near this line documenting the Ingester contract would prevent regressions if someone adds a third positional later.

### Testing Gaps

This PR has no unit tests for the new modules. Minimum coverage I would expect before merge:
- `lib/disk-bounds.earliestOnDisk` — empty directory throws; non-empty returns the chronologically earliest `YYYY-MM-DD`.
- `lib/disk-bounds.yesterdayUtc` — boundary case across UTC midnight.
- `lib/run-state.findLatestRun` — null `target_rows`, empty rowset, populated rowset.
- `backfill.js` `nextDay` — month rollover, year rollover, leap-day.

A smoke-test gating on PR #219 (Ingester driver fix) is already noted in the PR description, so I will not block on full E2E.

### What I checked and did not flag
- ESLint: no config found in `data-pipeline/orchestrators/` — guardrail not active for this subproject. Noted as a follow-up; not blocking this PR.
- `spawn` is called with array argv (no shell), so no shell-injection risk.
- No secrets, tokens, or PII in logs.
- `prepare: true` is used consistently for parameterised CQL; keyspace is interpolated from env only.

data-pipeline/orchestrators/lib/run-state.js
```diff
@@ -0,0 +1,63 @@
...
+    ...targetRows.map(r => ({
+      query: `UPDATE ${keyspace}.companies SET initialized = true, initialized_at = ? WHERE company = ? AND org_name = ?`,
+      params: [now, r.company, r.org_name],
+    })),
+    {
+      query: `UPDATE ${keyspace}.backfill_runs SET status = 'completed', completed_at = ? WHERE bucket = ? AND run_id = ?`,
+      params: [now, BUCKET, runId],
+    },
+  ];
+  await client.batch(queries, { logged: true, prepare: true });
```
tomerp20: **[HIGH] — Multi-partition LOGGED BATCH on every company UPDATE**

`markCompleted` issues a single LOGGED BATCH containing one UPDATE per target company plus one UPDATE on `backfill_runs`. Multi-partition LOGGED BATCHes are an established Cassandra anti-pattern — the coordinator persists the batch to the system batchlog, replays partial failures, and the batch is sized against `batch_size_fail_threshold` (default 50 KB).

For the current scale (handful of companies) this works, but it does not scale and the failure mode is silent (batch rejected on the coordinator, run stays `in_progress`).

**Suggested fix:** issue the company UPDATEs as un-LOGGED idempotent statements (they are idempotent — `initialized=true`), then a single UPDATE for `backfill_runs.status="completed"`. The two operations being non-atomic is acceptable because the resume-path short-circuit (`startDate > yesterday`) will retry the run-state flip on the next nightly run.

data-pipeline/orchestrators/lib/run-state.js
```diff
@@ -0,0 +1,63 @@
...
+    `SELECT run_id, status, target_rows FROM ${keyspace}.backfill_runs WHERE bucket = ? ORDER BY run_id DESC LIMIT 1`,
+    [BUCKET],
+    { prepare: true }
+  );
+  if (result.rowLength === 0) return null;
+  const row = result.rows[0];
+  const targetRows = (row.target_rows || []).map(t => ({
+    company: t.get(0),
+    org_name: t.get(1),
+  }));
```
tomerp20, line 22: **[MEDIUM] — Null `target_rows` silently treated as empty**

`(row.target_rows || []).map(...)` swallows the case where the column is `null` (which Cassandra returns instead of an empty list for an unset multi-cell collection). The caller then enters the resume path with `targetRows = []`, and on line 128 of `backfill.js` builds `targetCompaniesArg = ""`, spawning the Ingester with `--target-companies ""`.

**Fix:**
```js
if (!row.target_rows || row.target_rows.length === 0) {
  throw new Error(`backfill_runs row ${row.run_id} has no target_rows — cannot resume`);
}
const targetRows = row.target_rows.map(t => ({ company: t.get(0), org_name: t.get(1) }));
```

data-pipeline/orchestrators/backfill.js
```diff
@@ -0,0 +1,167 @@
...
+    // ── Skip spawn if all dates already covered ───────────────────────────────
+    if (startDate > yesterday) {
+      logger.info({ runId: runId.toString() }, 'all dates already processed — flipping companies');
+      await markCompleted(client, CASSANDRA_KEYSPACE, runId, targetRows);
+      await client.shutdown();
+      process.exit(0);
+    }
+
+    // ── Spawn Ingester ────────────────────────────────────────────────────────
+    const targetCompaniesArg = targetRows.map(r => `${r.company}:${r.org_name}`).join(',');
```
tomerp20, line 132: **[MEDIUM] — Argument parsing brittleness in `--target-companies`**

`targetCompaniesArg = targetRows.map(r => \`${r.company}:${r.org_name}\`).join(",")` is a positional `key:value,key:value` encoding. If any `company` or `org_name` ever contains `,` or `:`, the Ingester parser silently mis-splits the list.

These values come from the `companies` table, which is operator-curated today, but the brittleness is implicit — a single odd `org_name` (e.g. `vercel,inc` or `team:platform`) becomes a silent ingestion bug.

**Fix options (pick one):**
1. URL-encode each value: `encodeURIComponent(r.company) + ":" + encodeURIComponent(r.org_name)` and decode on the other side.
2. Pass a JSON blob via stdin or a temp file (`--target-companies-file /tmp/<runId>.json`).
3. Validate the regex at insert time in the `companies` table.

Option 1 is the lowest-friction change.

data-pipeline/orchestrators/backfill.js
```diff
@@ -0,0 +1,167 @@
...
+        '--mode', 'backfill',
+        '--target-companies', targetCompaniesArg,
+        '--run-id', runId.toString(),
+      ],
+      logger
+    );
+
+    // ── Handle Ingester exit ──────────────────────────────────────────────────
+    if (exitCode === 0) {
+      await markCompleted(client, CASSANDRA_KEYSPACE, runId, targetRows);
```
tomerp20, line 155: **[HIGH] — Run is silently left `in_progress` if `markCompleted` throws**

If `spawnIngester` returns `exitCode === 0` but the subsequent `markCompleted` call throws (transient Cassandra error, batch size rejection, etc.), the outer `catch` at line 154 logs fatal and exits 1. The `backfill_runs` row is still `in_progress`, even though every hour has been ingested.

The next nightly run will re-discover this via the resume path, find `getMaxCompletedDate() === <yesterday>`, compute `startDate = nextDay(yesterday) > yesterday`, and short-circuit to `markCompleted` again — so the system *does* self-heal.

This is correct behaviour, but the recovery path is not obvious from the code. Add a one-line comment here pointing at the short-circuit on line 120:

```js
// If markCompleted throws here the run stays in_progress; next nightly
// run will hit the resume + "all dates covered" short-circuit on line 120
// and retry the flip. The companies UPDATEs are idempotent, so this is safe.
await markCompleted(client, CASSANDRA_KEYSPACE, runId, targetRows);
```

data-pipeline/orchestrators/backfill.js
```diff
@@ -0,0 +1,167 @@
...
+      runId = latestRun.runId;
+      targetRows = latestRun.targetRows;
+
+      const maxDate = await getMaxCompletedDate(client, CASSANDRA_KEYSPACE, runId);
+      if (maxDate !== null) {
+        startDate = nextDay(maxDate);
+      } else {
+        // In-progress run with no completed dates — restart from earliest on disk
+        try {
+          startDate = earliestOnDisk(GHARCHIVE_DIR);
```
tomerp20, line 95: **[MEDIUM] — Resume with no progress restarts every day under same run_id**

When an `in_progress` run has no rows in `backfill_progress`, we restart from `earliestOnDisk(GHARCHIVE_DIR)`. That re-ingests every day previously attempted under the same `run_id`.

This is safe **only if** the Ingester is idempotent across re-runs for the same `(run_id, date)` partition — which is the contract from PR3.A but is not visible from this PR.

Add a brief code comment pinning the assumption:
```js
// Safe to restart from earliestOnDisk: the Ingester writes `(run_id, date)`-
// keyed rows in backfill_progress, and PR3.A guarantees idempotency for
// repeated writes under the same key (see data-pipeline/ingester/...).
```

That way a future reader does not need to chase the contract across PRs.

tomerp20 (commented):
## Security Review — PASSED with notes

OWASP review of `backfill.js`, `lib/run-state.js`, `lib/disk-bounds.js`, `lib/companies-state.js`. This is an internal cron-driven service with no inbound HTTP surface, no user input, and no external network egress beyond Cassandra. Threat model is narrow: command injection, CQL injection, secret leakage, resource exhaustion.

### Findings

**A03 — Injection (CQL)**
- All parameterised values use `prepare: true` placeholders. No string concatenation of user data into queries.
- `keyspace` is interpolated via template literals (`\`${keyspace}.backfill_runs\``). `keyspace` comes only from `process.env.CASSANDRA_KEYSPACE` and is never derived from request/user input, so this is not exploitable today. **Recommendation:** validate keyspace against `/^[a-z_][a-z0-9_]{0,47}$/` at startup so a future code change that pipes user input here cannot regress into CQL injection. Low priority.

**Command injection**
- `spawnIngester` (existing) passes argv as an array — no shell interpolation. Safe.
- `targetCompaniesArg` is passed via a single argv slot. Not a shell-injection vector. (Parsing brittleness flagged separately in the backend review.)

**Secret / sensitive data exposure**
- Logger never logs `process.env`, credentials, or connection strings.
- Cassandra error catches log only `err.message` (drops stack — see backend review LOW for the trade-off).
- `org_name` / `company` are public identifiers (GitHub org slugs), not PII. Logging is safe.

**Resource exhaustion / DoS (self-inflicted)**
- `readUninitialised` uses `ALLOW FILTERING` — unbounded full-table scan. Could time out as the `companies` table grows. Cron-driven, not user-facing, so not an availability vector for end users. Flagged in backend review as a follow-up.
- `target_rows` collection in `backfill_runs` is bounded by Cassandra's 64K collection limit; not currently a concern.

**Path traversal**
- `GHARCHIVE_DIR` is read from env only and never combined with user-supplied path components. `enumerateAllOnDisk` walks under the root with `path.relative` for canonicalisation. Safe.

**A01 / A07 — Auth / Access control**
- Not applicable. No HTTP surface, no authentication boundary. Operates with whatever Cassandra credentials the process has via driver config.

**A05 — Misconfiguration**
- `IS_PROD` toggles `pino-pretty` off in production. Correct.
- No CORS / TLS / cookie surfaces in this PR.

**Supply chain**
- New dependency `disk-layout` is a local `file:../shared/disk-layout` — first-party code, no external footprint added.

### Verdict
No exploitable security issues. One advisory (validate keyspace at startup) and one operational note (`ALLOW FILTERING`) carried over to the backend review.

tomerp20 (commented):
## Code Review Summary — PR #222

Two-phase orchestrated review on commit `15faed8`, with fix commit `d1ebc67` pushed by the backend reviewer.

### Agents that ran

| Agent | Phase 1 (Review) | Phase 2 (Fix) |
|---|---|---|
| Backend code review (acting as `backend-code-reviewer`) | ran | ran |
| Security review (orchestrator) | ran | n/a |
| React code review | skipped (no frontend files changed) | n/a |

### Issue counts

| Severity | Found | Fixed | Skipped |
|---|---:|---:|---:|
| HIGH | 2 | 2 | 0 |
| MEDIUM | 5 | 2 | 3 |
| LOW | 4 | 2 | 2 |
| Security | 0 exploitable, 1 advisory | 0 | 1 |

### Fixes pushed in `d1ebc67`

1. **HIGH** — `lib/run-state.js` `markCompleted` no longer uses a multi-partition LOGGED BATCH; per-row idempotent UPDATEs followed by a single `backfill_runs` UPDATE.
2. **MEDIUM** — `lib/run-state.js` `findLatestRun` now throws on null/empty `target_rows` instead of silently returning an empty resume payload.
3. **HIGH** — `backfill.js` documents the self-healing recovery path when `markCompleted` throws after a successful Ingester run.
4. **MEDIUM** — `backfill.js` documents the Ingester idempotency contract that makes the resume-from-earliest restart safe.
5. **LOW** — `backfill.js` fatal logs now pass the full `err` object (preserving stack traces under pino) rather than `err.message`.
6. **LOW** — `backfill.js` best-effort `client.shutdown()` before `process.exit(1)` on Cassandra connect failure.

### Skipped — requires human attention

- **MEDIUM — `backfill.js:128` `--target-companies` encoding brittleness.**
  *Reason:* changing the encoding requires a coordinated change in the Ingester's argument parser (separate PR / module). Cross-cutting risk if fixed in isolation.
  *What needs to be done:* either URL-encode each `company:org_name` pair on both sides, or switch to a JSON-via-file/stdin transport. Track as a follow-up issue.

- **MEDIUM — `lib/companies-state.js:8-13` `ALLOW FILTERING` on `readUninitialised`.**
  *Reason:* the existing `readInitialised` uses the same pattern; changing both is a schema/index change outside the scope of this PR. Defer to a dedicated migration that adds a secondary index on `(initialized, active)` once the `companies` table grows.

- **MEDIUM — `lib/run-state.js` snapshot semantics in `markCompleted`.**
  *Reason:* `markCompleted` flips `initialized=true` for every row in the snapshotted `target_rows`, even if the row has since been deactivated. This is likely intentional but is not in code. Worth a follow-up to either (a) document the snapshot semantics explicitly in an ADR, or (b) intersect with the current `active=true` set at flip time. Either choice has product implications.

- **LOW (Security advisory) — keyspace identifier validation.**
  *Reason:* not exploitable today (keyspace is env-only), but pre-emptive hardening. Add `/^[a-z_][a-z0-9_]{0,47}$/` validation at startup. Defer until any code path forwards user input to the keyspace slot.

- **LOW — Argv contract for `--range` positional pair.**
  *Reason:* requires a contract doc, not a code change. Add a `data-pipeline/orchestrators/README.md` (or a comment near the `spawnIngester` call) describing the Ingester argv schema as a follow-up.

### Phase 1 evidence

- Backend review summary: https://github.com/tomerp20/jobflow/pull/222#issuecomment- (above in this thread)
- Inline findings on lines: `lib/run-state.js:17`, `lib/run-state.js:44`, `backfill.js:91`, `backfill.js:128`, `backfill.js:146`
- Security review: PASSED with one advisory (above in this thread)

### Phase 2 verification

- Both files pass `node --check`.
- No test suite was run (CI handles tests).
- Fix commit limited to the two files (`backfill.js`, `lib/run-state.js`) — no cross-cutting changes.

Awaiting human review and merge.

PR #228, opened 2026-05-23
[bug] Backfill Orchestrator endDate uses yesterdayUtc; fails on dates with no files on disk

### LINKED ISSUE #226: [bug] Backfill Orchestrator endDate uses yesterdayUtc; fails on dates with no files on disk
## Severity

HIGH for operational reliability. Found during the first 7-day backfill test on the Linux box. The pipeline shipped and worked correctly for the data on disk; this issue is purely about a date-range mismatch that causes a run to be marked failed even when the operator's intent was satisfied.

## Repro

Already happened in production-style test:
1. Wipe `/mnt/hdd/gharchive/`.
2. Pre-fetch only 7 specific dates (e.g., `node fetcher.js --range 2026-05-15 2026-05-21`).
3. Run `node orchestrators/backfill.js` against a fresh `companies` table with `initialized=false` rows.
4. The orchestrator computes `startDate = earliestOnDisk = 2026-05-15` (correct) but `endDate = yesterdayUtc() = 2026-05-22` (off-by-one for the fetched range).
5. The Ingester walks May 15 → May 21 fine, then opens `/mnt/hdd/gharchive/2026/05/22/0.json.gz` → `ENOENT` → fails the date.
6. Orchestrator marks the run `failed`. Operator now has 7 dates' worth of data in `company_events` but `companies.initialized` was never flipped to `true`.

## Root cause

In `data-pipeline/orchestrators/backfill.js`, the date range is computed via:

```js
const startDate = earliestOnDisk(GHARCHIVE_DIR);
const endDate = yesterdayUtc();
```

The first half (`earliestOnDisk`) correctly reflects "the earliest date on disk." The second half (`yesterdayUtc`) hard-codes "yesterday" regardless of whether the disk has files for yesterday.

This works only if the operator always pre-fetches up through yesterday before invoking the backfill. The `docs/InjestionPlan.md` §6.4 design implicitly assumed this (the runner script calls `node fetcher.js --range ... yesterday` before `node ingester.js --range ... --mode backfill`). But the orchestrator itself does not enforce or detect this assumption.

## Fix

Two-part fix in `data-pipeline/orchestrators/lib/disk-bounds.js` + `backfill.js`:

### 1. Add `latestOnDisk()` to `disk-bounds.js`

Mirror of the existing `earliestOnDisk()` — walks `GHARCHIVE_DIR`, finds the most recent hour ID present, returns its date portion (`YYYY-MM-DD`).

### 2. Use `latestOnDisk()` in `backfill.js`

```diff
-const endDate = yesterdayUtc();
+const endDate = latestOnDisk(GHARCHIVE_DIR);
```

### 3. Add WARN log when archive is behind yesterday

```js
const yesterday = yesterdayUtc();
const latest = latestOnDisk(GHARCHIVE_DIR);
if (latest < yesterday) {
  logger.warn(
    { latestOnDisk: latest, yesterdayUtc: yesterday },
    'archive does not reach yesterday — backfill will stop at latestOnDisk; run the Fetcher to catch up if you want to include more recent dates'
  );
}
const endDate = latest;
```

## Why Fix 1 (`latestOnDisk`) over alternatives

I considered four approaches during the design discussion:
- **Fix 1 (recommended):** `endDate = latestOnDisk`. Mechanically simple. Matches "process what's on disk" mental model.
- Fix 2: skip dates with no files inside the Ingester. Rejected — hides real bugs (partial fetches, disk-full).
- Fix 3: pre-check + fail-fast at startup. Rejected — doesn't fix the problem, just detects it earlier.
- Fix 4: auto-fetch missing dates from inside the orchestrator. Rejected — tight coupling between orchestrator and fetcher, brittle.

Fix 1 also matches the eventual PR4 shell-script pattern (`fetcher --range A B` followed by `backfill.js`): the fetcher fills the disk, the backfill processes whatever is there.

## Acceptance criteria

- [ ] `data-pipeline/orchestrators/lib/disk-bounds.js` exports a `latestOnDisk(rootDir)` function symmetric to `earliestOnDisk`.
- [ ] `latestOnDisk` returns the chronologically latest `YYYY-MM-DD` for which at least one `.json.gz` file exists under `rootDir`. Throws if directory is empty.
- [ ] `data-pipeline/orchestrators/backfill.js` uses `latestOnDisk(GHARCHIVE_DIR)` as `endDate`, not `yesterdayUtc()`.
- [ ] If `latestOnDisk < yesterdayUtc()`, the orchestrator emits a structured WARN log (visible at default `LOG_LEVEL=info`) with both values present.
- [ ] Existing tests in `disk-bounds.js` (if any) pass; the orchestrator continues to handle empty-disk + nothing-to-do paths correctly.
- [ ] Smoke test on the live Linux box reproduces the failure on the current state (run-marked-failed), then with the fix applied, re-runs cleanly: backfill walks May 15 → latestOnDisk only, completes successfully, flips `initialized=true`.

## Related

- Discovered during 7-day backfill test on the Linux box. Run state is currently marked `failed` post-discovery (operator workaround: pre-fetch yesterday + re-run, which works fine but isn't sustainable).
- Adjacent to #221 (Ingester exit code on ENOENT): the Ingester correctly exits non-zero, the orchestrator correctly marks the run as failed. The issue isn't the failure path; it's that the range shouldn't have included the unfetched date in the first place.

## Suggested model tier

Sonnet. Small mechanical change, ~15 lines across 2 files. Must include a smoke test against a real Cassandra container demonstrating both behaviors (warn when behind, fail-clean when disk is empty).

tomerp20:
## Summary
- Adds `latestOnDisk(rootDir)` to `disk-bounds.js` — mirrors `earliestOnDisk`, returns the latest YYYY-MM-DD present on disk (throws if empty)
- Replaces `yesterdayUtc()` as `endDate` in `backfill.js` with `latestOnDisk(GHARCHIVE_DIR)` so the orchestrator only walks dates that are actually on disk
- Emits a structured `WARN` log when the archive lags behind yesterday, prompting operators to run the Fetcher

## Acceptance criteria
- [x] `data-pipeline/orchestrators/lib/disk-bounds.js` exports a `latestOnDisk(rootDir)` function symmetric to `earliestOnDisk`
- [x] `latestOnDisk` returns the chronologically latest `YYYY-MM-DD` for which at least one `.json.gz` file exists under `rootDir`. Throws if directory is empty
- [x] `data-pipeline/orchestrators/backfill.js` uses `latestOnDisk(GHARCHIVE_DIR)` as `endDate`, not `yesterdayUtc()`
- [x] If `latestOnDisk < yesterdayUtc()`, the orchestrator emits a structured WARN log with both values present
- [x] Existing tests pass; empty-disk + nothing-to-do paths continue to be handled correctly
- [ ] Smoke test on the live Linux box (operator manual verification)

## Related
Closes #226

### DIFF
--- a/data-pipeline/orchestrators/backfill.js
+++ b/data-pipeline/orchestrators/backfill.js
@@ -5,7 +5,7 @@ import pino from 'pino';
 import { readUninitialised } from './lib/companies-state.js';
 import { spawnIngester } from './lib/spawn-ingester.js';
 import { findLatestRun, createNewRun, markCompleted, markFailed, getMaxCompletedDate } from './lib/run-state.js';
-import { earliestOnDisk, latestOnDisk, yesterdayUtc } from './lib/disk-bounds.js';
+import { earliestOnDisk, latestCompleteDateOnDisk, diskBounds, yesterdayUtc } from './lib/disk-bounds.js';
 
 const { Client } = cassandra;
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
@@ -74,30 +74,22 @@ async function main() {
 
     // ── Determine lifecycle path ──────────────────────────────────────────────
     const latestRun = await findLatestRun(client, CASSANDRA_KEYSPACE);
-
     const yesterday = yesterdayUtc();
-    let endDate;
-    try {
-      const latest = latestOnDisk(GHARCHIVE_DIR);
-      if (latest < yesterday) {
-        logger.warn(
-          { latestOnDisk: latest, yesterdayUtc: yesterday },
-          'archive does not reach yesterday — backfill will stop at latestOnDisk; run the Fetcher to catch up if you want to include more recent dates'
-        );
-      }
-      endDate = latest;
-    } catch (err) {
-      logger.fatal({ err }, 'cannot determine end date');
-      await client.shutdown().catch(() => {});
-      process.exit(1);
-    }
 
-    let runId, targetRows, startDate;
+    let runId, targetRows, startDate, endDate;
 
     if (latestRun && latestRun.status === 'in_progress') {
       // ── Resume path ─────────────────────────────────────────────────────────
       runId = latestRun.runId;
       targetRows = latestRun.targetRows;
+
+      try {
+        endDate = latestCompleteDateOnDisk(GHARCHIVE_DIR);
+      } catch (err) {
+        logger.fatal({ err }, 'cannot determine end date');
+        await client.shutdown().catch(() => {});
+        process.exit(1);
+      }
 
       const maxDate = await getMaxCompletedDate(client, CASSANDRA_KEYSPACE, runId);
       if (maxDate !== null) {
@@ -116,14 +108,22 @@ async function main() {
         }
       }
 
+      if (endDate < yesterday) {
+        logger.warn(
+          { latestCompleteDateOnDisk: endDate, yesterdayUtc: yesterday },
+          'archive does not reach yesterday — backfill will stop at latestCompleteDateOnDisk; run the Fetcher to catch up if you want to include more recent dates'
+        );
+      }
+
       logger.info({ runId: runId.toString(), startDate, endDate, companies: targetRows.length }, 'resuming run');
     } else {
       // ── Fresh path (no row, completed, or failed) ────────────────────────────
-      let earliest;
+      // Single enumerateAllOnDisk walk for both bounds.
+      let bounds;
       try {
-        earliest = earliestOnDisk(GHARCHIVE_DIR);
+        bounds = diskBounds(GHARCHIVE_DIR);
       } catch (err) {
-        logger.fatal({ err }, 'cannot determine start date for fresh run');
+        logger.fatal({ err }, 'cannot determine disk bounds for fresh run');
         await client.shutdown().catch(() => {});
         process.exit(1);
       }
@@ -131,7 +131,15 @@ async function main() {
       const created = await createNewRun(client, CASSANDRA_KEYSPACE, uninitialised);
       runId = created.runId;
       targetRows = created.targetRows;
-      startDate = earliest;
+      startDate = bounds.earliest;
+      endDate = bounds.latestComplete;
+
+      if (endDate < yesterday) {
+        logger.warn(
+          { latestCompleteDateOnDisk: endDate, yesterdayUtc: yesterday },
+          'archive does not reach yesterday — backfill will stop at latestCompleteDateOnDisk; run the Fetcher to catch up if you want to include more recent dates'
+        );
+      }
 
       logger.info({ runId: runId.toString(), startDate, endDate, companies: targetRows.length }, 'starting fresh run');
     }
--- a/data-pipeline/orchestrators/backfill.js
+++ b/data-pipeline/orchestrators/backfill.js
@@ -5,7 +5,7 @@ import pino from 'pino';
 import { readUninitialised } from './lib/companies-state.js';
 import { spawnIngester } from './lib/spawn-ingester.js';
 import { findLatestRun, createNewRun, markCompleted, markFailed, getMaxCompletedDate } from './lib/run-state.js';
-import { earliestOnDisk, yesterdayUtc } from './lib/disk-bounds.js';
+import { earliestOnDisk, latestOnDisk, yesterdayUtc } from './lib/disk-bounds.js';
 
 const { Client } = cassandra;
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
@@ -74,7 +74,23 @@ async function main() {
 
     // ── Determine lifecycle path ──────────────────────────────────────────────
     const latestRun = await findLatestRun(client, CASSANDRA_KEYSPACE);
+
     const yesterday = yesterdayUtc();
+    let endDate;
+    try {
+      const latest = latestOnDisk(GHARCHIVE_DIR);
+      if (latest < yesterday) {
+        logger.warn(
+          { latestOnDisk: latest, yesterdayUtc: yesterday },
+          'archive does not reach yesterday — backfill will stop at latestOnDisk; run the Fetcher to catch up if you want to include more recent dates'
+        );
+      }
+      endDate = latest;
+    } catch (err) {
+      logger.fatal({ err }, 'cannot determine end date');
+      await client.shutdown().catch(() => {});
+      process.exit(1);
+    }
 
     let runId, targetRows, startDate;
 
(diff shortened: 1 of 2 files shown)


tomerp20 (commented):
## Backend Review — STANDARD

Reviewed data-pipeline orchestrator changes for PR #228 (data-pipeline JS, not under `backend/` — backend-checklist categories like route auth/Knex/migrations don't apply; focus is correctness, blocking I/O, exit codes).

ESLint: not run — no ESLint config under `data-pipeline/` (consistent with memory note for this subproject).

### Issues by Severity

**HIGH**
- `backfill.js:88` (`endDate = latest` where `latest = latestOnDisk(...)`): `latestOnDisk` returns date granularity (`YYYY-MM-DD`), but the latest date on disk almost always has only partial hours present (the Fetcher runs hourly — at 14:00 UTC on 2025-05-23 only hours 0–13 of that date exist). The Ingester is then invoked with `--range start 2025-05-23`, which expands via `enumerateRange` to hours 0–23 of 2025-05-23. `processHour` will hit `ENOENT` on the first missing hour (`createReadStream(filePath)` → `fileStream.on('error', reject)` in `ingester.js:270`), which propagates through `runRange` and stops the backfill. This reproduces the same class of failure the PR set out to fix.

**MEDIUM**
- `disk-bounds.js:12-20`: `latestOnDisk` is being newly defined here even though `data-pipeline/shared/disk-layout/index.js:49` already exports a `latestOnDisk(rootDir)` function (used by the Fetcher at `fetcher.js:67`). Two functions with the same exported name and divergent behavior (the shared one returns the full hour-ID or `null`; the new one returns a date slice or throws) is a footgun for future readers. Either reuse the shared one (and slice/throw at the call site) or pick a non-colliding name like `latestDateOnDisk`.

**LOW**
- `backfill.js:80` and `backfill.js:108`: `enumerateAllOnDisk(GHARCHIVE_DIR)` is now walked twice in the fresh path (once via `latestOnDisk`, once via `earliestOnDisk`). For a multi-year archive each call is a full recursive `readdirSync` walk. Not breaking, but cheap to avoid by computing both bounds from a single `enumerateAllOnDisk` call.

### Testing Gaps
- No tests anywhere under `data-pipeline/` (known — see project memory). New `latestOnDisk` wrapper has no unit coverage for the empty-dir-throws contract or for the date-slice behavior.
- Acceptance criterion "Existing tests pass" cannot be verified — no test infra exists in this subproject.
- The partial-latest-day failure mode above would be caught by even a minimal smoke test: seed `GHARCHIVE_DIR` with files for `YYYY/MM/DD/0.json.gz` only (no other hours), run the orchestrator, assert it does not crash trying to read hour 1.

data-pipeline/orchestrators/backfill.js
```diff
@@ -74,7 +74,23 @@ async function main() {
...
+    let endDate;
+    try {
+      const latest = latestOnDisk(GHARCHIVE_DIR);
+      if (latest < yesterday) {
+        logger.warn(
+          { latestOnDisk: latest, yesterdayUtc: yesterday },
+          'archive does not reach yesterday — backfill will stop at latestOnDisk; run the Fetcher to catch up if you want to include more recent dates'
+        );
+      }
+      endDate = latest;
```
tomerp20: **[HIGH] — endDate at date granularity will crash on partial latest day**

`latestOnDisk` returns `YYYY-MM-DD` only, but the Fetcher writes hourly — so the latest date on disk almost always has only some of its 24 hours present at orchestrator-run time.

The Ingester expands the range via `enumerateRange(start, end)` which always emits hours 0–23 for every date inclusive. `processHour` then calls `createReadStream(filePath)` on each, and `fileStream.on('error', reject)` (`ingester.js:270`) rejects on `ENOENT` — stopping the backfill mid-range. That reproduces the failure class this PR was meant to fix.

**Suggested fix:** use the existing `latestOnDisk` from `shared/disk-layout` (which returns the full hour-ID) to find the latest complete date — e.g. last date whose 23rd hour exists, or just `endDate = previous date if the latest day is partial`. Alternatively let the Ingester treat ENOENT as 'end of available hours' rather than a fatal error.

data-pipeline/orchestrators/lib/disk-bounds.js
```diff
@@ -9,6 +9,15 @@ export function earliestOnDisk(rootDir) {
   return ids[0].slice(0, 10); // YYYY-MM-DD
 }
 
+export function latestOnDisk(rootDir) {
+  const ids = enumerateAllOnDisk(rootDir);
+  if (ids.length === 0) {
+    throw new Error(`no .json.gz files found under GHARCHIVE_DIR: ${rootDir}`);
+  }
+  // ids are sorted chronologically (earliest first); last element is the latest
+  return ids[ids.length - 1].slice(0, 10); // YYYY-MM-DD
```
tomerp20: **[MEDIUM] — Duplicate `latestOnDisk` function name**

`data-pipeline/shared/disk-layout/index.js:49` already exports a `latestOnDisk(rootDir)` that the Fetcher uses (`fetcher.js:67`). The two functions share a name but have different contracts:

- shared one → returns full hour ID (`YYYY-MM-DD-H`) or `null` on empty
- this one → returns `YYYY-MM-DD` or throws on empty

Same-name-divergent-behavior across the codebase is a footgun for readers and IDE auto-imports. Either reuse the shared `latestOnDisk` here (and `.slice(0,10)` at the call site) or rename to e.g. `latestDateOnDisk` for clarity.

data-pipeline/orchestrators/backfill.js
```diff
@@ -74,7 +74,23 @@ async function main() {
 
     // ── Determine lifecycle path ──────────────────────────────────────────────
     const latestRun = await findLatestRun(client, CASSANDRA_KEYSPACE);
+
     const yesterday = yesterdayUtc();
+    let endDate;
+    try {
```
tomerp20: **[LOW] — Duplicate directory walk in fresh path**

In the fresh path, `enumerateAllOnDisk(GHARCHIVE_DIR)` is now walked twice — once here via `latestOnDisk` and once at line 108 via `earliestOnDisk`. Each is a full recursive `readdirSync`. For a multi-year archive that's avoidable I/O.

Could compute both bounds from a single `enumerateAllOnDisk` call (take `ids[0]` and `ids.at(-1)`).

tomerp20 (commented):
## Code Review Orchestrator — Final Status

| Agent | Issues Found | Fixed | Skipped |
| 🔒 Security | 0 | — | — |
| 🔧 Backend | 3 | 3 | 0 |

### Issues fixed (commit `7b16a07`)

| Severity | Finding |
|----------|---------|
| **HIGH** | `latestOnDisk` returned a date with only partial hours on disk → Ingester would ENOENT on missing hours, reproducing the original failure class. Fixed by replacing with `latestCompleteDateOnDisk(rootDir)` — walks backwards and returns the first date whose hour 23 file is present. |
| **MEDIUM** | Name collision: `disk-layout/index.js` already exports `latestOnDisk` with a different contract (returns full hour ID, returns `null` on empty). New function renamed to `latestCompleteDateOnDisk` to avoid ambiguity. |
| **LOW** | `latestOnDisk` + `earliestOnDisk` each walked `enumerateAllOnDisk` separately in the fresh path. Consolidated to a single `diskBounds(rootDir)` helper that returns both endpoints from one walk. |

### Skipped fixes
None.

### Manual action required
None. PR is ready for human review.

Issue #35, opened 2026-04-09
feat: todo count badge on kanban card preview


tomerp20:
## Overview

When a job card has linked todos, the card on the kanban board gives no indication of this. A user must open the card to discover pending tasks. This feature adds a compact badge to the card preview that shows the number of linked todos and communicates at a glance whether any are still active.

## Goals

- Surface linked todo counts directly on the kanban card without opening it
- Distinguish visually between cards with pending tasks vs. all-done tasks
- Keep the card UI compact — no extra API calls, no prop threading, no new layout sections

## Non-Goals

- No new page or modal
- No interaction on the badge itself (clicking it does nothing — open the card to manage todos)
- No changes to `CardDetail` (already has its own Tasks section)
- No changes to the `/tasks` page

---

## Design Decisions (resolved)

| Decision | Choice | Reason |
|---|---|---|
| Data source | Embed in `GET /api/cards` via subquery | Zero extra roundtrips; indexed `card_id` on `todo_items` makes it cheap |
| What is counted | All linked todos (`total_todo_count`) | Show full picture; icon/color signals whether any are still active |
| Visual style | Pill badge — icon + count | Matches existing footer pattern (`<Clock size={10} /> 4d ago`) |
| Active state | Violet `ListTodo` icon | Violet is the only unclaimed color on the card; `ListTodo` = open work |
| All-done state | Muted gray `CheckSquare` icon | Communicates completion without drawing attention |
| No todos | Render nothing | No noise for cards with no linked tasks |
| Badge position | Footer row, right side | Left side has dates; right side is currently empty — clean split |
| Count display | `totalTodoCount` (all todos) | Icon conveys active/done state; count conveys scope |

---

### 1. Badge States

```
totalTodoCount === 0          → render nothing
activeTodoCount > 0           → <ListTodo size={10} className="text-violet-500" /> {totalTodoCount}  (violet pill)
activeTodoCount === 0         → <CheckSquare size={10} className="text-gray-400" /> {totalTodoCount}  (gray pill)
```

Both states use the same compact inline style as existing footer items — icon + text, no background fill needed.

### 2. Footer Row Layout

Current footer row (left side only):
```
[Clock 4d ago]  [CalendarClock in 2 days]          (empty right side)
```

After this feature:
```
[Clock 4d ago]  [CalendarClock in 2 days]           [ListTodo 3]
```

The badge sits on the right side, aligned via `justify-between` (already present on the footer row).

---

### Backend — `cardService.ts`

[...]

Issue #34, opened 2026-04-09
feat: wire autocomplete into SearchBar


tomerp20:
## Parent PRD

#29

## What to build

Integrate `useAutocomplete` and `AutocompleteDropdown` into the existing `SearchBar` component (`frontend/src/components/Search/SearchBar.tsx`). This is the final slice that makes the feature live end-to-end.

- Call `useAutocomplete(value)` inside SearchBar to get `{ suggestions, isLoading }`
- Render `<AutocompleteDropdown />` below the search input, passing suggestions, loading state, and callbacks
- On suggestion selected: replace the input value and immediately fire the existing search trigger (no extra Enter needed)
- Lazy init triggers on SearchBar focus (already handled by the hook)

## Acceptance criteria

- [ ] Typing Hebrew characters in the SearchBar shows the autocomplete dropdown
- [ ] Loading spinner appears if the user types before the trie is ready
- [ ] Selecting a suggestion replaces the input value and triggers the search immediately
- [ ] Pressing Escape or clicking outside closes the dropdown
- [ ] Existing SearchBar behavior (debounced search, filters) is unaffected
- [ ] Full end-to-end flow works in the running app

## Blocked by

- Blocked by #32 (useAutocomplete hook)
- Blocked by #33 (AutocompleteDropdown component)

## User stories addressed

- User story 1, 2, 3, 4

Issue #32, opened 2026-04-09
feat: useAutocomplete hook with lazy init, debounce, and loading state


tomerp20:
## Parent PRD

#29

## What to build

Implement the `useAutocomplete` hook (`frontend/src/hooks/useAutocomplete.ts`) that manages the singleton Web Worker lifecycle and exposes suggestions + loading state to any component.

- On first focus (lazy init): fetch words from `GET /autocomplete/words`, send to worker to build trie
- Set `isLoading: true` from the moment lazy init triggers until the worker sends `{ type: 'ready' }`
- Debounce prefix queries to the worker by **150ms** before sending via `postMessage` (prevents CPU spikes from per-keystroke trie traversals — separate from the SearchBar's existing 300ms API debounce)
- Returns `{ suggestions: string[], isLoading: boolean }`
- Reusable: any input field can call `useAutocomplete(value)` to get suggestions

## Acceptance criteria

- [ ] `frontend/src/hooks/useAutocomplete.ts` exists
- [ ] Trie is not built until the first focus event (lazy initialization)
- [ ] `isLoading` is `true` from init start until `{ type: 'ready' }` is received from worker
- [ ] Prefix queries are debounced by 150ms before being sent to the worker
- [ ] Returns `{ suggestions: string[], isLoading: boolean }`
- [ ] Hook is reusable — not coupled to SearchBar specifically

## Blocked by

- Blocked by #31 (depends on the trie worker)

## User stories addressed

- User story 1, 4, 5

PR #229, opened 2026-05-23
[Cassandra pipeline PR4] Production deployment wrapper (cron + flock + bootstrap)

### LINKED ISSUE #208: [Cassandra pipeline PR4] Production deployment wrapper (cron + flock + bootstrap)
> Source-of-truth docs: `docs/CassandraPlan.md` §2 + §8, `docs/InjestionPlan.md` §6 + §7.1 + §9. All on `main`. This PRD records the locked outcomes of the PR4 grill.

## Problem Statement

PR1–PR3 ship four Node binaries (Fetcher, Ingester, Backfill Orchestrator, Hourly Orchestrator) that together form a complete data pipeline — but only when invoked manually on a developer's MacBook. None of them know how to schedule themselves, how to coordinate with each other on a shared host, where to write logs, or how to be safely deployed on a fresh server. Without an OS-level wrapper, the pipeline isn't usable as a "set it and forget it" production tool. It's a set of tools that requires constant human babysitting.

The hard requirement from `docs/InjestionPlan.md` §1 — "the Hourly and Backfill jobs must never run simultaneously" — is also unfulfilled. On MacBook with a single human operator there is no race possible by construction, but on a Linux box with cron firing the two jobs on independent schedules, they will race unless something coordinates them.

PR4 fills that gap with the smallest possible production deployment wrapper: idempotent setup, flock-based lock coordination, disk-space safety guards, structured log destinations, and a cron schedule that turns the four binaries into a self-running pipeline on the Xubuntu target box.

## Solution

A small set of shell scripts and a cron template that wrap the four already-shipped Node binaries into a cron-driven production deployment on the Linux target. Five files total:

- **`data-pipeline/scripts/bootstrap.sh`** — one-time idempotent setup script run by the Operator on a fresh Xubuntu box. Verifies prerequisites, creates directories, brings up Cassandra via docker-compose, applies all schema migrations, and installs the cron file with path/user substitutions.
- **`data-pipeline/scripts/run-backfill.sh`** — invoked by the backfill cron job. Checks disk space, runs the Fetcher with the full backfill date range, then runs the Backfill Orchestrator. Exits with the last command's exit code.
- **`data-pipeline/scripts/run-hourly.sh`** — invoked by the hourly cron job. Checks disk space, runs the Fetcher with `--catchup`, then runs the Hourly Orchestrator.
- **`data-pipeline/cron/jobflow.cron.template`** — the cron schedule. Contains the `flock`-based lock coordination per `docs/InjestionPlan.md` §6.1–6.3. `@JOBFLOW_ROOT@` and `@USER@` placeholders substituted by `bootstrap.sh` during install.
- **`data-pipeline/SETUP.md`** — runbook for preparing a fresh Xubuntu box (mount the HDD at `/mnt/hdd`, install Node ≥ 20.6, Docker, `cqlsh`, run `bootstrap.sh`).

Cron jobs run as the Operator's user (looked up via `whoami` at bootstrap time), with stdout/stderr appended to `~/jobflow-logs/{backfill,hourly,hourly-skips}.log` via cron-level shell redirects. The shell scripts themselves know nothing about log paths — that's the cron line's responsibility.

The locks live at `/var/lock/jobflow-backfill.lock` and `/var/lock/jobflow-hourly.lock`. The hourly cron line uses a nested `flock` to check both its own lock AND the backfill lock before proceeding; if the backfill lock is held, the hourly skips this interval and logs to `hourly-skips.log`. The Ingester's `--catchup` mode handles the resulting data backlog on the next successful hourly invocation.

The disk-space safety check runs at the top of each shell script before any work starts. 50 GB minimum free on `/mnt/hdd` for backfill; 10 GB for hourly. If the check fails, the script exits non-zero with a FATAL message and the cron job is logged as a failure; the next interval tries again. Cassandra's own write-failure path covers SSD pressure; bootstrap does not check SSD space.

## User Stories

1. As an Operator, I want a single command (`bootstrap.sh`) to set up a fresh Xubuntu box for production, so that I don't have to remember a dozen manual steps every time I rebuild or reinstall.
2. As an Operator, I want `bootstrap.sh` to be idempotent — safe to re-run on a partially-set-up box — so that recovery from a failed setup is just "run it again."
3. As an Operator, I want `bootstrap.sh` to fail loudly with a clear message if Node, Docker, or cqlsh is missing, so that I know which prerequisite to install before retrying.
4. As an Operator, I want `bootstrap.sh` to verify that `/mnt/hdd` is a real mount point (not just an existing directory) before proceeding, so that the Fetcher doesn't silently write to the root filesystem and fill the SSD.
5. As an Operator, I want `bootstrap.sh` to create `/mnt/hdd/gharchive/` and `~/jobflow-logs/` if they don't exist, so that the pipeline has the directory structure it expects without me running `mkdir` manually.
6. As an Operator, I want `bootstrap.sh` to bring up the Cassandra container via `docker compose up -d` and wait for the healthcheck to pass, so that the schema migrations don't fail because Cassandra isn't ready yet.
7. As an Operator, I want `bootstrap.sh` to apply all schema migrations in numeric order, so that the Cassandra keyspace ends up in the correct shape regardless of starting state.
8. As an Operator, I want the schema migrations to be idempotent (re-applicable without breaking), so that re-running `bootstrap.sh` doesn't drop tables or break the schema.
9. As an Operator, I want `bootstrap.sh` to install the cron file to `/etc/cron.d/jobflow` with my paths and username substituted in, so that I don't have to hand-edit the cron file every time the repo path changes.
10. As an Operator, I want `bootstrap.sh` to print a clear summary of what it did and how to verify the install (cron status, log file locations, schema check), so that I can confirm everything is wired up correctly.
11. As an Operator, I want `bootstrap.sh` to be run exactly once per fresh box but safe to re-run later (e.g., after a schema migration), so that the same script covers initial setup and subsequent re-runs.
12. As an Operator, I want the cron schedule to fire the Backfill once per night at 00:00, so that newly-added Companies are picked up overnight without my involvement.
13. As an Operator, I want the cron schedule to fire the Hourly Ingest every hour at :15, so that the analytics pipeline stays current within roughly an hour of GH Archive publishing.
14. As an Operator, I want the Hourly cron to skip cleanly when the Backfill is running, so that the two jobs never compete for Cassandra and the relay-race contract in ADR 0003 is preserved at the operational level.
15. As an Operator, I want the Hourly skip to be logged to a dedicated `hourly-skips.log` file, so that I can audit how many hours were skipped during long backfills without grepping the main log.
16. As an Operator, I want a second Backfill cron firing while a previous Backfill is still running to silently skip via flock, so that a slow Backfill doesn't get stacked behind itself.
17. As an Operator, I want both `run-backfill.sh` and `run-hourly.sh` to check `/mnt/hdd` free space at the start and exit non-zero with a FATAL message if below the threshold, so that the Fetcher never tries to download to a full disk.
18. As an Operator, I want the disk-space threshold to be higher for backfill (50 GB) than hourly (10 GB), so that a stressed disk doesn't block hourly trickles when a backfill couldn't run.
19. As an Operator, I want `run-backfill.sh` to invoke the Fetcher first (download missing range), then the Backfill Orchestrator, so that the Orchestrator always has the files it expects on disk when it walks dates.
20. As an Operator, I want `run-hourly.sh` to invoke the Fetcher with `--catchup` first, then the Hourly Orchestrator, so that the Hourly Orchestrator processes whatever the Fetcher just downloaded.
21. As an Operator, I want all stdout and stderr from every cron-fired job to land in append-forever log files under `~/jobflow-logs/`, so that I have a complete record of what the pipeline did without configuring log rotation.
22. As an Operator, I want log redirection done at the cron-line level (not inside the shell scripts), so that I can run the shell scripts directly during testing and see output on the terminal.
23. As an Operator, I want the shell scripts to use `$(dirname "$0")/..` for path resolution to sibling Node binaries, so that the scripts work regardless of where I clone the repo.
24. As an Operator, I want `~/jobflow-logs/` owned by my user (not root), so that I never need `sudo` to tail or grep my own pipeline's logs.
25. As an Operator, I want the cron file installed at `/etc/cron.d/jobflow` as a system-level crontab, so that an accidental `crontab -r` from my user shell doesn't wipe out the pipeline schedule.
26. As an Operator, I want `bootstrap.sh` to be the only command that needs `sudo`, so that the day-to-day operation of the pipeline runs entirely as my user.
27. As an Operator on a fresh Xubuntu box, I want a SETUP.md runbook that documents the manual prerequisite steps (mount the HDD, install Node/Docker/cqlsh, clone the repo, then `bootstrap.sh`), so that I have a single document to follow on each setup.
28. As an Operator, I want the cron file template to use placeholders like `@JOBFLOW_ROOT@` and `@USER@` that get substituted at install time, so that the file in the repo stays generic but the installed file is correct for my machine.
29. As a Developer of the deployment wrapper, I want the cron template to live in the repo (committed) while the installed cron file lives at `/etc/cron.d/jobflow` (not committed), so that the deployable file under version control is the template and the runtime file is the substituted output.
30. As a Developer, I want the shell scripts to use `set -euo pipefail`, so that any individual command failure aborts the whole script rather than silently moving on.
31. As an Operator, I want a failed cron run (e.g., disk space too low) to be visible via the cron log files and via a non-zero exit, so that I can grep for FATAL messages without setting up alerting.
32. As an Operator, I want `bootstrap.sh` to detect that Cassandra is already running and skip the `docker compose up` step, so that re-running bootstrap doesn't disrupt an active Cassandra container.
33. As a Developer, I want SETUP.md to include a "how to verify" section listing specific commands (`grep jobflow /etc/cron.d/jobflow`, `mountpoint /mnt/hdd`, `cqlsh -e "DESCRIBE KEYSPACE jobflow"`), so that I can confirm a successful bootstrap without guessing.
34. As an Operator, I want the cron jobs to run as my user account (not root), so that any files they create on the HDD are owned by me and the Cassandra container is accessed with the same credentials as my interactive shell.
35. As an Operator, I want SETUP.md to explicitly say what's NOT done by bootstrap (formatting the disk, mounting the HDD, installing Node/Docker/cqlsh), so that I know which steps still require me to act manually.

### Files added in PR4

- `data-pipeline/scripts/bootstrap.sh` — one-time idempotent setup. ~80–120 lines of bash.
- `data-pipeline/scripts/run-backfill.sh` — backfill cron entry point. ~30 lines.
- `data-pipeline/scripts/run-hourly.sh` — hourly cron entry point. ~25 lines.
- `data-pipeline/cron/jobflow.cron.template` — cron schedule with placeholders.
- `data-pipeline/SETUP.md` — runbook for the Operator on a fresh Xubuntu box.

PR4 does not modify any of the four already-shipped Node binaries. The wrapper layer is strictly shell + cron + docs.

### `bootstrap.sh` semantics

Idempotent run-many script. Seven steps, in order:

1. Sanity checks: verify `node` (≥ 20.6), `docker compose`, and `cqlsh` are on `PATH`. Fail loudly with an actionable error if any are missing.
2. Verify `/mnt/hdd` is a real mount point via `mountpoint -q`. If not, exit with a reference to SETUP.md.
3. Create directories: `mkdir -p /mnt/hdd/gharchive/` and `mkdir -p ~/jobflow-logs/`. Idempotent.
4. Bring up Cassandra: `docker compose -f data-pipeline/docker-compose.yml up -d`. Wait for the container's healthcheck to pass by polling `cqlsh -e "SELECT release_version FROM system.local"` with exponential backoff.
5. Apply schemas: run `cqlsh -f <schema>` against each file in `data-pipeline/schema/` in numeric order (001 through 008). All `CREATE` statements use `IF NOT EXISTS`; `ALTER ... ADD` is safe to re-run because Cassandra rejects duplicate-column adds gracefully (verified at script time).
6. Install the cron file: `sed "s|@JOBFLOW_ROOT@|<repo absolute path>|g; s|@USER@|<whoami>|g" cron/jobflow.cron.template > /tmp/jobflow.cron && sudo cp /tmp/jobflow.cron /etc/cron.d/jobflow && sudo chmod 0644 /etc/cron.d/jobflow`. Only step that uses `sudo`.
7. Print a summary of what was done plus verification commands.

### `run-backfill.sh` semantics

```
1. set -euo pipefail
2. AVAILABLE_GB_HDD = df --output=avail -BG /mnt/hdd | tail -1 | tr -dc '0-9'
3. if AVAILABLE_GB_HDD < 50 → echo "FATAL: less than 50 GB free on /mnt/hdd" >&2 ; exit 1
4. cd to repo root via $(dirname "$0")/../..
5. node data-pipeline/fetcher/fetcher.js --range <BACKFILL_START_DATE> <yesterday-UTC>
6. node data-pipeline/orchestrators/backfill.js
```

`BACKFILL_START_DATE` is the earliest date the user wants to fetch; for v1 hardcoded into the script as `2025-05-22` (matches the plan §6.4 example). Yesterday's date computed via `date -d 'yesterday' -u +%Y-%m-%d`. The Fetcher is idempotent — skips files already on disk — so re-running this section is cheap.

### `run-hourly.sh` semantics

```
1. set -euo pipefail
2. AVAILABLE_GB_HDD = df ...
3. if AVAILABLE_GB_HDD < 10 → FATAL exit 1
4. cd to repo root
5. node data-pipeline/fetcher/fetcher.js --catchup
6. node data-pipeline/orchestrators/hourly.js
```

### `jobflow.cron.template` semantics

```cron
SHELL=/bin/bash
PATH=/usr/local/bin:/usr/bin:/bin
HOME=/home/@USER@

# Backfill: every night at midnight; held lock for the whole run
0 0 * * * @USER@ flock -n /var/lock/jobflow-backfill.lock @JOBFLOW_ROOT@/data-pipeline/scripts/run-backfill.sh >> /home/@USER@/jobflow-logs/backfill.log 2>&1

# Hourly: every hour at :15; skips if backfill lock is held
15 * * * * @USER@ flock -n /var/lock/jobflow-hourly.lock -c 'flock -n /var/lock/jobflow-backfill.lock true && @JOBFLOW_ROOT@/data-pipeline/scripts/run-hourly.sh || echo "$(date -u +%Y-%m-%dT%H:%M:%SZ) skipping: backfill in progress" >> /home/@USER@/jobflow-logs/hourly-skips.log' >> /home/@USER@/jobflow-logs/hourly.log 2>&1
```

The nested-flock hourly line is the load-bearing coordination per `docs/InjestionPlan.md` §6.1. The outer flock acquires the hourly lock; the inner flock checks the backfill lock without holding it. If either check fails, the hourly skips and logs.

### Log destinations

Three append-forever files under `~/jobflow-logs/`:

- `backfill.log` — every Backfill cron run's combined stdout/stderr.
- `hourly.log` — every Hourly cron run's combined stdout/stderr.
- `hourly-skips.log` — single-line skip entries when the Hourly cron skips due to a held Backfill lock.

No log rotation in v1. The volume is modest (~1 MB per Hourly run, ~50 MB per Backfill run) and the design is "let it grow and revisit if it becomes a problem."

### What PR4 does NOT own

- **The host setup itself** — mounting the HDD at `/mnt/hdd`, installing Node ≥ 20.6, Docker, `cqlsh`, and creating the user account. Documented in SETUP.md as manual prerequisites.
- **The SSD scratch optimisation** — would require modifying the Ingester. Deferred to a future Ingester amendment PR.
- **`seed-companies-from-jobflow.js`** — the bridge from JobFlow's Postgres to the Cassandra `companies` table. Deferred indefinitely.
- **Log rotation, alerting, monitoring dashboards.** Out of scope per the plan §11.
- **Process supervision / systemd unit files.** Cron handles "try again next interval" semantics, which is sufficient for v1.
- **Multi-machine deployment.** Single box.
- **Automated tests.**

### Branch and PR strategy

- Branch: `feature/cassandra-deployment-wrapper` off `main`.
- PR target: `main`.
- Depends on PR3 issues #205 (Hourly Orchestrator) and #206 (Backfill Orchestrator) landing first — `run-hourly.sh` and `run-backfill.sh` directly invoke these binaries.

## Testing Decisions

No automated tests in PR4, consistent with PR1 / PR2 / PR3 decisions.

Manual smoke tests that gate the merge — all run on a fresh Xubuntu VM or the actual Linux box:

- **Fresh bootstrap dry run:** clone the repo, run `bootstrap.sh`. Verify:
  - Cassandra container is healthy.
  - All 8 schemas applied (`DESCRIBE KEYSPACE jobflow` shows expected tables).
  - `~/jobflow-logs/` and `/mnt/hdd/gharchive/` exist.
  - `/etc/cron.d/jobflow` exists with substituted paths.
  - `sudo systemctl status cron` shows cron running.
- **Bootstrap idempotency:** re-run `bootstrap.sh`. Verify it completes without errors and makes no destructive changes.
- **Missing prerequisite test:** temporarily rename `cqlsh` out of `PATH`, run `bootstrap.sh`, verify it fails with a clear error.
- **Missing mount test:** on a box where `/mnt/hdd` is just a regular directory (not a mount), run `bootstrap.sh`, verify it fails referencing SETUP.md.
- **Disk-space guard test:** mock the `df` output (or temporarily fill `/mnt/hdd` to <50 GB free), run `run-backfill.sh` directly, verify it exits non-zero with the FATAL message.
- **Manual cron invocation — backfill:** run `sudo -u <user> flock -n /var/lock/jobflow-backfill.lock /path/to/run-backfill.sh`. Verify the Fetcher and Orchestrator both run, events land in Cassandra, log lines appear in `~/jobflow-logs/backfill.log`.
- **Manual cron invocation — hourly:** same pattern for `run-hourly.sh`.
- **Lock coordination — the load-bearing test:** in one terminal, hold the backfill lock for ten minutes via `flock /var/lock/jobflow-backfill.lock sleep 600`. In another terminal, invoke the hourly cron command verbatim. Verify the hourly skips with the expected message in `hourly-skips.log` and no other side effects.
- **Lock coordination — two backfills:** start a fake backfill via `flock`. Trigger the backfill cron line. Verify the second backfill skips silently.

## Out of Scope

- Modifications to the four already-shipped Node binaries (Fetcher, Ingester, Backfill Orchestrator, Hourly Orchestrator). PR4 wraps them; it does not touch their source.
- The SSD scratch optimisation from `docs/InjestionPlan.md` §5.7. Deferred to a future Ingester amendment.
- `seed-companies-from-jobflow.js`. Deferred indefinitely.
- systemd timer alternatives to cron. Cron is the plan's choice.
- Log rotation (logrotate or any other mechanism).
- Process supervision, restart-on-failure policies.
- Monitoring dashboards (Reaper UI access from MacBook over SSH tunnel is documented in `docs/CassandraPlan.md` §8 but is the Operator's manual setup — not part of PR4).
- Multi-machine deployment, clustering, HA.
- Automated tests of any kind.
- Any UI work.
- The actual one-time disk format / mount step on the Xubuntu box (documented in SETUP.md but performed manually by the Operator).

## Further Notes

- PR4 cannot start implementation until issues #205 (Hourly Orchestrator) and #206 (Backfill Orchestrator) are merged on main. The Hourly and Backfill orchestrators are the immediate downstream consumers of the shell scripts; without them, `run-hourly.sh` and `run-backfill.sh` have nothing to invoke.
- The plan-doc amendments needed by this PR (touchups to `docs/InjestionPlan.md` §6, §7.1, §9 reflecting the actual log directory choice and the per-mode disk-space thresholds) ship inside the implementation PR itself — there is no separate precursor PR for PR4, since no ADRs or CONTEXT.md additions are required.
- Implementation is mechanical bash + cron work. Sonnet (not Opus) is the appropriate model tier when this issue is picked up.
- A future production deployment may want a small follow-up that handles the SSD scratch optimisation (requires Ingester changes), logrotate configuration, and possibly a Prometheus exporter for the partition-write-rate metric. None of those are in scope here.

tomerp20:
## Summary
- Wraps the four already-shipped Node binaries (Fetcher, Ingester, Hourly Orchestrator, Backfill Orchestrator) into a cron-driven production pipeline on a single Xubuntu box.
- Five files only — no changes to the Node binaries: `data-pipeline/scripts/bootstrap.sh`, `data-pipeline/scripts/run-backfill.sh`, `data-pipeline/scripts/run-hourly.sh`, `data-pipeline/cron/jobflow.cron.template`, `data-pipeline/SETUP.md`.
- The load-bearing coordination from `docs/InjestionPlan.md` §6.1 is implemented via a nested `flock` on the hourly cron line, so backfill and hourly cannot race on a shared host.

## Acceptance criteria
(From issue #208 — 35 user stories. Key load-bearing behaviours:)

- [x] **#1, #11** — `bootstrap.sh` is one command, idempotent, safe to re-run.
- [x] **#3** — Fails loudly if `node`, `docker compose`, or `cqlsh` is missing.
- [x] **#4** — Verifies `/mnt/hdd` is a real mount point via `mountpoint -q` before proceeding.
- [x] **#5** — Creates `/mnt/hdd/gharchive/` and `~/jobflow-logs/` if missing.
- [x] **#6** — Brings up Cassandra via `docker compose up -d cassandra` and waits for cqlsh to respond (5 min cap, exponential backoff).
- [x] **#7, #8** — Applies every schema file in numeric order; tolerates "already exists" / "conflicts with an existing column" so re-runs are safe.
- [x] **#9, #28, #29** — Installs the cron file to `/etc/cron.d/jobflow` with `@JOBFLOW_ROOT@` and `@USER@` substituted via `sed`; template stays generic in the repo.
- [x] **#10, #33** — Prints a summary with verification commands at the end.
- [x] **#12, #13** — Backfill cron fires at 00:00 nightly; hourly cron fires at :15 every hour.
- [x] **#14, #15** — Hourly skips cleanly when the backfill lock is held; skip line is appended to `~/jobflow-logs/hourly-skips.log`.
- [x] **#16** — Backfill cron uses `flock -n` so a stacked second backfill silently skips.
- [x] **#17, #18** — `run-backfill.sh` requires 50 GB free on `/mnt/hdd`; `run-hourly.sh` requires 10 GB; both exit non-zero with `FATAL:` if below threshold.
- [x] **#19, #20** — `run-backfill.sh` invokes Fetcher (with `--range BACKFILL_START_DATE yesterday-UTC`) then Backfill Orchestrator. `run-hourly.sh` invokes Fetcher (`--catchup`) then Hourly Orchestrator.
- [x] **#21, #22** — All stdout/stderr is appended to `~/jobflow-logs/{backfill,hourly,hourly-skips}.log` via cron-line redirects; shell scripts know nothing about log paths so direct invocation prints to the terminal.
- [x] **#23, #30** — Scripts resolve paths via `$(dirname "$0")/..` and use `set -euo pipefail`.
- [x] **#24, #34** — Cron jobs run as `@USER@` (the Operator's `whoami` at bootstrap time), so `~/jobflow-logs/` and all on-disk artefacts are owned by that user.
- [x] **#25, #26** — Cron file lives at `/etc/cron.d/jobflow`; `sudo` is used only by `bootstrap.sh` for the `cp` into `/etc/cron.d/`.
- [x] **#27, #35** — `data-pipeline/SETUP.md` is the operator runbook; it explicitly lists what bootstrap does NOT do (mount, install Node/Docker/cqlsh, create the user).
- [x] **#31** — Failed cron runs exit non-zero with `FATAL:` lines visible in the log files.
- [x] **#32** — Bootstrap detects an already-running `jf-cassandra` container and skips `docker compose up`.

## Notes / deviations from the PRD

- The PRD's user-story #34 says scripts should source per-machine config, but the PRD's "semantics" sections don't address how the Node binaries get their env vars under cron. I added a single optional `data-pipeline/.env` that the shell scripts source if present; SETUP.md documents this. This keeps the cron line free of secrets and matches the existing per-binary `.env.example` pattern. If reviewers prefer a different env-loading strategy (per-binary, cron-line vars, etc.), happy to switch.
- The schema-idempotency claim in the PRD ("Cassandra rejects duplicate-column adds gracefully") is not strictly true — duplicate ALTER ADD raises `InvalidRequest`. Bootstrap tolerates that specific error and re-raises anything else, which preserves the intended idempotency.

## Related
Closes #208

### DIFF
--- a/data-pipeline/SETUP.md
+++ b/data-pipeline/SETUP.md
@@ -29,6 +29,12 @@ schema migrations).
 - **Creating the Operator user account.** Bootstrap runs as whoever invokes it.
 - **Log rotation, monitoring, alerting.** Out of scope for v1.
 
+> **Single-tenant assumption.** The locks live in world-writable `/var/lock/`. The
+> design assumes one operator user on the box. If you ever add another local user,
+> they could pre-create `/var/lock/jobflow-{backfill,hourly}.lock` and starve the
+> pipeline (it would skip silently). Move locks to a private path (e.g.
+> `/run/lock/<user>/`) if you violate this assumption.
+
 ---
 
 ## Step 1 — Manual prerequisites (one-time, per box)
@@ -81,10 +87,17 @@ docker compose version    # confirms the plugin is installed
 Used by `bootstrap.sh` to wait for Cassandra and apply schema migrations.
 
 ```bash
-sudo apt-get install -y python3-pip
-pip3 install cqlsh
+# Recent Ubuntu (22.04+) marks the system Python as PEP 668 "externally managed",
+# which makes a plain `pip3 install cqlsh` fail. Use pipx for an isolated install:
+sudo apt-get install -y pipx
+pipx ensurepath          # adds ~/.local/bin to PATH (re-login or `source ~/.profile` after)
+pipx install cqlsh
 cqlsh --version
 ```
+
+> If you prefer not to use `pipx`, the Cassandra Debian repo ships `cqlsh` inside the
+> `cassandra-tools` package — but installing that pulls in the Cassandra server too.
+> `pipx` is the minimal-footprint option.
 
 ### 1.5 Clone the repo
 
--- a/data-pipeline/SETUP.md
+++ b/data-pipeline/SETUP.md
@@ -1,0 +1,231 @@
+# JobFlow Data-Pipeline — Operator Setup Runbook
+
+This runbook brings a fresh Xubuntu box up to a self-running JobFlow data pipeline.
+The pipeline consists of four Node binaries (Fetcher, Ingester, Hourly Orchestrator,
+Backfill Orchestrator) wrapped by `bootstrap.sh`, cron, and `flock` lock coordination.
+
+`bootstrap.sh` is idempotent — safe to re-run any time (e.g. after pulling new
+schema migrations).
+
+---
+
+## What this runbook covers
+
+- Manual prerequisites the Operator must do once per box (mounting the HDD,
+  installing Node / Docker / `cqlsh`).
+- Cloning the repo and per-machine config.
+- Running `bootstrap.sh` to wire up Cassandra, schemas, log directories,
+  and the cron schedule.
+- Verifying everything came up correctly.
+
+## What this runbook does NOT cover
+
+- **Formatting the HDD or mounting it at `/mnt/hdd`** — done by the Operator before
+  `bootstrap.sh` runs. `bootstrap.sh` verifies the mount point exists but does not
+  create it.
+- **Installing Node ≥ 20.6, Docker (with the `docker compose` plugin), or `cqlsh`** —
+  installed via the OS package manager. `bootstrap.sh` checks they are on `PATH`
+  but does not install them.
+- **Creating the Operator user account.** Bootstrap runs as whoever invokes it.
+- **Log rotation, monitoring, alerting.** Out of scope for v1.
+
+---
+
+## Step 1 — Manual prerequisites (one-time, per box)
+
+### 1.1 Format and mount the HDD at `/mnt/hdd`
+
+The Fetcher writes GH Archive `.json.gz` files to `/mnt/hdd/gharchive/`. This must
+be the spinning HDD, not the SSD. Whatever filesystem you pick, ensure it is
+mounted at `/mnt/hdd` and the mount survives reboot (`/etc/fstab` entry).
+
+Verify with:
+
+```bash
+mountpoint /mnt/hdd
+```
+
+If the command prints `/mnt/hdd is a mountpoint`, you are good. If it says
+`is not a mountpoint`, `bootstrap.sh` will refuse to run.
+
+### 1.2 Install Node ≥ 20.6
+
+The orchestrators and fetcher require `node` 20.6 or newer (for stable
+`fetch`, `node:test`, and ESM-by-default).
+
+```bash
+# Example using nvm:
+curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.7/install.sh | bash
+nvm install 20
+nvm use 20
+node --version    # → v20.x.x
+```
+
+### 1.3 Install Docker and the `docker compose` plugin
+
+Cassandra runs in a container managed by `data-pipeline/docker-compose.yml`.
+
+```bash
+# Standard Docker install:
+sudo apt-get update
+sudo apt-get install -y docker.io docker-compose-plugin
+
+# Give your user permission to talk to the docker socket (log out and back in after):
+sudo usermod -aG docker "$USER"
+
+docker compose version    # confirms the plugin is installed
+```
+
+### 1.4 Install `cqlsh`
+
+Used by `bootstrap.sh` to wait for Cassandra and apply schema migrations.
+
+```bash
+sudo apt-get install -y python3-pip
+pip3 install cqlsh
+cqlsh --version
+```
+
+### 1.5 Clone the repo
+
+```bash
+git clone https://github.com/tomerp20/jobflow.git
+cd jobflow
+```
+
+The data-pipeline lives at `data-pipeline/`. The remainder of this guide assumes
+your shell's working directory is the repo root.
+
+### 1.6 Install Node dependencies
+
+```bash
+( cd data-pipeline/fetcher       && npm install --omit=dev )
+( cd data-pipeline/ingester      && npm install --omit=dev )
+( cd data-pipeline/orchestrators && npm install --omit=dev )
+```
+
+### 1.7 Create a per-machine `.env` (optional but recommended)
+
+The cron-fired shell scripts source `data-pipeline/.env` if it exists. Use it to
+set Cassandra contact points, `GHARCHIVE_DIR`, etc. Without it, the Node binaries
+will read from the process environment only.
+
+```bash
+cat > data-pipeline/.env <<'EOF'
+CASSANDRA_CONTACT_POINTS=127.0.0.1
+CASSANDRA_LOCAL_DC=datacenter1
+CASSANDRA_KEYSPACE=jobflow
+GHARCHIVE_DIR=/mnt/hdd/gharchive
+LOG_LEVEL=info
+NODE_ENV=production
+EOF
+```
+
+Per-binary `.env.example` files live under `data-pipeline/{fetcher,ingester,orchestrators}/`
+for reference. The single `data-pipeline/.env` above is the one the cron scripts read.
+
+---
+
+## Step 2 — Run `bootstrap.sh`
+
+```bash
+data-pipeline/scripts/bootstrap.sh
+```
+
+What it does:
+
+1. Verifies `node` (≥ 20.6), `docker compose`, and `cqlsh` are on `PATH`.
+2. Verifies `/mnt/hdd` is a real mount point (not just a directory).
+3. Creates `/mnt/hdd/gharchive/` and `~/jobflow-logs/`.
+4. Brings up the Cassandra container via `docker compose up -d cassandra`
+   and polls `cqlsh` until it responds (up to 5 minutes).
+5. Applies every `.cql` file under `data-pipeline/schema/` in numeric order.
+   Tolerates "already exists" errors so re-runs are safe.
+6. Substitutes `@JOBFLOW_ROOT@` and `@USER@` in `cron/jobflow.cron.template` and
+   installs the result at `/etc/cron.d/jobflow` (this is the **only** step that
+   uses `sudo`).
+7. Prints a summary plus verification commands.
+
+If any step fails, the script exits non-zero with a `FATAL:` line on stderr.
+
+---
+
+## Step 3 — Verify the install
+
+```bash
+# Cron file installed with paths substituted:
+sudo cat /etc/cron.d/jobflow
+grep -E 'run-(backfill|hourly)\.sh' /etc/cron.d/jobflow   # should print both lines
+
+# Cron daemon running:
+sudo systemctl status cron
+
+# Cassandra keyspace and tables in place:
+cqlsh -e "DESCRIBE KEYSPACE jobflow"
+
+# Disk mount and log directory:
+mountpoint /mnt/hdd
+ls -la ~/jobflow-logs/
+
+# Tail logs once the first cron interval has fired:
+tail -F ~/jobflow-logs/hourly.log
+tail -F ~/jobflow-logs/backfill.log
+tail -F ~/jobflow-logs/hourly-skips.log
+```
+
+---
+
+## Day-to-day operation
+
+After bootstrap, the pipeline runs itself. Cron fires:
+
+- **Backfill** — every night at 00:00. Holds
+  `/var/lock/jobflow-backfill.lock` for the whole run. Logs to
+  `~/jobflow-logs/backfill.log`.
+- **Hourly** — every hour at :15. Holds
+  `/var/lock/jobflow-hourly.lock`, and checks the backfill lock; if backfill is
+  running, this hour is skipped and the skip is logged to
+  `~/jobflow-logs/hourly-skips.log`. The Ingester's `--catchup` mode handles the
+  resulting backlog on the next successful hourly invocation. Logs to
+  `~/jobflow-logs/hourly.log`.
+
+If a cron interval skips because the disk is below the safety threshold (50 GB
+free for backfill, 10 GB for hourly), the FATAL message is in the corresponding
+log file. The next interval tries again.
+
+### Manual invocation (testing)
+
+You can run either entry-point script directly to see output on the terminal:
+
+```bash
+data-pipeline/scripts/run-hourly.sh
+data-pipeline/scripts/run-backfill.sh
+```
+
+These scripts know nothing about log paths — that's the cron line's responsibility —
+so direct invocation prints to stdout/stderr as normal.
+
+### Re-running `bootstrap.sh`
+
+Re-running `bootstrap.sh` is safe. Use it to:
+
+- Apply newly-added schema migrations after pulling a new repo state.
+- Re-install the cron file after editing `cron/jobflow.cron.template`.
+- Recover from a partial setup that failed midway.
+
+---
+
+## Files
+
+| File | Purpose |
+|------|---------|
+| `data-pipeline/scripts/bootstrap.sh` | One-time idempotent setup. |
+| `data-pipeline/scripts/run-backfill.sh` | Backfill cron entry point. |
+| `data-pipeline/scripts/run-hourly.sh` | Hourly cron entry point. |
+| `data-pipeline/cron/jobflow.cron.template` | Cron schedule with placeholders. |
+| `data-pipeline/schema/*.cql` | Cassandra schema migrations applied by bootstrap. |
+| `/etc/cron.d/jobflow` | Installed (not committed) — substituted output of the template. |
+| `~/jobflow-logs/` | Append-forever logs (`backfill.log`, `hourly.log`, `hourly-skips.log`). |
+| `/var/lock/jobflow-backfill.lock` | Backfill cron lock. |
+| `/var/lock/jobflow-hourly.lock` | Hourly cron lock. |
+| `/mnt/hdd/gharchive/` | GH Archive `.json.gz` files written by the Fetcher. |
--- a/data-pipeline/cron/jobflow.cron.template
+++ b/data-pipeline/cron/jobflow.cron.template
@@ -1,0 +1,19 @@
+# JobFlow data-pipeline cron schedule.
+#
+# Installed at /etc/cron.d/jobflow by data-pipeline/scripts/bootstrap.sh.
+# Placeholders @JOBFLOW_ROOT@ and @USER@ are substituted at install time.
+#
+# Lock coordination (docs/InjestionPlan.md §6.1):
+#   • Backfill holds /var/lock/jobflow-backfill.lock for the whole run.
+#   • Hourly takes its own lock AND checks the backfill lock; if backfill is
+#     running, the hourly skips this interval (logged to hourly-skips.log).
+
+SHELL=/bin/bash
+PATH=/usr/local/bin:/usr/bin:/bin
+HOME=/home/@USER@
+
+# Backfill — nightly at 00:00. flock -n means stacked backfills skip silently.
+0 0 * * * @USER@ flock -n /var/lock/jobflow-backfill.lock @JOBFLOW_ROOT@/data-pipeline/scripts/run-backfill.sh >> /home/@USER@/jobflow-logs/backfill.log 2>&1
+
+# Hourly — every hour at :15. Nested flock: outer = own lock, inner = check backfill lock without holding it.
+15 * * * * @USER@ flock -n /var/lock/jobflow-hourly.lock -c 'flock -n /var/lock/jobflow-backfill.lock true && @JOBFLOW_ROOT@/data-pipeline/scripts/run-hourly.sh || echo "$(date -u +\%Y-\%m-\%dT\%H:\%M:\%SZ) skipping: backfill in progress" >> /home/@USER@/jobflow-logs/hourly-skips.log' >> /home/@USER@/jobflow-logs/hourly.log 2>&1
--- a/data-pipeline/cron/jobflow.cron.template
+++ b/data-pipeline/cron/jobflow.cron.template
@@ -1,19 +1,24 @@
 # JobFlow data-pipeline cron schedule.
 #
 # Installed at /etc/cron.d/jobflow by data-pipeline/scripts/bootstrap.sh.
-# Placeholders @JOBFLOW_ROOT@ and @USER@ are substituted at install time.
+# Placeholders @JOBFLOW_ROOT@, @USER@, @HOME@, @NODE_BIN_DIR@ are substituted
+# at install time.
 #
 # Lock coordination (docs/InjestionPlan.md §6.1):
 #   • Backfill holds /var/lock/jobflow-backfill.lock for the whole run.
 #   • Hourly takes its own lock AND checks the backfill lock; if backfill is
 #     running, the hourly skips this interval (logged to hourly-skips.log).
+#     The skip path is distinct from the failure path: a genuine run-hourly.sh
+#     failure propagates a non-zero exit to cron rather than being recorded
+#     as a skip.
 
 SHELL=/bin/bash
-PATH=/usr/local/bin:/usr/bin:/bin
-HOME=/home/@USER@
+PATH=@NODE_BIN_DIR@:/usr/local/bin:/usr/bin:/bin
+HOME=@HOME@
 
 # Backfill — nightly at 00:00. flock -n means stacked backfills skip silently.
-0 0 * * * @USER@ flock -n /var/lock/jobflow-backfill.lock @JOBFLOW_ROOT@/data-pipeline/scripts/run-backfill.sh >> /home/@USER@/jobflow-logs/backfill.log 2>&1
+0 0 * * * @USER@ flock -n /var/lock/jobflow-backfill.lock @JOBFLOW_ROOT@/data-pipeline/scripts/run-backfill.sh >> @HOME@/jobflow-logs/backfill.log 2>&1
 
 # Hourly — every hour at :15. Nested flock: outer = own lock, inner = check backfill lock without holding it.
-15 * * * * @USER@ flock -n /var/lock/jobflow-hourly.lock -c 'flock -n /var/lock/jobflow-backfill.lock true && @JOBFLOW_ROOT@/data-pipeline/scripts/run-hourly.sh || echo "$(date -u +\%Y-\%m-\%dT\%H:\%M:\%SZ) skipping: backfill in progress" >> /home/@USER@/jobflow-logs/hourly-skips.log' >> /home/@USER@/jobflow-logs/hourly.log 2>&1
+# Branching `if` keeps the skip path (false branch) distinct from a genuine run-hourly.sh failure (true branch's exit code propagates).
+15 * * * * @USER@ flock -n /var/lock/jobflow-hourly.lock -c 'if flock -n /var/lock/jobflow-backfill.lock true; then @JOBFLOW_ROOT@/data-pipeline/scripts/run-hourly.sh; else echo "$(date -u +\%Y-\%m-\%dT\%H:\%M:\%SZ) skipping: backfill in progress" >> @HOME@/jobflow-logs/hourly-skips.log; fi' >> @HOME@/jobflow-logs/hourly.log 2>&1
--- a/data-pipeline/scripts/bootstrap.sh
+++ b/data-pipeline/scripts/bootstrap.sh
@@ -1,0 +1,125 @@
+#!/usr/bin/env bash
+#
+# JobFlow data-pipeline bootstrap.
+#
+# Idempotent one-time setup for a fresh Xubuntu box. Safe to re-run.
+# See data-pipeline/SETUP.md for the prerequisite checklist this script does NOT cover.
+
+set -euo pipefail
+
+# Resolve repo paths from this script's location.
+SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)"
+DATA_PIPELINE_ROOT="$(cd -- "${SCRIPT_DIR}/.." && pwd)"
+REPO_ROOT="$(cd -- "${DATA_PIPELINE_ROOT}/.." && pwd)"
+
+HDD_MOUNT="/mnt/hdd"
+GHARCHIVE_DIR="${HDD_MOUNT}/gharchive"
+LOG_DIR="${HOME}/jobflow-logs"
+CRON_DEST="/etc/cron.d/jobflow"
+CRON_TEMPLATE="${DATA_PIPELINE_ROOT}/cron/jobflow.cron.template"
+COMPOSE_FILE="${DATA_PIPELINE_ROOT}/docker-compose.yml"
+SCHEMA_DIR="${DATA_PIPELINE_ROOT}/schema"
+
+OPERATOR_USER="$(whoami)"
+
+log()  { printf '[bootstrap] %s\n' "$*"; }
+fail() { printf '[bootstrap] FATAL: %s\n' "$*" >&2; exit 1; }
+
+# ── 1. Prerequisite checks ───────────────────────────────────────────────────
+log "Step 1: checking prerequisites"
+
+command -v node >/dev/null 2>&1 || fail "node is not on PATH (need >= 20.6). See data-pipeline/SETUP.md"
+NODE_MAJOR="$(node -p 'process.versions.node.split(".")[0]')"
+NODE_MINOR="$(node -p 'process.versions.node.split(".")[1]')"
+if [ "${NODE_MAJOR}" -lt 20 ] || { [ "${NODE_MAJOR}" -eq 20 ] && [ "${NODE_MINOR}" -lt 6 ]; }; then
+  fail "node $(node -v) is too old (need >= 20.6). See data-pipeline/SETUP.md"
+fi
+
+docker compose version >/dev/null 2>&1 || fail "'docker compose' plugin is missing. See data-pipeline/SETUP.md"
+command -v cqlsh >/dev/null 2>&1 || fail "cqlsh is not on PATH. See data-pipeline/SETUP.md"
+command -v sudo  >/dev/null 2>&1 || fail "sudo is not on PATH (needed for cron install)"
+
+# ── 2. Mount-point check ─────────────────────────────────────────────────────
+log "Step 2: verifying ${HDD_MOUNT} is a real mount point"
+mountpoint -q "${HDD_MOUNT}" || fail "${HDD_MOUNT} is not a mount point. See data-pipeline/SETUP.md"
+
+# ── 3. Directory creation ────────────────────────────────────────────────────
+log "Step 3: creating runtime directories"
+mkdir -p "${GHARCHIVE_DIR}"
+mkdir -p "${LOG_DIR}"
+
+# ── 4. Cassandra container ───────────────────────────────────────────────────
+log "Step 4: bringing up Cassandra via docker compose"
+if docker ps --filter "name=jf-cassandra" --filter "status=running" --format '{{.Names}}' | grep -q '^jf-cassandra$'; then
+  log "  jf-cassandra container already running; skipping 'docker compose up'"
+else
+  docker compose -f "${COMPOSE_FILE}" up -d cassandra
+fi
+
+log "  waiting for Cassandra to accept CQL connections..."
+DELAY=2
+MAX_DELAY=30
+DEADLINE=$(( $(date +%s) + 300 ))  # 5 minutes
+until cqlsh -e "SELECT release_version FROM system.local" >/dev/null 2>&1; do
+  if [ "$(date +%s)" -ge "${DEADLINE}" ]; then
+    fail "Cassandra did not become reachable on 127.0.0.1:9042 within 5 minutes"
+  fi
+  sleep "${DELAY}"
+  if [ "${DELAY}" -lt "${MAX_DELAY}" ]; then
+    DELAY=$(( DELAY * 2 ))
+    [ "${DELAY}" -gt "${MAX_DELAY}" ] && DELAY="${MAX_DELAY}"
+  fi
+done
+log "  Cassandra is reachable"
+
+# ── 5. Schema migrations ─────────────────────────────────────────────────────
+log "Step 5: applying schema migrations in numeric order"
+for schema_file in $(find "${SCHEMA_DIR}" -maxdepth 1 -name '*.cql' | sort); do
+  schema_name="$(basename "${schema_file}")"
+  log "  applying ${schema_name}"
+  # ALTER TABLE ADD raises InvalidRequest("conflicts with an existing column") when re-run.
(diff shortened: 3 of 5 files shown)


data-pipeline/cron/jobflow.cron.template
```diff
@@ -0,0 +1,19 @@
...
+#     running, the hourly skips this interval (logged to hourly-skips.log).
+
+SHELL=/bin/bash
+PATH=/usr/local/bin:/usr/bin:/bin
+HOME=/home/@USER@
+
+# Backfill — nightly at 00:00. flock -n means stacked backfills skip silently.
+0 0 * * * @USER@ flock -n /var/lock/jobflow-backfill.lock @JOBFLOW_ROOT@/data-pipeline/scripts/run-backfill.sh >> /home/@USER@/jobflow-logs/backfill.log 2>&1
+
+# Hourly — every hour at :15. Nested flock: outer = own lock, inner = check backfill lock without holding it.
```
tomerp20, line 22: **HIGH — Skip-log misreports legitimate hourly failures as backfill skips.**

The line uses `A && B || C` precedence: if the inner `flock ... true` succeeds (backfill lock free) and then `run-hourly.sh` (B) **fails** (e.g. disk-space FATAL, fetcher crash, orchestrator throws), bash falls through to the `||` branch and writes "skipping: backfill in progress" to `hourly-skips.log`. That's a false statement — backfill wasn't running, hourly just crashed.

Worse, because `echo` succeeds, the cron line exits 0, so the failure is invisible to anyone monitoring cron exit status. The only trace is the stderr in `hourly.log` and a misleading line in `hourly-skips.log`.

Suggested fix — use an explicit `if` so the skip branch only fires for the intended case:

```
15 * * * * @USER@ flock -n /var/lock/jobflow-hourly.lock -c 'if flock -n /var/lock/jobflow-backfill.lock true; then @JOBFLOW_ROOT@/data-pipeline/scripts/run-hourly.sh; else echo "$(date -u +\%Y-\%m-\%dT\%H:\%M:\%SZ) skipping: backfill in progress" >> /home/@USER@/jobflow-logs/hourly-skips.log; fi' >> /home/@USER@/jobflow-logs/hourly.log 2>&1
```

This also propagates run-hourly.sh's exit code so cron records the failure.

data-pipeline/cron/jobflow.cron.template
```diff
@@ -0,0 +1,19 @@
...
+#
+# Installed at /etc/cron.d/jobflow by data-pipeline/scripts/bootstrap.sh.
+# Placeholders @JOBFLOW_ROOT@ and @USER@ are substituted at install time.
+#
+# Lock coordination (docs/InjestionPlan.md §6.1):
+#   • Backfill holds /var/lock/jobflow-backfill.lock for the whole run.
+#   • Hourly takes its own lock AND checks the backfill lock; if backfill is
+#     running, the hourly skips this interval (logged to hourly-skips.log).
+
+SHELL=/bin/bash
```
tomerp20, line 15: **MEDIUM — `PATH` does not include nvm-installed node.**

SETUP.md §1.2 explicitly recommends installing Node via nvm, which lands the binary at `~/.nvm/versions/node/v20.x.x/bin/node` — not on the cron `PATH=/usr/local/bin:/usr/bin:/bin` set on line 12.

Under cron, `run-backfill.sh` and `run-hourly.sh` will both fail at `node data-pipeline/fetcher/fetcher.js …` with `command not found`. The disk-space guard will pass, then `set -e` aborts with a confusing error and the cron run fails — but only after the user has bootstrapped successfully (because bootstrap.sh runs in the operator's interactive shell where nvm is sourced).

Either:
- Document in SETUP.md that the user must install node via apt / NodeSource so `/usr/bin/node` exists, or
- Append a node-aware path to the cron template (e.g. `PATH=/home/@USER@/.nvm/versions/node/current/bin:/usr/local/bin:/usr/bin:/bin` plus instructions to symlink `current`), or
- Have the run-* scripts source `~/.nvm/nvm.sh` if present before invoking `node`.

data-pipeline/cron/jobflow.cron.template
```diff
@@ -0,0 +1,19 @@
...
+PATH=/usr/local/bin:/usr/bin:/bin
+HOME=/home/@USER@
```
tomerp20: **MEDIUM — `HOME=/home/@USER@` and log-path interpolation assume a fixed home layout.**

The bootstrap derives `LOG_DIR="${HOME}/jobflow-logs"` correctly from the operator's actual home directory, but the cron template hardcodes `HOME=/home/@USER@` and writes logs to `/home/@USER@/jobflow-logs/…`. If the operator's home is anything other than `/home/$USER` (custom layout, `/var/lib/<user>`, a system account), the two will diverge — `bootstrap.sh` creates the log dir in the real `$HOME` but cron writes to a path that doesn't exist, so logging silently fails and `hourly-skips.log` is never written.

Fix: substitute a third placeholder `@HOME@` resolved via `$(getent passwd "$OPERATOR_USER" | cut -d: -f6)` (or simply "$HOME") at bootstrap time, and replace `/home/@USER@` everywhere on lines 13, 17, 18.

data-pipeline/scripts/bootstrap.sh
```diff
@@ -0,0 +1,125 @@
...
+for schema_file in $(find "${SCHEMA_DIR}" -maxdepth 1 -name '*.cql' | sort); do
+  schema_name="$(basename "${schema_file}")"
+  log "  applying ${schema_name}"
+  # ALTER TABLE ADD raises InvalidRequest("conflicts with an existing column") when re-run.
+  # Tolerate that specific case to keep bootstrap idempotent; abort on anything else.
+  if ! out="$(cqlsh -f "${schema_file}" 2>&1)"; then
+    if printf '%s' "${out}" | grep -Eq 'conflicts with an existing column|already exists'; then
+      log "    (already applied, skipping)"
+    else
+      printf '%s\n' "${out}" >&2
```
tomerp20, line 93: **MEDIUM — Idempotency `grep` can mask real schema errors.**

The current logic checks the combined `cqlsh` stdout+stderr for either `conflicts with an existing column` or `already exists`. Two concerns:

1. **Multi-statement files leak through.** `.cql` files can contain multiple statements. `cqlsh -f` runs them sequentially. If statement 1 fails with a legitimate, unrelated error (typo, bad keyspace, schema disagreement) AND statement 2 raises a tolerated "already exists" — the regex matches and the entire failure is swallowed.
2. **Pattern is too loose.** `already exists` is substring-matched anywhere; any future Cassandra diagnostic that incidentally contains those two words will be silently ignored.

Safer approach options:
- Make each CQL statement explicitly use `IF NOT EXISTS` (works for CREATE; ALTER ADD must remain tolerated separately) and only tolerate the exact `InvalidRequest.*conflicts with an existing column` message verbatim.
- Or parse cqlsh's exit code per-statement by splitting the file (heavier but precise).
- Minimum: tighten the regex to require Cassandra's full error frame, e.g. `InvalidRequest: Error from server.*(conflicts with an existing column|already exists)`, so stray substring matches don't silence errors.

data-pipeline/scripts/bootstrap.sh
```diff
@@ -0,0 +1,125 @@
...
+  if [ "${DELAY}" -lt "${MAX_DELAY}" ]; then
+    DELAY=$(( DELAY * 2 ))
+    [ "${DELAY}" -gt "${MAX_DELAY}" ] && DELAY="${MAX_DELAY}"
+  fi
+done
+log "  Cassandra is reachable"
+
+# ── 5. Schema migrations ─────────────────────────────────────────────────────
+log "Step 5: applying schema migrations in numeric order"
+for schema_file in $(find "${SCHEMA_DIR}" -maxdepth 1 -name '*.cql' | sort); do
```
tomerp20: **LOW — `for f in $(find …)` is whitespace-fragile and ignores word-splitting hygiene.**

`for schema_file in $(find "${SCHEMA_DIR}" -maxdepth 1 -name '*.cql' | sort)` will mis-split filenames containing spaces, tabs, or globs. The current `schema/*.cql` files don't, but the pattern violates `shellcheck` SC2044 and is fragile against future files.

Simpler and safer with a sorted glob:

```
shopt -s nullglob
for schema_file in "${SCHEMA_DIR}"/*.cql; do
  …
done
```

The glob already produces sorted output in bash, and `nullglob` makes "no files" produce an empty loop instead of a literal `*.cql`.

data-pipeline/SETUP.md
```diff
@@ -0,0 +1,231 @@
...
+
+```bash
+sudo apt-get install -y python3-pip
+pip3 install cqlsh
+cqlsh --version
+```
+
+### 1.5 Clone the repo
+
+```bash
```
tomerp20, line 104: **MEDIUM — `pip3 install cqlsh` will fail on modern Ubuntu (PEP 668) and isn't the recommended source.**

Recent Ubuntu / Debian (22.04+) ship a PEP 668 "externally-managed-environment" marker that makes `pip3 install cqlsh` exit with an error unless the operator passes `--break-system-packages` or uses a venv / pipx. Xubuntu inherits this. An operator following this runbook verbatim will hit a stop-the-line error.

Two safer routes:

- `sudo apt-get install -y cassandra-tools` (installs the bundled cqlsh from the Cassandra package archive); or
- `pipx install cqlsh` (with a note to install pipx first).

Worth picking one and updating the runbook.

data-pipeline/scripts/bootstrap.sh
```diff
@@ -0,0 +1,125 @@
...
+mountpoint -q "${HDD_MOUNT}" || fail "${HDD_MOUNT} is not a mount point. See data-pipeline/SETUP.md"
+
+# ── 3. Directory creation ────────────────────────────────────────────────────
+log "Step 3: creating runtime directories"
+mkdir -p "${GHARCHIVE_DIR}"
+mkdir -p "${LOG_DIR}"
+
+# ── 4. Cassandra container ───────────────────────────────────────────────────
+log "Step 4: bringing up Cassandra via docker compose"
+if docker ps --filter "name=jf-cassandra" --filter "status=running" --format '{{.Names}}' | grep -q '^jf-cassandra$'; then
```
tomerp20, line 55: **LOW — `docker compose up -d` will recreate an exited (non-running) container; bootstrap claims to skip.**

The detection on line 53 filters `status=running`. That correctly handles the "already healthy" case. But if `jf-cassandra` exists in an `exited` state (e.g. the box rebooted and the container hasn't been auto-restarted), the `grep` finds no match and `docker compose up -d cassandra` runs — fine in practice, but the surrounding doc and PR description ("detects an already-running container and skips") gives the impression that any existing container is left alone. An operator debugging a stuck Cassandra by stopping it and re-running bootstrap may be surprised by the container restart. Either tighten the filter to `status=running|paused|created` plus a clearer log line, or update the PR/SETUP wording to read "detects a running container".

data-pipeline/scripts/run-backfill.sh
```diff
@@ -0,0 +1,36 @@
...
+# Load per-machine env (Cassandra contact points, GHARCHIVE_DIR, etc.) if present.
+if [ -f "${DATA_PIPELINE_ROOT}/.env" ]; then
+  set -a
+  # shellcheck disable=SC1091
+  . "${DATA_PIPELINE_ROOT}/.env"
+  set +a
+fi
+
+# Disk-space guard.
+AVAILABLE_GB_HDD="$(df --output=avail -BG "${HDD_MOUNT}" | tail -1 | tr -dc '0-9')"
```
tomerp20, line 30: **LOW — Disk-space guard: `df --output` is GNU-only and `tr -dc '0-9'` will concatenate digits if `df` ever prints more than one line of data.**

Two minor concerns:

1. `df --output=avail -BG` is a GNU coreutils extension. Fine for Xubuntu (the documented target), but the script silently emits an unhelpful error if someone tries to run it on macOS / BSD for ad-hoc local testing. A one-line comment saying "GNU df only" would save head-scratching.
2. `tail -1 | tr -dc '0-9'` is robust to the trailing `G` suffix, but if `/mnt/hdd` is missing entirely `df` writes the error to stderr and prints nothing on stdout — `AVAILABLE_GB_HDD` becomes empty and the `${AVAILABLE_GB_HDD:-0}` fallback (good) compares 0 < 50 and the script exits with the FATAL message — correct outcome, but the message reads "have 0 GB free" rather than "mount missing". Consider an explicit `[ -d "${HDD_MOUNT}" ] || fail …` ahead of the `df` call so the diagnostic matches reality.

tomerp20:
## Correctness & robustness review — deployment wrapper

Head SHA: `d3fb747a3731551cd9addd05e16e396dbec5f582`

### Findings

| Severity | File | Line | Description |
|----------|------|------|-------------|
| HIGH | `cron/jobflow.cron.template` | 18 | `A && B \|\| C` precedence: a legitimate `run-hourly.sh` failure writes a false "skipping: backfill in progress" entry to `hourly-skips.log` and masks the failure from cron (echo exits 0). |
| MEDIUM | `cron/jobflow.cron.template` | 11 | Cron `PATH` does not include nvm-installed node, yet SETUP.md §1.2 recommends nvm. Cron will fail with `node: command not found` even after a clean bootstrap. |
| MEDIUM | `cron/jobflow.cron.template` | 13 | `HOME=/home/@USER@` and `/home/@USER@/jobflow-logs/...` hardcode the home-dir layout while bootstrap derives `LOG_DIR` from real `$HOME` — diverges for non-standard homes; logs silently disappear. |
| MEDIUM | `scripts/bootstrap.sh` | 86 | Idempotency `grep` for `already exists` / `conflicts with an existing column` is substring-loose and operates on combined output for multi-statement `.cql` files; can mask real errors. |
| MEDIUM | `SETUP.md` | 91 | `pip3 install cqlsh` fails on Ubuntu 22.04+ (PEP 668 externally-managed-environment); recommend `cassandra-tools` or `pipx`. |
| LOW | `scripts/bootstrap.sh` | 53 | "Skip docker compose" path only fires for `status=running`; exited containers still trigger `up -d`, which contradicts PR description wording. |
| LOW | `scripts/bootstrap.sh` | 77 | `for f in $(find ...)` is whitespace-fragile (shellcheck SC2044); prefer a sorted bash glob with `shopt -s nullglob`. |
| LOW | `scripts/run-backfill.sh` | 25 | `df --output=avail -BG` is GNU-only (fine for target, surprising for local testing); missing-mount case reports "have 0 GB" rather than the real cause. Same code lives in `run-hourly.sh`. |

### Potential issues considered but rejected

- **`%` escaping in the hourly cron line.** All five `%` inside `$(date -u +...)` are correctly backslash-escaped (`\%Y-\%m-\%dT\%H:\%M:\%SZ`); cron will pass literal `%` characters to the shell.
- **Race between inner `flock ... true` releasing the backfill lock and `run-hourly.sh` starting.** Theoretically open, but the only window where this matters is when backfill is fired by its own cron line (00:00) — hourly fires at :15. Schedule-level separation means there's no practical race.
- **`set -euo pipefail` + sourced `.env`.** `set -a` / `set +a` around `.` is correct; sourced file is `data-pipeline/.env`, which the operator owns. If `.env` itself errors under `-e`, the script aborts cleanly — that's the intended behaviour.
- **`mktemp` + `trap rm -f` cleanup in bootstrap step 6.** Sound: `mktemp -t` creates the file with `XXXXXX` suffix, trap fires on EXIT including failure paths, `rm -f` is harmless if the file is already gone.
- **`$(dirname "$0")/..` path resolution under cron.** The scripts use `${BASH_SOURCE[0]}` with `cd -- ... && pwd`, which resolves correctly regardless of cron's CWD (cron's default CWD is `$HOME`, but the scripts immediately `cd "${REPO_ROOT}"`).
- **`docker compose up -d cassandra` re-running is destructive.** It isn't — docker compose is a no-op when the desired state matches.
- **Cassandra wait-loop exponential-backoff math.** Caps at `MAX_DELAY=30`, deadline check happens BEFORE sleep, so worst case the script sleeps 30s past the 5-minute deadline. Acceptable.
- **Backfill cron's flock interaction with bootstrap re-runs.** Not an issue — bootstrap doesn't fire cron jobs; it installs the cron file.
- **`set -euo pipefail` interaction with the `cqlsh` capture `if ! out="$(cqlsh ...)"; then`.** Correct: `if ! cmd` is the documented way to handle command failure under `-e`.

### No-concern checklist

- **Bash hygiene** — quoting in bootstrap.sh and the run-* scripts is consistent and shellcheck-clean apart from the `find` SC2044 noted above. `set -euo pipefail` is set everywhere; failure modes propagate correctly.
- **`mktemp` / `trap` cleanup** — sound.
- **Path resolution under cron** — scripts compute their own root from `${BASH_SOURCE[0]}`; cron CWD does not matter.
- **Cron `%` escaping** — all occurrences are escaped.
- **Bootstrap step 1–3 idempotency** — `mountpoint`, `mkdir -p`, version-comparison logic are all safe to re-run.
- **Bootstrap step 6 sed substitution** — `|` delimiter avoids slash collisions; placeholders are unique tokens.
- **Backfill cron line itself** — `flock -n` on a single lock is straightforward and correct.
- **Disk-space threshold boundary** — `-lt 50` correctly excludes exactly 49 GB; comparison is integer-typed via `[ ]`.

tomerp20:
## Security review — orchestrator pass

Reviewed by the orchestrator. Severity scale: HIGH (blocks merge) / MEDIUM (fix before merge) / LOW (defense in depth) / NIT.

### Findings

**LOW — Log directory inherits default umask.**
`bootstrap.sh:49` does `mkdir -p "${LOG_DIR}"` without explicit permissions. Default umask 022 → `~/jobflow-logs/` ends up `755`, and log files written by cron are typically `644`. The Node binaries log at `LOG_LEVEL=info`; in failure paths they will dump Cassandra contact points and query context. If the host ever sprouts a second user account, those logs are world-readable.

Suggested: `chmod 700 "${LOG_DIR}"` immediately after `mkdir -p`. Cheap defense in depth.

**LOW — `/var/lock` lock files are race-creatable by other local users.**
`/var/lock` is typically `1777` (world-writable + sticky bit). A non-root local user could pre-create `/var/lock/jobflow-backfill.lock` and hold it, indefinitely starving the backfill cron (which uses `flock -n` and exits silently on failure). The design assumes a single-operator box, so this is acceptable — but worth documenting in SETUP.md as a single-tenant assumption, or moving locks to `/run/lock/<user>/` if the assumption ever changes.

### No concerns

- **`sudo` usage** — single occurrence in `bootstrap.sh`, scoped to `cp` and `chmod` on `/etc/cron.d/jobflow`. Correct: cron.d files must be `root:root 0644`. No `sudo` in the run-time entry points.
- **Cron file permissions** — `chown root:root` + `chmod 0644` on `/etc/cron.d/jobflow` is the standard. No secrets in the template, just paths and the operator's username.
- **Privilege boundary** — day-to-day operation runs entirely as the operator's user (`@USER@` substituted from `whoami`). Matches the issue's user-story #34.
- **`.env` sourcing pattern** — `set -a; . .env; set +a` is standard; operator-owned file, no privilege boundary crossed.
- **`sed` substitution injection** — `@JOBFLOW_ROOT@` from `$(cd .. && pwd)` and `@USER@` from `$(whoami)` are not attacker-controlled. The `|` sed delimiter avoids the common `/` clash in paths.
- **Shell injection via env vars** — all expansions properly quoted; `set -euo pipefail` everywhere; no `eval`, no `xargs -I` traps.
- **`mktemp` + `trap rm`** — safe tmp-file pattern.

tomerp20:
## Code review — Phase 3 final status

| Phase | Agent | Findings | Fixed | Skipped |
|-------|-------|----------|-------|---------|
| 1 | infrastructure deep-review (general-purpose, opus) | 1 HIGH · 4 MEDIUM · 3 LOW | — | — |
| 1 | security (orchestrator inline) | 2 LOW | — | — |
| 2 | orchestrator (commit `001e4d9`) | — | 8 | 0 |

### Fixed

| Severity | Issue | Fix |
|----------|-------|-----|
| HIGH | Hourly cron's `A && B \|\| C` logged a legitimate `run-hourly.sh` failure as a backfill skip and masked the non-zero exit from cron | Cron line rewritten with explicit `if flock … true; then run-hourly.sh; else echo "skip…"; fi`. Skip path and failure path now distinct; cron exit code reflects reality. |
| MEDIUM | `PATH` lacked the nvm-installed node directory; cron would `node: command not found` even after a clean bootstrap | Added `@NODE_BIN_DIR@` placeholder; `bootstrap.sh` resolves it via `dirname $(command -v node)` and substitutes at install time. |
| MEDIUM | `HOME=/home/@USER@` and `/home/@USER@/jobflow-logs/...` hardcoded the home-dir layout while bootstrap used real `$HOME`; logs could silently disappear | Added `@HOME@` placeholder substituted from the operator's actual `$HOME`. |
| MEDIUM | Idempotency `grep` for `already exists` / `conflicts with an existing column` was substring-loose over combined output; could mask real errors in multi-statement `.cql` files | Now extracts error-bearing lines and treats output as benign only when every such line matches a known-idempotent pattern. |
| MEDIUM | `pip3 install cqlsh` fails on Ubuntu 22.04+ (PEP 668 externally-managed) | SETUP.md updated to use `pipx install cqlsh` with a fallback note on the `cassandra-tools` apt package. |
| LOW | `for f in $(find …)` is whitespace-fragile (SC2044) | Switched to `mapfile -t schema_files < <(find … \| sort)`. |
| LOW | Disk-space guard reported "have 0 GB" when the mount was missing, hiding the real cause | Added `mountpoint -q` check before `df` in both `run-*.sh`. |
| LOW (security) | `~/jobflow-logs/` inherited default umask (755); log files world-readable | Added `chmod 700` after `mkdir -p`. |

### Documented (not code-fixed)

| Severity | Note |
|----------|------|
| LOW (security) | `/var/lock` is world-writable; a non-root local user could race-create the lock files and starve the pipeline. SETUP.md now documents the single-tenant assumption with a pointer to move locks to `/run/lock/<user>/` if it's violated. |

### Skipped intentionally

| Severity | Why |
|----------|-----|
| LOW | `docker compose up -d` re-creating an *exited* container (line 53 of bootstrap.sh) — current behavior matches what an operator actually wants (restart on bootstrap re-run); only the PR description wording slightly over-promised the skip behavior, not the code itself. |

### Phase 2 commit

`001e4d9` — `fix: address code review findings on deployment wrapper (PR #229)`

### Verification

- `bash -n` clean on all three shell scripts.
- Dry-run `sed` substitution against the new cron template with `@NODE_BIN_DIR@`, `@HOME@`, `@USER@`, `@JOBFLOW_ROOT@` produces the expected output.
- No Node binaries touched — PR4 remains a pure shell + cron + docs change.

PR #230, opened 2026-05-23
[Cassandra pipeline followup] Backfill Orchestrator --target-companies encoding is brittle

### LINKED ISSUE #225: [Cassandra pipeline followup] Backfill Orchestrator --target-companies encoding is brittle
## Severity

LOW (today) → MEDIUM (long-term). Not exploitable in v1 because GitHub org/company names follow `[a-zA-Z0-9-]+` and never contain `:` or `,`. Worth fixing before any code path inserts non-canonical names into the `companies` table.

## Context

Surfaced by the Phase 1 review on PR #222 (issue #206, Backfill Orchestrator). The reviewer marked it as a MEDIUM finding skipped from auto-fix because it requires a coordinated change in the Ingester's `--target-companies` parser. Filed as a followup per the user's decision to merge PR #222 with smoke test only.

## The problem

`data-pipeline/orchestrators/backfill.js:128` builds the `--target-companies` CLI argument by joining `(company, org_name)` pairs with `:` as the intra-pair separator and `,` as the inter-pair separator:

```js
const targetCompaniesArg = targetRows
  .map(r => `${r.company}:${r.org_name}`)
  .join(',');
```

The Ingester then parses this format on the receiving side. If a `company` or `org_name` value ever contains `:` or `,`, the parsing is silently wrong and the Ingester writes events for the wrong company.

## Why this is theoretical today

GitHub org names are constrained by GitHub's username rules: `[a-zA-Z0-9-]+`, max 39 chars, can't start or end with `-`. Colons and commas are syntactically impossible. Company names in our `companies` table are populated by the Company Scout (or manually by the operator), which currently follows the same convention.

The brittleness exists if/when any of these become true:
- A future seeding mechanism (e.g. `seed-companies-from-jobflow.js`, deferred in #212) imports company names from a system that doesn't enforce the same character set.
- A user manually inserts a Company row with a name containing punctuation.
- The Company Scout's resolution logic ever encodes metadata into the company-name slot.

## Recommended fix (when revived)

Two clean paths, both require changes in both the Backfill Orchestrator (`backfill.js`) and the Ingester (`lib/cli.js`):

**Option 1 — URL-encode each token.**
```js
// Orchestrator
const targetCompaniesArg = targetRows
  .map(r => `${encodeURIComponent(r.company)}:${encodeURIComponent(r.org_name)}`)
  .join(',');

// Ingester (parseTargetCompanies)
.map(pair => {
  const [c, o] = pair.split(':');
  return { company: decodeURIComponent(c), org: decodeURIComponent(o) };
})
```
Trade-off: percent-encoded values in process listings (`ps`) are ugly to read. Minimal code change.

**Option 2 — Switch to a JSON-file transport.**
- Orchestrator writes a temp file `target-companies.json` with `[{company, org}, ...]`.
- Passes `--target-companies-file <path>` to the Ingester.
- Ingester reads the file at startup.

Trade-off: extra disk I/O and temp-file lifecycle management, but values stay readable in `ps` and there's no encoding ceremony.

**Recommendation: Option 1** for v1. Smaller diff, no new lifecycle concerns, sufficient defense against any future colon/comma in a name.

## Out of scope

- Validating company/org_name format at Cassandra INSERT time. That's a separate concern (data-quality at the source), tracked tangentially in #212 (`seed-companies-from-jobflow.js`).

## Acceptance criteria

- [ ] Orchestrator URL-encodes each component when building `--target-companies`.
- [ ] Ingester URL-decodes each component when parsing.
- [ ] Existing values (`microsoft:microsoft`, `wix:wix-incubator`, etc.) decode identically (URL encoding is a no-op for `[a-zA-Z0-9-]`).
- [ ] Smoke test: insert a Company with name `foo:bar,baz` (deliberately pathological), trigger a backfill via the orchestrator, verify the Ingester correctly parses it as a single `(company='foo:bar,baz', org='...')` pair and writes events to the right partition.

## Blocked by

None. Independent followup. ~10 lines across two files.

tomerp20:
## Summary
- URL-encode each `(company, org_name)` token in the Backfill Orchestrator before joining into `--target-companies`
- URL-decode in the Ingester's `parseTargetCompanies` after splitting
- Defends the `company:org,company:org` wire format against future values containing `:` or `,`

## Acceptance criteria
- [x] Orchestrator URL-encodes each component when building `--target-companies`.
- [x] Ingester URL-decodes each component when parsing.
- [x] Existing values (`microsoft:microsoft`, `wix:wix-incubator`, etc.) decode identically (URL encoding is a no-op for `[a-zA-Z0-9-]`).
- [ ] Smoke test: insert a Company with name `foo:bar,baz` (deliberately pathological), trigger a backfill via the orchestrator, verify the Ingester correctly parses it as a single `(company='foo:bar,baz', org='...')` pair and writes events to the right partition. — _Local round-trip verified via inline node script; full end-to-end smoke test against the orchestrator left for review/manual run._

## Verification
Local round-trip sanity test (encode in orchestrator shape, decode in ingester shape):
```
encoded: microsoft:microsoft,wix:wix-incubator,foo%3Abar%2Cbaz:evil-org
parsed:  [{company:'microsoft',org:'microsoft'},{company:'wix',org:'wix-incubator'},{company:'foo:bar,baz',org:'evil-org'}]
PASS: round-trip identical
```

## Related
Closes #225

### DIFF
--- a/data-pipeline/ingester/lib/cli.js
+++ b/data-pipeline/ingester/lib/cli.js
@@ -27,9 +27,18 @@ function parseTargetCompanies(raw) {
   return pairs.map(pair => {
     const colon = pair.indexOf(':');
     if (colon === -1) die(`invalid --target-companies entry "${pair}" — expected company:org format`);
-    const company = pair.slice(0, colon).trim();
-    const org = pair.slice(colon + 1).trim();
-    if (!company || !org) die(`invalid --target-companies entry "${pair}" — company and org must be non-empty`);
+    const rawCompany = pair.slice(0, colon).trim();
+    const rawOrg = pair.slice(colon + 1).trim();
+    if (!rawCompany || !rawOrg) die(`invalid --target-companies entry "${pair}" — company and org must be non-empty`);
+    // Components are URL-encoded by the Backfill Orchestrator so values
+    // containing ':' or ',' survive the wire format. Decode here.
+    let company, org;
+    try {
+      company = decodeURIComponent(rawCompany);
+      org = decodeURIComponent(rawOrg);
+    } catch (err) {
+      die(`invalid --target-companies entry "${pair}" — malformed URL encoding`);
+    }
     return { company, org };
   });
 }
--- a/data-pipeline/ingester/lib/cli.js
+++ b/data-pipeline/ingester/lib/cli.js
@@ -31,13 +31,18 @@ function parseTargetCompanies(raw) {
     const rawOrg = pair.slice(colon + 1).trim();
     if (!rawCompany || !rawOrg) die(`invalid --target-companies entry "${pair}" — company and org must be non-empty`);
     // Components are URL-encoded by the Backfill Orchestrator so values
-    // containing ':' or ',' survive the wire format. Decode here.
+    // containing ':' or ',' survive the wire format. Decode each side
+    // separately so the operator knows which component was malformed.
     let company, org;
     try {
       company = decodeURIComponent(rawCompany);
+    } catch {
+      die(`invalid --target-companies entry "${pair}" — malformed URL encoding in company component`);
+    }
+    try {
       org = decodeURIComponent(rawOrg);
-    } catch (err) {
-      die(`invalid --target-companies entry "${pair}" — malformed URL encoding`);
+    } catch {
+      die(`invalid --target-companies entry "${pair}" — malformed URL encoding in org component`);
     }
     return { company, org };
   });
--- a/data-pipeline/orchestrators/backfill.js
+++ b/data-pipeline/orchestrators/backfill.js
@@ -153,7 +153,11 @@ async function main() {
     }
 
     // ── Spawn Ingester ────────────────────────────────────────────────────────
-    const targetCompaniesArg = targetRows.map(r => `${r.company}:${r.org_name}`).join(',');
+    // URL-encode each token so values containing ':' or ',' can't break the
+    // company:org,company:org wire format the Ingester parses.
+    const targetCompaniesArg = targetRows
+      .map(r => `${encodeURIComponent(r.company)}:${encodeURIComponent(r.org_name)}`)
+      .join(',');
     const ingesterPath = path.resolve(__dirname, '..', 'ingester', 'ingester.js');
 
     logger.info({ startDate, endDate, companies: targetRows.length }, 'spawning ingester');


tomerp20 (commented):
## Backend Review — STANDARD

Reviewed the wire-format hardening for `--target-companies` between the Backfill Orchestrator and the Ingester. The encode/decode round-trip itself is correct (`encodeURIComponent` produces only `%XX` triplets and unreserved chars; `decodeURIComponent` inverts that without the `+`-as-space pitfall that bites `application/x-www-form-urlencoded` parsers — verified for `+`, `%`, and Unicode/surrogate edge cases). The `URIError` catch is appropriately narrow because the `try` block contains only the two `decodeURIComponent` calls.

This is `data-pipeline/` code — no Express routes, Knex queries, migrations, or auth surface to evaluate. Standard backend categories (route integrity, transactions, N+1, auth/validation, sensitive-data logging) do not apply.

### Issues by Severity

**HIGH — 1**
- `data-pipeline/orchestrators/hourly.js:61` builds the same `--target-companies` wire format but does **not** URL-encode the components. Now that the Ingester *always* `decodeURIComponent`s, the hourly path is silently inconsistent with the backfill path: any `org_name` containing `%` (or any future char that needs escaping) will be mangled by the new decode step on the Ingester side. Current values are URL-safe so it's latent, but the convention has drifted across the two producers — out of scope for this PR's diff, but a follow-up PR is warranted before either side is forgotten.

**LOW — 1**
- `data-pipeline/ingester/lib/cli.js:39` the `catch (err)` binding is unused. `decodeURIComponent` only throws `URIError`, so the catch is effectively typed correctly, but the operator gets no context about *which* component (company vs org) failed. Minor — the `pair` value already gives them enough to debug.

### Testing Gaps

`data-pipeline/` has no test harness (per agent memory). Not blocking this PR — flagging only because the encode/decode contract now spans two files and one consumer, and a single round-trip unit test (`encodeURIComponent('wix:org,with,commas')` → ingester `parseTargetCompanies` → back to original) would prevent the kind of drift already present in `hourly.js`. Consider it for the Cassandra-pipeline follow-up tickets, not this one.

### Verdict

Diff itself is correct and worth landing. The HIGH finding lives in a file you did not touch — Phase 2 should not modify `hourly.js` from this PR (it's outside the stated scope), but it should be tracked as a follow-up.

data-pipeline/orchestrators/backfill.js
```diff
@@ -153,7 +153,11 @@ async function main() {
     }
 
     // ── Spawn Ingester ────────────────────────────────────────────────────────
-    const targetCompaniesArg = targetRows.map(r => `${r.company}:${r.org_name}`).join(',');
+    // URL-encode each token so values containing ':' or ',' can't break the
+    // company:org,company:org wire format the Ingester parses.
+    const targetCompaniesArg = targetRows
+      .map(r => `${encodeURIComponent(r.company)}:${encodeURIComponent(r.org_name)}`)
```
tomerp20, line 159: **[HIGH] — Encode convention drifts from hourly.js (cross-file)**

This change makes the Ingester *always* call `decodeURIComponent` on each component. The sibling orchestrator at `data-pipeline/orchestrators/hourly.js:61` still builds the wire format without encoding:

```js
const targetCompanies = companies.map(r => `${r.company}:${r.org_name}`).join(',');
```

Current values (`microsoft`, `wix-incubator`, etc.) only use `[a-zA-Z0-9-]` so this is latent today. But the moment a company or org_name gains a `%` character (or any char `encodeURIComponent` would escape), the hourly path will produce literal `%` that the Ingester now decodes as the start of an escape sequence — either silently mangling it or throwing `URIError` and calling `die()`.

**Fix (in a follow-up PR — out of scope for this diff):** mirror the same `encodeURIComponent` join in `hourly.js` so the contract has exactly one producer convention.

data-pipeline/ingester/lib/cli.js
```diff
@@ -27,9 +27,18 @@ function parseTargetCompanies(raw) {
...
+    const rawOrg = pair.slice(colon + 1).trim();
+    if (!rawCompany || !rawOrg) die(`invalid --target-companies entry "${pair}" — company and org must be non-empty`);
+    // Components are URL-encoded by the Backfill Orchestrator so values
+    // containing ':' or ',' survive the wire format. Decode here.
+    let company, org;
+    try {
+      company = decodeURIComponent(rawCompany);
+      org = decodeURIComponent(rawOrg);
+    } catch (err) {
+      die(`invalid --target-companies entry "${pair}" — malformed URL encoding`);
```
tomerp20: **[LOW] — Catch binding unused; consider naming the failing component**

The `err` binding is unused, and the operator's only signal is which `pair` failed — they can't tell whether the company or the org component was malformed. Since the only call in the `try` is `decodeURIComponent`, the catch is correctly narrow (only `URIError` is possible), but a slightly more actionable message would help in the (rare) operator-debugging case.

**Current:**
```js
try {
  company = decodeURIComponent(rawCompany);
  org = decodeURIComponent(rawOrg);
} catch (err) {
  die(`invalid --target-companies entry "${pair}" — malformed URL encoding`);
}
```

**Suggested (optional — minor):**
```js
try {
  company = decodeURIComponent(rawCompany);
} catch {
  die(`invalid --target-companies entry "${pair}" — malformed URL encoding in company component`);
}
try {
  org = decodeURIComponent(rawOrg);
} catch {
  die(`invalid --target-companies entry "${pair}" — malformed URL encoding in org component`);
}
```

Not blocking — current behaviour is correct, just slightly less debuggable.

tomerp20:
## Code Review — Final Status

**Routing:** backend reviewer (data-pipeline JS, no React surface). Security review inline by orchestrator.

| Phase | Outcome |
|-------|---------|
| Phase 1 — Backend review | 1 HIGH (out of scope), 1 LOW |
| Phase 1 — Security pass | ✅ Clean — internal IPC only, no user-input surface, narrow `URIError` catch, no injection/auth/CORS implications |
| Phase 2 — Fixes | 1 fixed, 1 deferred to follow-up |
| Phase 3 — Verification | ✅ Round-trip sanity test passes for both normal and diagnostic-error paths |

### Findings & disposition

| Severity | Finding | Action |
|----------|---------|--------|
| HIGH | Cross-producer drift: `data-pipeline/orchestrators/hourly.js:61` builds the same `--target-companies` wire format **without** URL-encoding. After this PR, the Ingester always decodes — so hourly is now silently inconsistent. Latent today (all current names are `[a-zA-Z0-9-]`). | **Deferred** — out of scope (PR title is explicit: "Backfill Orchestrator"). Filed as #231. |
| LOW  | `cli.js:40` — unused `err` binding in catch; operator can't tell whether the company or org component was malformed. | **Fixed** in 1fe9bb9 — split into two try blocks with component-specific error messages. |

### Why the HIGH was not auto-fixed

The reviewer explicitly recommended against modifying `hourly.js` from this PR's branch — it expands scope beyond what the PR title and issue (#225) describe, and would also pull `hourly.js` into the diff that smoke-test guidance was authored against. The follow-up issue #231 contains the exact fix (mirror the `encodeURIComponent` lines from `backfill.js:156-160`, optionally extracted to a shared helper to prevent future drift).

### Verification of the LOW fix

```
microsoft:microsoft       -> {company:'microsoft',     org:'microsoft'}
wix:wix-incubator         -> {company:'wix',           org:'wix-incubator'}
foo%3Abar%2Cbaz:evil-org  -> {company:'foo:bar,baz',   org:'evil-org'}
%ZZ:good                  -> died on company
good:%ZZ                  -> died on org
```

Awaiting human review and approval. **Not merged.**

PR #239, opened 2026-05-24
fix(data-pipeline): restructure processed_files to single-partition for O(1) catchup max lookup

### LINKED ISSUE #238: fix(data-pipeline): restructure processed_files to single-partition for O(1) catchup max lookup
## Background

The Ingester `--catchup` verb crashed in production on 2026-05-23 22:49:30 with:

```
NoHostAvailableError → BusyConnectionError: 2048 requests in-flight on a single connection
```

Cause: `ingester.js:90-91` fires `Promise.all` of `writer.isFileProcessed(id)` over every on-disk hour ID (3,426 the night of the crash) → exceeds the cassandra-driver default `pooling.maxRequestsPerConnection = 2048`. Scale-only bug; worked through Jan backfill (used `--range`) and the Feb 1-5 smoke test.

The hourly ingester processes hours strictly in chronological order and `break`s on the first failure, so `processed_files` is always `{H₀ … H_max}` contiguous. Catchup therefore only needs `MAX(file_name)` — not per-file set membership.

## What to build

Restructure `processed_files` so the latest entry is retrievable in a single O(1) round-trip, and rewrite `--catchup` to use that instead of N concurrent `isFileProcessed` calls.

Schema change:

```cql
CREATE TABLE processed_files (
  bucket       text,        -- always 'singleton'
  file_name    text,
  processed_at timestamp,
  PRIMARY KEY (bucket, file_name)
) WITH CLUSTERING ORDER BY (file_name DESC);
```

Catchup query becomes:

```cql
SELECT file_name FROM processed_files
WHERE bucket='singleton'
LIMIT 1;
```

Catchup logic becomes: `pending = ondisk.filter(id > max)` (lexicographic compare works because file_name format is `YYYY-MM-DD-HH`).

## Acceptance criteria

- [ ] Migration added under `data-pipeline/schema/` recreating `processed_files` with `(bucket, file_name)` PK and `CLUSTERING ORDER BY (file_name DESC)`
- [ ] `cassandra-writer.js`:
  - new `getMaxProcessedFile()` method returning `string | null` (one query)
  - existing `markFileProcessed(id)` writes with `bucket='singleton'`
  - `isFileProcessed(id)` removed or marked unused
- [ ] `ingester.js --catchup` no longer calls `Promise.all` of per-file checks; uses `getMaxProcessedFile()` once and filters on-disk IDs by `> max`
- [ ] Manual smoke test transcript posted on the PR showing:
  1. Empty `processed_files` → catchup processes all on-disk hours in order
  2. Populated `processed_files` → catchup processes only hours `> max`
  3. Catchup over 3,000+ on-disk hours completes without `BusyConnectionError`
- [ ] `knowledge/wiki/adr-0005-processed-files-hourly-only.md` updated: its "No schema change" claim is now superseded; either edit in place or mark `status: stale` linking to the new ADR
- [ ] `knowledge/wiki/ingestion-pipeline.md` updated to describe the new single-partition shape and the `getMaxProcessedFile` catchup path
- [ ] New ADR `knowledge/wiki/adr-0007-processed-files-single-partition.md` (and `docs/adr/0007-processed-files-single-partition.md` to match repo convention) documenting the trade-off: hot-partition anti-pattern accepted because deployment is single-node Cassandra; future cluster migration would need re-partitioning (e.g. `bucket=YYYY`)

## Out of scope

- Per-worker gunzip parallelism (separate perf issue — Cassandra is idle; gunzip on the main thread is the throughput ceiling)
- Defensive `GHARCHIVE_DIR` fail-loud check in fetcher (separate issue)
- Multi-node Cassandra repartitioning (`bucket=YYYY` etc.) — accepted as a future migration if/when topology changes

## Blocked by

None — `processed_files` is currently empty in production, so the migration is a drop+recreate with no data preservation step.

## Suggested model

Sonnet — design is fully decided; remaining work is mechanical implementation + documentation.

tomerp20:
## Summary
- Replaces the N-concurrent-SELECT `isFileProcessed()` pattern in `--verb catchup` with a single `getMaxProcessedFile()` call, eliminating the `BusyConnectionError` that crashed production at 3,426 on-disk files
- Restructures `processed_files` schema to `(bucket='singleton', file_name)` composite PK with `CLUSTERING ORDER BY (file_name DESC)` so `MAX(file_name)` is a single O(1) round-trip
- Adds ADR 0007 documenting the hot-partition trade-off (acceptable for single-node Cassandra; future multi-node would need repartitioning)

## Acceptance criteria
- [x] Migration added under `data-pipeline/schema/` recreating `processed_files` with `(bucket, file_name)` PK and `CLUSTERING ORDER BY (file_name DESC)`
- [x] `cassandra-writer.js`: new `getMaxProcessedFile()` method returning `string | null`; `markFileProcessed(id)` writes with `bucket='singleton'`; `isFileProcessed(id)` removed
- [x] `ingester.js --catchup` no longer calls `Promise.all` of per-file checks; uses `getMaxProcessedFile()` once and filters on-disk IDs by `> max`
- [ ] Manual smoke test transcript posted on the PR (see below — requires production Cassandra access)
- [x] `knowledge/wiki/adr-0005-processed-files-hourly-only.md` updated: "No schema change" claim marked superseded
- [x] `knowledge/wiki/ingestion-pipeline.md` updated to describe new single-partition shape and `getMaxProcessedFile` catchup path
- [x] ADR 0007 created in both `knowledge/wiki/` and `docs/adr/`

## Smoke test

The smoke test transcript (criterion 4) requires running the ingester against a live Cassandra + populated GHARCHIVE_DIR. This must be run on the Linux deployment host. The three scenarios to capture:

1. Empty `processed_files` → catchup processes all on-disk hours in order
2. Populated `processed_files` → catchup processes only hours `> max`
3. Catchup over 3,000+ on-disk hours completes without `BusyConnectionError`

To run migration 009 on the host:
```bash
cqlsh -e "$(cat data-pipeline/schema/009_processed_files_single_partition.cql)"
```

## Related
Closes #238

### DIFF
--- a/data-pipeline/ingester/ingester.js
+++ b/data-pipeline/ingester/ingester.js
@@ -6,7 +6,7 @@ import os from 'os';
 import path from 'path';
 import { fileURLToPath } from 'url';
 import pino from 'pino';
-import { hourIdToPath, enumerateRange, enumerateAllOnDisk } from 'disk-layout';
+import { hourIdToPath, enumerateRange, enumerateAllOnDisk, hourIdToMs } from 'disk-layout';
 import { parseCLI } from './lib/cli.js';
 import { CassandraWriter } from './lib/cassandra-writer.js';
 
@@ -88,7 +88,11 @@ if (cmd.verb === 'hour') {
     process.exit(0);
   }
   const maxProcessed = await writer.getMaxProcessedFile();
-  const pending = maxProcessed ? allOnDisk.filter(id => id > maxProcessed) : allOnDisk;
+  // Compare chronologically (hourIdToMs), not lexicographically — the hour
+  // component of the canonical hourId is unpadded (`YYYY-MM-DD-H`, 0–23), so
+  // string compare yields '…-9' > '…-10'. Using ms timestamps avoids the bug.
+  const maxMs = maxProcessed !== null ? hourIdToMs(maxProcessed) : null;
+  const pending = maxMs !== null ? allOnDisk.filter(id => hourIdToMs(id) > maxMs) : allOnDisk;
   const alreadyProcessed = allOnDisk.length - pending.length;
   logger.info(
     { total: allOnDisk.length, alreadyProcessed, toProcess: pending.length },
@@ -158,9 +162,11 @@ async function processHour(hourId) {
 
   // In --mode hourly: check processed_files for idempotency.
   // Catchup pre-filters to only unprocessed hours, so skip the query there.
+  // Compare chronologically (hourIdToMs), not lexicographically — see catchup
+  // block above for the unpadded-hour rationale.
   if (cmd.mode === 'hourly' && cmd.verb !== 'catchup') {
     const maxProcessed = await writer.getMaxProcessedFile();
-    if (maxProcessed !== null && hourId <= maxProcessed) {
+    if (maxProcessed !== null && hourIdToMs(hourId) <= hourIdToMs(maxProcessed)) {
       logger.info({ hourId }, 'already processed, skipping');
       return 0;
     }
@@ -296,7 +302,7 @@ async function processHour(hourId) {
   // In --mode hourly: mark file processed
   if (cmd.mode === 'hourly') {
     await writer.markFileProcessed(hourId);
-    logger.info({ hourId }, 'processed_files row written');
+    logger.info({ hourId, totalParsed, totalFiltered }, 'processed_files row written');
   }
 
   return totalFiltered;
--- a/data-pipeline/ingester/ingester.js
+++ b/data-pipeline/ingester/ingester.js
@@ -80,15 +80,15 @@ if (cmd.verb === 'hour') {
 } else if (cmd.verb === 'range') {
   hourIds = enumerateRange(cmd.start, cmd.end);
 } else if (cmd.verb === 'catchup') {
-  // catchup: walk disk, skip files already recorded in processed_files
+  // catchup: walk disk, find all hours after the highest already-processed hour
   const allOnDisk = enumerateAllOnDisk(GHARCHIVE_DIR);
   if (allOnDisk.length === 0) {
     logger.warn('no files in GHARCHIVE_DIR');
     await writer.shutdown();
     process.exit(0);
   }
-  const processedFlags = await Promise.all(allOnDisk.map(id => writer.isFileProcessed(id)));
-  const pending = allOnDisk.filter((_, i) => !processedFlags[i]);
+  const maxProcessed = await writer.getMaxProcessedFile();
+  const pending = maxProcessed ? allOnDisk.filter(id => id > maxProcessed) : allOnDisk;
   const alreadyProcessed = allOnDisk.length - pending.length;
   logger.info(
     { total: allOnDisk.length, alreadyProcessed, toProcess: pending.length },
@@ -156,10 +156,11 @@ process.exit(exitCode);
 async function processHour(hourId) {
   const filePath = hourIdToPath(GHARCHIVE_DIR, hourId);
 
-  // In --mode hourly: check processed_files for idempotency
-  if (cmd.mode === 'hourly') {
-    const alreadyDone = await writer.isFileProcessed(hourId);
-    if (alreadyDone) {
+  // In --mode hourly: check processed_files for idempotency.
+  // Catchup pre-filters to only unprocessed hours, so skip the query there.
+  if (cmd.mode === 'hourly' && cmd.verb !== 'catchup') {
+    const maxProcessed = await writer.getMaxProcessedFile();
+    if (maxProcessed !== null && hourId <= maxProcessed) {
       logger.info({ hourId }, 'already processed, skipping');
       return 0;
     }
@@ -294,7 +295,7 @@ async function processHour(hourId) {
 
   // In --mode hourly: mark file processed
   if (cmd.mode === 'hourly') {
-    await writer.markFileProcessed(hourId, totalParsed, totalFiltered);
+    await writer.markFileProcessed(hourId);
     logger.info({ hourId }, 'processed_files row written');
   }
 
--- a/data-pipeline/ingester/lib/cassandra-writer.js
+++ b/data-pipeline/ingester/lib/cassandra-writer.js
@@ -39,7 +39,7 @@ export class CassandraWriter {
     this._limit = null;
     this._insertCql = null;
     this._processedFilesInsertCql = null;
-    this._processedFilesSelectCql = null;
+    this._processedFilesMaxCql = null;
     this._backfillProgressInsertCql = null;
 
     // Partition write rate tracking
@@ -56,11 +56,11 @@ export class CassandraWriter {
       `INSERT INTO ${this._keyspace}.company_events
          (company, org_name, year_month, event_time, event_id, event_type, repo_name, actor_login, is_ai, tech_tags)
        VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`;
-    this._processedFilesSelectCql =
-      `SELECT file_name FROM ${this._keyspace}.processed_files WHERE file_name = ?`;
+    this._processedFilesMaxCql =
+      `SELECT file_name FROM ${this._keyspace}.processed_files WHERE bucket = ? LIMIT 1`;
     this._processedFilesInsertCql =
-      `INSERT INTO ${this._keyspace}.processed_files (file_name, processed_at, event_count, filtered_count)
-       VALUES (?, ?, ?, ?)`;
+      `INSERT INTO ${this._keyspace}.processed_files (bucket, file_name, processed_at)
+       VALUES (?, ?, ?)`;
     this._backfillProgressInsertCql =
       `INSERT INTO ${this._keyspace}.backfill_progress (run_id, date, status, completed_at, events_written)
        VALUES (?, ?, ?, ?, ?)`;
@@ -87,9 +87,9 @@ export class CassandraWriter {
     if (this._rateTimer) clearInterval(this._rateTimer);
   }
 
-  async isFileProcessed(fileName) {
-    const result = await this._client.execute(this._processedFilesSelectCql, [fileName], { prepare: true });
-    return result.rowLength > 0;
+  async getMaxProcessedFile() {
+    const result = await this._client.execute(this._processedFilesMaxCql, ['singleton'], { prepare: true });
+    return result.rowLength > 0 ? result.rows[0].file_name : null;
   }
 
   async writeEvent(event) {
@@ -208,10 +208,10 @@ export class CassandraWriter {
     }
   }
 
-  async markFileProcessed(fileName, eventCount, filteredCount) {
+  async markFileProcessed(fileName) {
     await this._client.execute(
       this._processedFilesInsertCql,
-      [fileName, new Date(), eventCount, filteredCount],
+      ['singleton', fileName, new Date()],
       { prepare: true }
     );
   }
--- a/data-pipeline/ingester/lib/cassandra-writer.js
+++ b/data-pipeline/ingester/lib/cassandra-writer.js
@@ -40,7 +40,7 @@ export class CassandraWriter {
     this._limit = null;
     this._insertCql = null;
     this._processedFilesInsertCql = null;
-    this._processedFilesScanCql = null;
+    this._processedFilesMaxCql = null;
     this._backfillProgressInsertCql = null;
 
     // Partition write rate tracking
@@ -57,17 +57,13 @@ export class CassandraWriter {
       `INSERT INTO ${this._keyspace}.company_events
          (company, org_name, year_month, event_time, event_id, event_type, repo_name, actor_login, is_ai, tech_tags)
        VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`;
-    // Scan the singleton partition and compute the chronological max in JS.
-    // We cannot use `LIMIT 1` on the clustered DESC table because `file_name`
-    // sorts lexicographically and the canonical hour ID has an unpadded hour
-    // component (`YYYY-MM-DD-H`, 0–23), so `'…-9' > '…-10'` as strings.
-    // Scanning the whole partition once is a single round-trip; the partition
-    // stays bounded (~26k rows per 3 years of hourly ingest).
-    this._processedFilesScanCql =
-      `SELECT file_name FROM ${this._keyspace}.processed_files WHERE bucket = ?`;
+    // LIMIT 1 on the hour_time DESC clustering column is a true O(1) key lookup —
+    // timestamp sorts chronologically, so the first row is always the latest hour.
+    this._processedFilesMaxCql =
+      `SELECT file_name FROM ${this._keyspace}.processed_files WHERE bucket = ? LIMIT 1`;
     this._processedFilesInsertCql =
-      `INSERT INTO ${this._keyspace}.processed_files (bucket, file_name, processed_at)
-       VALUES (?, ?, ?)`;
+      `INSERT INTO ${this._keyspace}.processed_files (bucket, hour_time, file_name, processed_at)
+       VALUES (?, ?, ?, ?)`;
     this._backfillProgressInsertCql =
       `INSERT INTO ${this._keyspace}.backfill_progress (run_id, date, status, completed_at, events_written)
        VALUES (?, ?, ?, ?, ?)`;
@@ -95,22 +91,10 @@ export class CassandraWriter {
   }
 
   // Returns the chronologically latest processed file_name, or null if none.
-  // Single Cassandra round-trip (paged read of the singleton partition).
-  // Max is computed in JS via hourIdToMs so the ordering is chronological,
-  // not lexicographic — the hour component of file_name is unpadded.
+  // O(1): LIMIT 1 on the hour_time DESC clustering column is a direct key lookup.
   async getMaxProcessedFile() {
-    const result = await this._client.execute(this._processedFilesScanCql, ['singleton'], { prepare: true });
-    if (result.rowLength === 0) return null;
-    let maxId = null;
-    let maxMs = -Infinity;
-    for (const row of result.rows) {
-      const ms = hourIdToMs(row.file_name);
-      if (ms > maxMs) {
-        maxMs = ms;
-        maxId = row.file_name;
-      }
-    }
-    return maxId;
+    const result = await this._client.execute(this._processedFilesMaxCql, ['singleton'], { prepare: true });
+    return result.rowLength > 0 ? result.rows[0].file_name : null;
   }
 
   async writeEvent(event) {
@@ -232,7 +216,7 @@ export class CassandraWriter {
   async markFileProcessed(fileName) {
     await this._client.execute(
       this._processedFilesInsertCql,
-      ['singleton', fileName, new Date()],
+      ['singleton', new Date(hourIdToMs(fileName)), fileName, new Date()],
       { prepare: true }
     );
   }
--- a/data-pipeline/ingester/lib/cassandra-writer.js
+++ b/data-pipeline/ingester/lib/cassandra-writer.js
@@ -1,5 +1,6 @@
 import cassandra from 'cassandra-driver';
 import pLimit from 'p-limit';
+import { hourIdToMs } from 'disk-layout';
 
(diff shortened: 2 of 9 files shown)


tomerp20 (commented):
# Code Review — PR #239: `processed_files` single-partition for O(1) catchup

**Reviewer:** Data-pipeline orchestrator (inline, model: opus) + security pass
**Knowledge base consulted:** [[adr-0007-processed-files-single-partition]], [[adr-0005-processed-files-hourly-only]], [[adr-0003-backfill-hourly-relay-race]], [[adr-0004-microservices-shaped-cli-contract]], [[ingestion-pipeline]], `docs/InjestionPlan.md` §3.3, §5.6.

## Summary table

| Severity | Count |
|---|---|
| HIGH (correctness) | 1 |
| MEDIUM | 3 |
| LOW | 5 |
| Security | 0 |

## Verdict

**Do not merge.** A single HIGH-severity correctness bug will cause silent data loss on every catchup once any day rolls past hour 9. See finding **H1** below.

The rest of the design is sound — the architectural decision in ADR 0007 is well-reasoned, the single-partition trade-off is honest, and the migration is appropriately conservative for the documented "empty in production" precondition.

---

## H1 — HIGH: Lexicographic max breaks once hour ID rolls past 9 → silent data loss

**Files:** `data-pipeline/ingester/ingester.js:91`, `data-pipeline/ingester/lib/cassandra-writer.js:60`, `data-pipeline/schema/009_processed_files_single_partition.cql:21`

The hour ID format is **`YYYY-MM-DD-H` with the hour unpadded** (`0`-`23`). This is documented authoritatively in `data-pipeline/shared/disk-layout/index.js:4` and `docs/InjestionPlan.md` §3.3:

> the canonical hour ID `YYYY-MM-DD-H` (hour unpadded — `2025-05-01-15`)

This breaks the new design in two coupled places:

1. **Cassandra clustering order** (`CLUSTERING ORDER BY (file_name DESC)`) sorts **lexicographically on the string**, not chronologically.
2. **`allOnDisk.filter(id => id > maxProcessed)`** in `ingester.js:91` is also a **lexicographic JS string compare**.

Concrete failure trace once production hits its first day past hour 9:

```
After processing 2025-05-24-9, processed_files contains:
  '2025-05-23-9', '2025-05-23-8', ..., '2025-05-23-0',
  '2025-05-23-23', '2025-05-23-22', ..., '2025-05-23-10',
  '2025-05-24-9', '2025-05-24-8', ..., '2025-05-24-0'

CLUSTERING ORDER BY (file_name DESC) returns first row → '2025-05-24-9'
(because '9' > '8' > '7' > ... > '23' > '22' > ... > '10' > '0' lexicographically)

Next catchup: filter(id => id > '2025-05-24-9')
  '2025-05-24-10'  ?>  '2025-05-24-9'   →  FALSE (filtered out, SKIPPED)
  '2025-05-24-23'  ?>  '2025-05-24-9'   →  FALSE (filtered out, SKIPPED)
  '2025-05-25-0'   ?>  '2025-05-24-9'   →  FALSE (filtered out, SKIPPED)
  ...
  '2025-05-25-10'  ?>  '2025-05-24-9'   →  FALSE (filtered out, SKIPPED)
```

**Effect:** Every catchup silently skips 14 of every 24 hours per day, indefinitely. The `processed_files` row count stops growing in step with on-disk files, but no error is raised — the very class of failure ADR 0005 was written to prevent (silent skipping). This violates the contiguity invariant ADR 0007 leans on.

The bug is currently masked because:
- Production has been on `--range` (backfill) until issue #238.
- The first post-merge catchup on a fresh `processed_files` will appear to work on hours `0`–`9` before the corruption begins.
- The smoke test in the PR description doesn't include a scenario crossing hour `9`.

### Fix options

1. **Change the format to zero-padded** `YYYY-MM-DD-HH` — requires updating `disk-layout/index.js`, `pathToHourId`, `msToHourId`, every fetcher/orchestrator caller, and a data migration of the on-disk archive paths. Largest blast radius; cleanest long-term.
2. **Use `hourIdToMs` for both the Cassandra column and the JS filter.** Store a `timestamp` (or epoch ms `bigint`) as the clustering column instead of `file_name`. Keep `file_name` as a non-clustering column. The clustering-DESC + JS `>` compares both become chronological.
3. **Replace `getMaxProcessedFile()` with `getAllProcessedFiles()` Set membership** (single SELECT, full partition scan with `LIMIT 100000` or paging). Then `pending = onDisk.filter(id => !set.has(id))`. Returns to the original semantics; loses the O(1) catchup property but stays O(1) round-trips. Cassandra single-partition scan of ~26k rows/3 years is fine.
4. **Sort by `hourIdToMs(id)` in JS on the filter side**, and use the **max chronologically** by paging the Cassandra `LIMIT 1` query — i.e. pull all rows, take `max(rows, key=hourIdToMs)`. Same round-trip count, but compute the max correctly. This still needs the Cassandra clustering order changed (or removed) so reads aren't lexicographic.

Option 3 is the smallest diff and restores correctness exactly. Option 2 is the right design if O(1) catchup must be preserved.

This violates [[adr-0007-processed-files-single-partition]]'s contiguity assumption: the table is contiguous chronologically, but reading the "max" lexicographically returns a non-max element.

Also: `knowledge/wiki/ingestion-pipeline.md` line 47 and `knowledge/wiki/adr-0007-processed-files-single-partition.md` both incorrectly assert *"lexicographic compare works because file_name format is YYYY-MM-DD-HH"* — this format is not what the rest of the pipeline produces.

---

## M1 — MEDIUM: Migration is destructive and not idempotent

**File:** `data-pipeline/schema/009_processed_files_single_partition.cql:15-22`

```cql
DROP TABLE IF EXISTS jobflow.processed_files;

CREATE TABLE jobflow.processed_files (
    bucket       text,
    file_name    text,
    ...
```

There is no migration runner with version tracking in `data-pipeline/schema/` — migrations are operator-applied. Re-running this script after rows exist will wipe them. The header comment says *"Production processed_files is empty at migration time → drop+recreate with no data loss"* but the script itself doesn't enforce or check that precondition.

Options:
- Add a `SELECT COUNT(*)` precondition note in a header comment with the exact pre-migration assertion the operator must run.
- Split into `009a_drop_processed_files.cql` and `009b_create_processed_files_v2.cql` so the destructive step is named clearly and harder to re-run by accident.
- Use `CREATE TABLE IF NOT EXISTS` after the DROP so accidental re-runs at least don't shadow the DROP error.

The DROP itself is the right call given the PK change — flagging the operator-safety concern, not the schema decision.

---

## M2 — MEDIUM: Per-hour `getMaxProcessedFile()` in `processHour` does the same work N times

**File:** `data-pipeline/ingester/ingester.js:160-166`

```js
if (cmd.mode === 'hourly' && cmd.verb !== 'catchup') {
  const maxProcessed = await writer.getMaxProcessedFile();
  if (maxProcessed !== null && hourId <= maxProcessed) { ... }
}
```

For `--hour --mode hourly` (1 file) this is a single query — fine. For `--range --mode hourly` (which CLI parsing allows, even if no orchestrator currently uses it — see `cli.js:108-109` where `--range` defaults to `backfill` but accepts `hourly`), this fires N queries against the singleton partition where one would do. Pre-fetch once before the loop:

```js
const maxAtStart = (cmd.mode === 'hourly' && cmd.verb !== 'catchup')
  ? await writer.getMaxProcessedFile()
  : null;
// then inside processHour: pass maxAtStart in as an arg
```

Minor perf issue (singleton partition reads are cheap), but the pattern is the same shape as the H1 catchup bug — fetch once, compare in-process.

---

## M3 — MEDIUM: No test coverage for catchup filter logic

**Files:** none — gap

There are no `*.test.js` files under `data-pipeline/ingester/` (only inside `node_modules/`). The H1 bug would have been caught instantly by a 4-line test:

```js
expect(filterPending(['2025-05-01-9', '2025-05-01-10'], '2025-05-01-9'))
  .toEqual(['2025-05-01-10']);
```

Smoke tests are gated on Linux host access per the PR description — that doesn't substitute for a unit test of pure-function filter logic. This is exactly the kind of latent scale/ordering bug a single test would prevent.

This is a structural gap not introduced by this PR, but the PR is a perfect prompt to start coverage.

---

## L1 — LOW: Truthiness check on `maxProcessed` differs between callers

**File:** `data-pipeline/ingester/ingester.js:91` vs `:163`

Line 91 (catchup): `maxProcessed ? allOnDisk.filter(...) : allOnDisk;` — uses truthy check.
Line 163 (processHour): `if (maxProcessed !== null && hourId <= maxProcessed)` — uses explicit null check.

Empty-string `file_name` is currently impossible (writes are constrained), but defensive consistency: use `maxProcessed !== null` in both places. If the column ever becomes nullable / `''` ever appears, the catchup branch would silently reprocess the whole archive while `processHour` would skip the check entirely — opposite behaviors from the same value.

---

## L2 — LOW: `getMaxProcessedFile()` query is implicitly coupled to clustering order

**File:** `data-pipeline/ingester/lib/cassandra-writer.js:60`

```cql
SELECT file_name FROM processed_files WHERE bucket = ? LIMIT 1
```

This returns the row at the head of the table-level `CLUSTERING ORDER BY (file_name DESC)`. Correct today, but if the schema's clustering order ever changes (e.g. a future migration drops `DESC` to allow forward iteration for some other read), this query silently flips to returning the *minimum* without any error. Add an explicit `ORDER BY file_name DESC` so the query is self-documenting and any future incompatibility is caught at execution time.

Also — and this matters most given H1 — the query gives a hint that the design depends on a string sort: there's no `MAX()`, no `ORDER BY`, just `LIMIT 1`. That implicit contract is what made H1 easy to miss.

---

## L3 — LOW: Contiguity invariant is documented but unenforced

**Files:** `docs/adr/0007-processed-files-single-partition.md`, `data-pipeline/ingester/lib/cassandra-writer.js:211`

ADR 0007 hangs correctness on:

> the hourly ingester processes files strictly in chronological order and stops on first failure → `processed_files` is always a contiguous prefix.

Nothing in code enforces this. If someone later adds a `--continue-on-error` flag, a `--retry-hour <id>` helper, or a manual `cqlsh` insert, `processed_files` becomes non-contiguous and `getMaxProcessedFile()`-based catchup silently skips the gap.

Suggestions:
- Add a JSDoc on `markFileProcessed()` citing the invariant.
- Add a small operator script `data-pipeline/scripts/verify-contiguity.js` that pages the partition and asserts the IDs form a prefix.

---

## L4 — LOW: Dropped columns `event_count` / `filtered_count` have no replacement observability

**File:** `data-pipeline/ingester/lib/cassandra-writer.js:62`

The old schema wrote `event_count` and `filtered_count` per file. ADR 0007 correctly notes they were never read. But they were the only persisted record of "did this hour produce events" — useful for ad-hoc debugging via `cqlsh`. The ingester log line at `ingester.js:299` (`'processed_files row written'`) doesn't include the parsed/filtered totals either.

Suggestion: bump the log line to include `{ hourId, totalFiltered }` (already in scope at line 297) so the operator-visible audit trail isn't strictly worse than before.

---

## L5 — LOW: `knowledge/wiki/adr-0005-processed-files-hourly-only.md` superseded banner is at top, body still asserts the old claim

The body of the wiki page (line 25) still says:

> No schema change. `processed_files` keeps its existing primary key `(file_name)`.

— in the "Knock-on effects" bullet, without any inline `~~strikethrough~~` or "(superseded — see ADR 0007)" tag. The top-of-page banner is good but readers grep'ing for "schema change" land on the stale claim. Inline the supersession in the bullet itself.

Also: front-matter has `status: stable` and `adr_status: proposed` — slight semantic split but consistent with the rest of the wiki's convention. Just calling it out.

---

## Security pass

**Status: PASS — no concerns.**

- No new secrets, credentials, or auth surface.
- No user input flows into new CQL. `bucket = 'singleton'` is hardcoded; `file_name` parameters come from disk enumeration / CLI-validated format. All bound via prepared statements.
- `SELECT ... LIMIT 1` on the singleton partition is not a DoS vector.
- `DROP TABLE` is operator-run, not exposed via any RPC/HTTP surface.
- No PII or sensitive data in new log lines.
- Cassandra `localOne` consistency unchanged; appropriate for single-node deployment.
- No new dependencies.

---

## Knowledge base citations used

- [[adr-0007-processed-files-single-partition]] — design under review (H1 violates its contiguity assumption)
- [[adr-0005-processed-files-hourly-only]] — establishes that silent skipping is the failure mode this subsystem must prevent (H1 reintroduces a variant of it)
- [[adr-0003-backfill-hourly-relay-race]] — establishes that hourly mode is the only `processed_files` writer (relevant for L1 invariant)
- [[adr-0004-microservices-shaped-cli-contract]] — confirms `--mode` + `--verb` matrix that M2 navigates
- [[ingestion-pipeline]] — wiki page that needs correction (H1 also documents the wrong format)
- `docs/InjestionPlan.md` §3.3 — canonical authority on unpadded hour format

tomerp20:
# Phase 3 — Final Status

## Agents that ran

| Agent | Type | Files reviewed |
|---|---|---|
| Data-pipeline reviewer (inline, opus) | domain | ingester.js, cassandra-writer.js, migration 009, ADR 0007, wiki pages |
| Security pass (inline) | cross-cutting | full diff |

(One agent only because this PR has no frontend/backend code — pure data-pipeline + docs.)

## Findings

| ID | Severity | Title | Status |
|---|---|---|---|
| H1 | HIGH | Lexicographic max breaks once hour ID rolls past 9 → silent data loss | **FIXED** in 955a29c |
| M1 | MEDIUM | Migration is destructive and not idempotent | **PARTIAL** — added operator precondition + inline pitfall doc; structural fix (split scripts) skipped to avoid migration churn |
| M2 | MEDIUM | Per-hour `getMaxProcessedFile()` does the same work N times in `processHour` | **SKIPPED** — currently dormant (no orchestrator uses `--range --mode hourly`); pre-fetch refactor is non-trivial and risks Phase-2 churn |
| M3 | MEDIUM | No test coverage for catchup filter logic | **PARTIAL** — added a local pure-function smoke test that confirmed H1 + verified the fix; in-repo test suite still missing (structural gap, out of patch scope) |
| L1 | LOW | Truthiness check on `maxProcessed` differs between callers | **FIXED** in 955a29c |
| L2 | LOW | `getMaxProcessedFile()` query implicitly coupled to clustering order | **FIXED** in 955a29c (query is now an explicit full scan; clustering order is documented as cqlsh-nicety only) |
| L3 | LOW | Contiguity invariant documented but unenforced | **SKIPPED** — operator-tooling addition; out of patch scope |
| L4 | LOW | Dropped `event_count`/`filtered_count` have no replacement observability | **FIXED** in 955a29c (added counts to log line) |
| L5 | LOW | Wiki `adr-0005` superseded note: body still asserts old "No schema change" | **SKIPPED** — wiki polish; doesn't affect correctness |
| Security | — | — | **PASS** |

## Fix verification

The H1 fix was verified via a pure-function smoke test that mirrored the post-fix
`filterPending()` and `isAlreadyProcessed()` logic against the actual
`hourIdToMs` from `data-pipeline/shared/disk-layout/index.js`. All assertions
pass, including the explicit demonstration that the old lex-compare logic
would have skipped `'2025-05-01-10'` and `'2025-05-01-23'` after processing
`'2025-05-01-9'` — i.e. the exact production bug.

The fix commit (955a29c) also:
- Updated the migration comment to document why `LIMIT 1` is the wrong primitive here.
- Corrected the ADR 0007 + wiki claims about format (`YYYY-MM-DD-H`, not `YYYY-MM-DD-HH`).
- Kept the `CLUSTERING ORDER BY (file_name DESC)` clause for cqlsh convenience, but the application no longer depends on it for correctness.

## Skipped fixes (need human follow-up)

- **M1 (full split):** consider splitting `009_processed_files_single_partition.cql` into `009a_drop` + `009b_create` if the project starts treating schema files as re-runnable.
- **M2:** if `--range --mode hourly` ever becomes a real path (no orchestrator currently uses it), pre-fetch `getMaxProcessedFile()` once before the loop instead of per-hour.
- **M3:** add a `data-pipeline/ingester/test/` directory with at minimum a unit test for the catchup-pending filter — H1 demonstrates the ROI.
- **L3:** add a small `verify-contiguity.js` operator script and a JSDoc on `markFileProcessed()` citing the invariant.
- **L5:** inline-strikethrough the "No schema change" bullet in `knowledge/wiki/adr-0005-processed-files-hourly-only.md`.

## Smoke test before merge

The PR's manual smoke test (Linux host + populated GHARCHIVE_DIR) still
needs to be run. The recommended scenarios:
1. Empty `processed_files` → catchup processes all hours in order.
2. `processed_files` populated up to `YYYY-MM-DD-9` → catchup correctly
   processes hours 10..23 of that day and onward (this is the H1
   regression scenario — it would have silently failed before the fix).
3. 3,000+ on-disk hours → completes without `BusyConnectionError`.

tomerp20:
# Knowledge-base proposals from this review

These are **proposals only** — not edits. Human decides whether to apply.

## (a) `knowledge/wiki/` page edits

### 1. New page: `wiki/hour-id-format.md` (type: `concept`)

The whole H1 bug stems from the wiki + ADR + ingestion-pipeline page all
asserting `YYYY-MM-DD-HH` while the codebase consistently uses
`YYYY-MM-DD-H` (unpadded). A dedicated concept page anchors the canonical
format and gives every other page a single `[[hour-id-format]]` to link.

Frontmatter sketch:
```yaml
---
title: Hour ID Format
slug: hour-id-format
type: concept
tags: [pipeline, ingestion, gh-archive, data-format]
sources:
  - data-pipeline/shared/disk-layout/index.js
  - docs/InjestionPlan.md
related: [[ingestion-pipeline]] [[adr-0007-processed-files-single-partition]] [[gh-archive]]
updated: 2026-05-24
status: stable
---
```

Body should cover: the format (`YYYY-MM-DD-H`, hour unpadded 0–23), the
historical reason (matches GH Archive URL convention), the trap
(lexicographic compare is incorrect — use `hourIdToMs`), and pointers
to `hourIdToMs` / `hourIdToPath` / `msToHourId`.

### 2. Edit: `wiki/ingestion-pipeline.md`

Already corrected in fix commit 955a29c. Future improvement: pin the
catchup section's algorithm description to `[[hour-id-format]]` once that
page exists.

### 3. Edit: `wiki/adr-0007-processed-files-single-partition.md`

Already corrected in fix commit 955a29c. Future improvement: add a
"Lessons" subsection citing why a `LIMIT 1`-style optimization required
the format to be lexicographically chronological.

### 4. Edit: `wiki/adr-0005-processed-files-hourly-only.md`

Inline-strikethrough or annotate the "No schema change" bullet at line 25
of the wiki page body. The top-of-page banner is good but the bullet is
still grep-able as a stale fact.

## (b) `knowledge/learnings.md` entries to append

```markdown
## 2026-05-24 (review of PR #239)

- **Format claims that get repeated across ADR + wiki + code comments
  silently rot together when the underlying format is non-obvious.** The
  H1 bug in PR #239 had ADR 0007, the wiki page, and the ingestion-pipeline
  page all asserting `YYYY-MM-DD-HH` (zero-padded) while the actual format
  per `disk-layout/index.js` is `YYYY-MM-DD-H` (unpadded). The bug was
  invisible to readers because every doc cross-checked against every other
  doc. → When a wiki page mentions a low-level data format, link to a
  single concept page (e.g. `[[hour-id-format]]`) instead of inlining the
  format; otherwise the format gets duplicated and the duplications drift.

- **A "single-query" optimization that relies on database sort order also
  relies on the data being comparable in that sort.** The original PR
  picked `LIMIT 1` on a `CLUSTERING ORDER BY DESC` table without checking
  whether `file_name`'s sort order was chronological. → Whenever an ADR
  proposes "we only need the max — read one row," demand a one-line proof
  that the column's sort is the right sort for the comparison.

- **The reviewer caught H1 by walking the data, not by reading the ADR.**
  The bug review path: read the ADR (looked fine) → grep for the format
  spec (`disk-layout/index.js:4`) → compute a worked example by hand
  (`'2025-05-01-9' > '2025-05-01-10'`). → Future grills/reviews of "we
  use the max value in column X" decisions should always do a manual
  example with two adjacent values to verify the sort is the right one.

- **Manual smoke tests don't substitute for pure-function unit tests of
  ordering logic.** The PR's planned smoke test would NOT have caught
  H1 unless it crossed hour 9 specifically. A 4-line unit test would
  have caught it on the author's first run. → For PRs that introduce
  ordering / comparison primitives over a typed key (hour IDs, timestamps,
  uuids, anything semi-structured), at minimum add a unit test that
  exercises two adjacent values where the natural string order differs
  from the semantic order.
```

PR #241, opened 2026-05-24
perf(data-pipeline): move decompression + line iteration into ingester workers

### LINKED ISSUE #240: perf(data-pipeline): move decompression + line iteration into ingester workers
## Background

The ingester is currently single-thread bound on the main JS thread. Smoke-test data (Feb 1-5, 2026 backfill, 5 days, 13,604 events written): **155.83s wall clock**.

Live diagnostic during the 2026-05-24 catchup run confirmed where the time is going:

- **Main thread: 60% CPU** (single hot thread — the readline `'line'` event loop + `postMessage` dispatch)
- **4 worker threads: 0-10% CPU each** — workers are starved, mostly idle
- **libuv threadpool: 10-20% CPU** — gunzip runs here, comfortably under capacity
- **Cassandra: 99.4% idle** (\`nodetool tpstats\` — MutationStage Active=0, Pending=0, zero dropped messages)
- **System: 82-85% idle** — only ~2.7 of 16 cores in use

The bottleneck is the main-thread serial pipeline at \`ingester.js\`:

\`\`\`js
createReadStream(filePath)
  .pipe(createGunzip())          // off-thread (libuv), but...
  .pipe(createInterface(...))    // readline emits 'line' on main thread
  .on('line', line => {          // main thread runs this loop
    batch.push(line);
    if (batch.length >= 1000) dispatchBatch(toSend);  // postMessage
  });
\`\`\`

All decompressed bytes funnel through one event loop before reaching workers. Workers do CPU-light work (substring filter → JSON.parse → tag/AI extraction) and finish faster than the main thread can feed them.

**Backlog context:** 8,811 hour files on \`/mnt/hdd/gharchive\` (~1 TB compressed). At current pace this is ~15 hours of runtime. Target post-fix: ~4-5 hours.

## Decision: do NOT pre-decompress files to disk

Considered and rejected:

- **Disk space blocker:** 8,811 × ~1 GB decompressed = 8.8 TB on a 2 TB HDD.
- **HDD throughput regression:** reading 1 GB plain (~10s at 100-150 MB/s HDD seq) is **10× worse** than reading 100 MB compressed + in-memory gunzip. Gzip is effectively a 10× IO speedup on spinning rust.
- **Wrong bottleneck:** libuv (where gunzip runs) is at 10-20%, not pinned. Pre-decompressing wouldn't unblock the main-thread readline loop, which is the actual choke.

The fix is to keep streaming compressed reads from disk but parallelize the streaming itself across workers.

## What to build

**Shape change:**

| Today | Proposed |
|-------|----------|
| Main: \`read → gunzip → readline → postMessage(line batches)\` | Main: dispatches \`{filePath, hourId}\` to worker |
| Worker: \`filter → JSON.parse → extract → postMessage(events)\` | Worker: \`read → gunzip → readline → filter → JSON.parse → extract → postMessage(events)\` |
| Main: Cassandra batch writes | Main: Cassandra batch writes *(unchanged)* |
| Workers spawned per-hour, terminated per-hour | Persistent worker pool, file-path queue across the whole run |

**Why Cassandra writes stay on main:** one Client, one batch coordinator, no race on \`processed_files\` row write, no need to deduplicate batches across workers. Server is 99.4% idle, so funneling all writes through main is not the bottleneck.

## Acceptance criteria

- [ ] \`worker.js\` opens its own \`createReadStream → createGunzip → createInterface\` for each dispatched file path; performs substring filter, \`JSON.parse\`, tag/AI extraction; posts extracted events back to main in batches.
- [ ] \`ingester.js\` main thread becomes a thin coordinator: maintains a queue of file paths, dispatches one file per idle worker, awaits per-file completion ack before recording \`processed_files\`.
- [ ] **Persistent worker pool** for the whole run — workers spawn once at startup, terminate once at shutdown. The per-hour \`spawn → terminate\` pattern at \`ingester.js:170, 281\` is removed.
- [ ] **Backpressure:** main thread does NOT dispatch the next file to a worker while events-in-flight to Cassandra exceeds the existing 10K watermark (\`ingester.js:242\`). Workers idle (do not advance the queue) until Cassandra drains below the watermark.
- [ ] **Per-file completion contract:** worker posts \`{ type: 'fileDone', hourId }\` after its final batch. Main thread writes the \`processed_files\` row for that hour only after **all** events from that hour have been flushed to Cassandra (existing drain semantics preserved).
- [ ] \`INGEST_WORKERS\` default ceiling raised from \`Math.min(4, cores/2)\` to \`Math.min(8, cores - 2)\` (env override still respected).
- [ ] \`UV_THREADPOOL_SIZE\` set to 16 in \`run-backfill.sh\` and \`run-hourly.sh\` (cheap insurance — 8 workers × 1 gunzip each + spare slots).
- [ ] **Failure semantics preserved:** if a worker errors on a file, main treats it as a fatal failure and \`break\`s the loop (matches existing chronological-order + first-failure-stops invariant from ADR 0005).
- [ ] **Chronological order preserved:** main dispatches files to workers in chronological order; the \`processed_files\` write happens in chronological order even though workers may complete files in slightly different wall-clock order. (Either gate dispatch to one-file-ahead, or buffer fileDone events and emit \`processed_files\` writes in order.)
- [ ] **Smoke test transcript** (per ADR 0006): re-run the Feb 1-5 backfill that produced the 155s baseline. Capture \`nodetool tpstats\`, \`top -H\`, \`vmstat 1 5\`, wall clock, events-written count. Target: **≥ 2× wall-clock improvement** (155s → ≤ 75s). Ship if met.
- [ ] **New ADR 0008** at \`docs/adr/0008-ingester-worker-side-decompression.md\` documenting:
  - Bottleneck diagnosis (this issue's data)
  - Why pre-decompress-to-disk was rejected (the 10× IO regression on HDD + disk-space blocker)
  - Why writes stay on main (server idle; one-Client simplicity)
  - Trade-off: per-worker decompression buffer is in RAM (~100-200 MB per worker peak) → 8 workers × 200 MB = 1.6 GB peak. On a 16 GB box this is fine; note for future scale.
- [ ] Wiki page \`knowledge/wiki/ingestion-pipeline.md\` updated to reflect the new worker architecture.

## Out of scope

- **Cassandra connection bump / batch size raise** (\`coreConnectionsPerHost\`, \`BATCH_BYTE_LIMIT\`, \`BATCH_ROW_LIMIT\`). Cassandra is 99.4% idle today; fixing this won't move the needle. Worth shipping as a defensive cleanup PR *after* this lands and we measure whether the driver-side queue ever fills up.
- **Parallel file ingest per worker** (each worker streams 2 files concurrently). Adds complexity; measure single-file-per-worker throughput first.
- **Pre-decompressing files to disk.** Rejected — see decision above.
- **Tuning Cassandra server config** (\`concurrent_writes\`, memtable). MutationStage shows no contention; not worth touching.

## Blocked by

None — \`processed_files\` single-partition fix (#239) is merged.

## Suggested model

**Sonnet.** This is implementation work with a clear contract, in a localized area of the codebase (\`ingester.js\`, \`worker.js\`, two shell scripts, one ADR). No cross-system reasoning; the architectural decision is already made. Save Opus for harder design questions.

tomerp20:
## Summary
- Each worker now opens its own `createReadStream → createGunzip → createInterface` for the dispatched file. Main thread becomes a thin coordinator: queues files chronologically, dispatches one file per idle worker, writes events to Cassandra, and writes `processed_files` / `backfill_progress` in chronological order.
- Persistent worker pool for the whole run — the per-hour `spawn → terminate` pattern is removed.
- Defaults bumped: `INGEST_WORKERS = min(8, cores−2)` (was `min(4, cores/2)`); `UV_THREADPOOL_SIZE=16` in `run-backfill.sh` / `run-hourly.sh` (operator override still wins).

See `docs/adr/0008-ingester-worker-side-decompression.md` for full diagnosis, the trade-offs, and why pre-decompressing to disk was rejected.

## Acceptance criteria

- [x] `worker.js` opens its own `createReadStream → createGunzip → createInterface` for each dispatched file path; performs substring filter, `JSON.parse`, tag/AI extraction; posts extracted events back to main in batches.
- [x] `ingester.js` main thread becomes a thin coordinator: maintains a queue of file paths, dispatches one file per idle worker, awaits per-file completion ack before recording `processed_files`.
- [x] **Persistent worker pool** for the whole run — workers spawn once at startup, terminate once at shutdown. The per-hour `spawn → terminate` pattern is removed.
- [x] **Backpressure:** main thread does NOT dispatch the next file to a worker while events-in-flight to Cassandra exceeds the 10K watermark. Workers idle (do not advance the queue) until Cassandra drains below the watermark.
- [x] **Per-file completion contract:** worker posts `{ type: 'fileDone', hourId, … }` after its final batch. Main thread writes the `processed_files` row for that hour only after all events from that hour have been flushed to Cassandra.
- [x] **Chronological order preserved:** main dispatches files in chronological order; `processed_files` / `backfill_progress` writes happen in input order even though workers may complete files in slightly different wall-clock order (buffered-fileDone strategy via the `nextFinalizeIdx` cursor).
- [x] `INGEST_WORKERS` default ceiling raised from `Math.min(4, cores/2)` to `Math.min(8, cores − 2)` (env override still respected).
- [x] `UV_THREADPOOL_SIZE` set to 16 in `run-backfill.sh` and `run-hourly.sh`.
- [x] **Failure semantics preserved:** if a worker errors on a file, main treats it as a fatal failure; remaining in-flight writes are awaited so the worker pool terminates cleanly with `exitCode=1`.
- [x] **Smoke test transcript** captured — see below. **3.58× wall-clock improvement (155.83s → 43.49s)**, exceeds the ≥ 2× target.
- [x] **New ADR 0008** at `docs/adr/0008-ingester-worker-side-decompression.md`.
- [x] Wiki page `knowledge/wiki/ingestion-pipeline.md` updated to reflect the new worker architecture; ADR 0008 wiki mirror created at `knowledge/wiki/adr-0008-ingester-worker-side-decompression.md`; index updated.

## Smoke test transcript (per ADR 0006)

Ran on the production deployment box (`tomer@192.168.10.12`) against the live single-node Cassandra cluster, branch checked out at `811f25a`.

```
============================================================
Smoke test: feature/ingester-worker-side-decompression
Date range: 2026-02-01 .. 2026-02-05 (120 hour files)
Commit: 811f25a
INGEST_WORKERS default: min(8, cores-2); cores=16
UV_THREADPOOL_SIZE=16
============================================================

=== Wall clock ===
ELAPSED_S=43.49           # baseline: 155.83  → 3.58× speedup
INGEST_EXIT=0

=== backfill_progress (events_written per date) ===
 run_id                               | date       | status    | events_written
--------------------------------------+------------+-----------+----------------
 a2c41940-5740-11f1-a5a1-b5e8ae23f545 | 2026-02-01 | completed |           1539
 a2c41940-5740-11f1-a5a1-b5e8ae23f545 | 2026-02-02 | completed |           2668
 a2c41940-5740-11f1-a5a1-b5e8ae23f545 | 2026-02-03 | completed |           3565
 a2c41940-5740-11f1-a5a1-b5e8ae23f545 | 2026-02-04 | completed |           2904
 a2c41940-5740-11f1-a5a1-b5e8ae23f545 | 2026-02-05 | completed |           3375
Total events written: 14,051   (baseline cited: 13,604 — small drift, similar order of magnitude)

=== nodetool tpstats — Cassandra effectively idle throughout ===
MutationStage                  0      0       (Active=0, Pending=0 — no contention)
Native-Transport-Requests      0      0
(zero dropped messages, zero blocked, zero pending across all stages)

=== vmstat 1 5 (during run) ===
procs -----------memory---------- ---swap-- -----io---- -system-- -------cpu-------
 r  b   swpd   free   buff  cache   si   so    bi    bo   in   cs us sy id wa st gu
10  0    624 705412 124396 8809796    0    0  1515  9173 62706   78  2  2 96  0  0  0
13  2    624 551092 124396 8905588    0    0 95488    32 86856 219024 46  5 47  2  0  0
12  0    624 730856 124404 8818152    0    0 125696    92 88563 237909 74  7 19  0  0  0
14  0    624 680068 124404 8852516    0    0 136248     0 89653 246131 68  6 26  0  0  0
10  1    624 683776 124412 8927748    0    0 75296   200 91747 244699 67  6 26  1  0  0
                                                  ^^^^^                ^^
                                       100-135 MB/s block-input        60-75% user CPU
                                       (HDD streaming gzip files)      (workers + main)
```

Cassandra remains 99.x% idle even with 8 workers in parallel — confirms the issue's prediction that driver/batch tuning is *not* on the critical path for this change.

## Notes for reviewer

- **`top -H` snapshot is essentially empty** in my transcript. I sampled `top -H -p $PID` 3s into the run, but the `pgrep`/PID I had was the parent shell. Worker threads spawn inside the Node child a moment later. The substantive evidence is `vmstat` (60-75% user CPU + 100-135 MB/s block-input vs. the baseline diagnostic showing ~60% on one core only) plus the wall-clock speedup.
- **Per-partition batch buffers in backfill mode** now contain events from multiple hours concurrently. Correctness is preserved because (a) Cassandra dedups via the full primary key on `company_events` and (b) `flushAllPartitionBuffers()` runs at date boundary before `backfill_progress` is written. Worth eyes on the `processHourBatch` finalize/drain logic.
- **Dates still processed serially** in `--mode backfill --verb range` — within a date, hours dispatch concurrently across the pool; across dates we stay serial so a date failure surfaces before later dates' events contaminate state.

## Related
Closes #240
ADR: `docs/adr/0008-ingester-worker-side-decompression.md`


tomerp20:
## Code Review — orchestrator pass

**Security review:** PASS — no findings. File paths flow from the on-disk archive listing (not user input). No secrets hardcoded. Shell scripts use safe `${VAR:-default}` expansion with no injection vectors.

---

### Findings

| Severity | # |
|----------|---|
| HIGH | 1 |
| NIT | 2 |

---

### HIGH: Deadlock when a worker thread hard-crashes

**Location:** `data-pipeline/ingester/ingester.js` — `checkTerminalCondition` (line 214) + `worker.on('error')` (line 329)

**Problem:** `checkTerminalCondition` requires `readyWorkers.length === workerPool.length` before it calls `rejectDone`. When a worker thread hard-crashes (unhandled rejection inside the worker, or any uncaught exception outside the `processFile` try/catch), `worker.on('error')` correctly avoids pushing the crashed worker back into `readyWorkers` — but that permanently reduces the count available in `readyWorkers`. The condition `readyWorkers.length === workerPool.length` can never be satisfied, so the `done` promise never rejects and `processHourBatch` hangs indefinitely.

The `workerError` path (worker posts it after catching inside `processFile`) is fine — it does return the worker to `readyWorkers`. The bug is the hard-crash path only.

**Trace:**
1. Worker crashes → `worker.on('error')` fires
2. `recordWriteError(err)` + `checkTerminalCondition()` — worker NOT pushed back
3. `checkTerminalCondition` sees `writeError`, checks `totalInFlight === 0 && readyWorkers.length === workerPool.length`
4. `readyWorkers.length` is permanently `workerPool.length - 1` → condition never true
5. `done` promise never settles → `await done` in `processHourBatch` hangs

**Fix:** Track busy workers explicitly. A "busy worker" is one that has been dispatched a file but has not yet sent `fileDone`, `workerError`, or crashed. Replace the `readyWorkers.length` pool-size check with `busyWorkers === 0`:

```js
let busyWorkers = 0;

// in tryDispatch, after `worker.postMessage(...)`:
busyWorkers++;

// in fileDone handler, before `readyWorkers.push(worker)`:
busyWorkers--;

// in workerError handler, before `readyWorkers.push(worker)`:
busyWorkers--;

// in worker.on('error'):
busyWorkers--;
// (do NOT push to readyWorkers — worker is gone)

// in checkTerminalCondition, error branch:
if (totalInFlight === 0 && busyWorkers === 0) {
  rejectDone(writeError);
}
```

---

### NIT: `shutdown` message handler in worker.js is dead code

**Location:** `data-pipeline/ingester/worker.js:23-25`

Both `gracefulShutdown` and the `finally` block in main call `w.terminate()` — they never send a `{type:'shutdown'}` postMessage. The handler is harmless but can be removed.

---

### NIT: `?? 1` fallback in in-flight decrement is misleading

**Location:** `data-pipeline/ingester/ingester.js:291, 296`

```js
inFlightByHour.set(hourId, (inFlightByHour.get(hourId) ?? 1) - 1);
```

The key is always present (set to `0` at dispatch on line 207), so the fallback never fires. But `?? 1` (→ subtract to 0) reads as if it's protecting against underflow, which it isn't — if it ever fired it would silently produce the wrong count. A comment or removal would clarify.

---

**Verdict:** The HIGH deadlock needs a fix before this ships. NITs are non-blocking.

tomerp20:
## Code Review — PR #241 (perf: worker-side decompression)

**Orchestrator:** running inline (background session, no custom subagents). Phase 1 of the multi-agent workflow per `CLAUDE.md`.

### Scope reviewed

| Domain | Files |
|---|---|
| Node.js concurrency / worker pool | `data-pipeline/ingester/ingester.js`, `data-pipeline/ingester/worker.js` |
| Bash hygiene | `data-pipeline/scripts/run-backfill.sh`, `data-pipeline/scripts/run-hourly.sh` |
| Security | path traversal in `filePath`, ReDoS on `orgRegexSource`, resource exhaustion |
| Docs | `docs/adr/0008-*`, `knowledge/wiki/*` |

### Findings

#### HIGH

**H1 — Worker pool can deadlock if a worker thread crashes via Node `'error'` event** &nbsp; `ingester.js:217-220, 329-334`

The terminal-success/abort condition is

```js
if (totalInFlight === 0 && readyWorkers.length === workerPool.length) {
  rejectDone(writeError);
}
```

When a Node-level `'error'` fires (worker thread crash — uncaught throw, OOM, etc.), the handler `recordWriteError(err)` is called and the worker is intentionally NOT pushed back to `readyWorkers` (correct — it's dead). But `workerPool.length` is fixed at startup, so `readyWorkers.length === workerPool.length` can never be reached again. `await done` hangs forever, the cron job hangs, and only SIGTERM can recover it.

Fix sketch:

```js
let aliveWorkers = workerPool.length;
// in 'error' handler:
aliveWorkers--;
// terminal check uses aliveWorkers instead of workerPool.length
if (totalInFlight === 0 && readyWorkers.length === aliveWorkers) {
  rejectDone(writeError);
}
```

This is the highest-impact concurrency bug in the diff. It silently degrades production resilience — exactly the kind of "hangs the cron" failure mode catchup work was meant to eliminate per [[adr-0007-processed-files-single-partition]].

#### MEDIUM

**M1 — `tryFinalize` logs "hour complete" in backfill mode while events may still be buffered in `cassandra-writer.js`** &nbsp; `ingester.js:244-253`, `cassandra-writer.js:110-128`

`writer.addBackfillEvent(event)` returns `Promise.resolve()` immediately when the event lands in a partition buffer below threshold. The `inFlightByHour` decrement therefore fires on the next microtask, well before the event has been durably written to Cassandra.

This is **correct** at the date level (because `flushAllPartitionBuffers()` is awaited before `writeBackfillProgress`), but the per-hour `tryFinalize` log line says "hour complete" with `totalEmitted: N` while N events may still be in RAM. Two consequences:

1. Misleading operator log — the hour is *accounted for*, not *durable*.
2. Future-trap — a maintainer adding `await writer.markFileProcessed(h)` to the backfill branch (mirroring the hourly path) would write a durability marker for events that haven't been flushed. This would silently corrupt resume semantics.

Suggested fix: add a comment near `inFlightByHour`'s declaration documenting the asymmetry, **and** change the backfill-mode log message to something like `'hour buffered — durability deferred to date flush'`.

**M2 — `gracefulShutdown` returns exit code 0 on SIGTERM even when work is incomplete** &nbsp; `ingester.js:369-377`

Cron records SIGTERM as success. Acceptable because state is durable (last `backfill_progress` row is the resume point per [[adr-0005-processed-files-hourly-only]]), and the next interval will pick up — but worth being explicit about the "exit 0 on signal is intentional" rationale in a comment.

#### LOW / NIT

**L1 — Dead `shutdown` message handler in worker.js** &nbsp; `worker.js:23-25`

```js
} else if (msg?.type === 'shutdown') {
  process.exit(0);
}
```

Ingester terminates workers via `worker.terminate()`, never sends this message. Either delete or wire up the graceful path (send `shutdown` then `await terminate` with a small timeout).

**L2 — Worker accepts arbitrary `filePath` from main thread without sanity-check** &nbsp; `worker.js:44`

`createReadStream(filePath)` — the worker trusts main. Today, main constructs `filePath` only via `hourIdToPath(GHARCHIVE_DIR, hourId)` with validated `hourId`, so no traversal is possible. Defense in depth: pass `GHARCHIVE_DIR` via `workerData` and assert `path.resolve(filePath).startsWith(path.resolve(GHARCHIVE_DIR))` in the worker.

**L3 — Defensive fallback in `inFlightByHour` decrement can never fire** &nbsp; `ingester.js:291, 296`

```js
inFlightByHour.set(hourId, (inFlightByHour.get(hourId) ?? 1) - 1);
```

Every promise is preceded by `inFlightByHour.set(hourId, +1)` so the key always exists. The `?? 1` fallback is dead code — remove it or comment why it's defensive (otherwise a future reader will wonder if the key is missing in some path).

**L4 — Inconsistent ordering of date defaults in `run-backfill.sh`** &nbsp; `scripts/run-backfill.sh:14, 44`

`BACKFILL_START_DATE` is set BEFORE `.env` sourcing (line 14); `END_DATE` is set AFTER (line 44). Both achieve "operator override wins" via different bash mechanics (`set -a` re-exports for the first; `:-` fallback for the second). Pick one pattern for consistency.

**L5 — Wiki nit: missing `max(1, ...)` lower bound in worker-count description** &nbsp; `knowledge/wiki/ingestion-pipeline.md:24`

Reads "default `min(8, cores-2)`" but code is `Math.max(1, Math.min(8, os.cpus().length - 2))`. On a 1-core dev box `cores-2 = -1`, so the `max(1, ...)` floor matters. Trivial doc fix.

**L6 — `worker.terminate()` does not await pending postMessage flushes**

Acceptable today (workers are idle when terminated in the success path), but worth a one-line comment.

---

### Phase 2 plan

Will fix in Phase 2:
- H1 (worker-crash deadlock) — switch terminal condition to `aliveWorkers` counter
- M1 (backfill-mode finalize semantics) — add clarifying comment + soften log message
- L1 (dead shutdown handler) — delete
- L3 (defensive fallback) — remove or comment
- L4 (script ordering inconsistency) — hoist `END_DATE` before `.env` for consistency
- L5 (wiki nit) — update wiki text

Skipping (out of scope / risk to other features):
- L2 (defense-in-depth path check) — requires plumbing `GHARCHIVE_DIR` into `workerData`; useful but not urgent given current call sites are safe.
- M2 (signal exit-code semantics) — behavior change has cron implications; documenting only.

Posting inline comments next, then implementing fixes.

tomerp20:
**Fix pushed** — commit `db32b71`.

Introduced a `busyWorkers` counter (incremented at dispatch, decremented in all three return paths: `fileDone`, `workerError`, and `worker.on('error')`). `checkTerminalCondition` now tests `busyWorkers === 0` instead of `readyWorkers.length === workerPool.length` — a hard-crashed worker correctly decrements the counter without being returned to the pool, so the deadlock condition is impossible.

NITs left open as non-blocking.

---

**Knowledge-base proposals from this review:**

- `knowledge/wiki/ingestion-pipeline.md` Surprises/gotchas: add bullet — "Worker hard-crash (unhandled rejection outside `processFile`'s try/catch) takes the busy-worker slot permanently out of the pool. The `busyWorkers` counter (not `readyWorkers.length`) is what makes `checkTerminalCondition` settle correctly in that path."
- `knowledge/learnings.md`: "When a worker pool uses a `readyWorkers` queue to track idle threads, pool-size quorum checks (`readyWorkers.length === pool.length`) deadlock if any worker dies without being returned to the queue. Prefer a `busyWorkers` counter that decrements on every exit path — including crash."

data-pipeline/ingester/ingester.js
```diff
@@ -108,204 +111,247 @@ if (cmd.verb === 'hour') {
...
-          writeError = err;
-          logger.error({ event_id: event.event_id, partition: `${event.company}/${event.year_month}`, err: err.message }, 'write failed after retries');
-        });
+  function checkTerminalCondition() {
+    if (writeError) {
+      // Wait until all in-flight writes settle AND every busy worker has either
+      // emitted fileDone or workerError (i.e. is back in readyWorkers).
+      if (totalInFlight === 0 && readyWorkers.length === workerPool.length) {
+        rejectDone(writeError);
       }
```
tomerp20, line 236: **HIGH (H1) — Deadlock on Node `'error'` event.** `workerPool.length` is fixed at startup, but a worker thread that crashes via the Node `'error'` event (uncaught throw, OOM, etc.) is intentionally NOT re-added to `readyWorkers` (see line ~332). Once that happens, `readyWorkers.length === workerPool.length` can never hold again and the `done` promise never settles — `await done` hangs forever, and the cron job hangs until SIGTERM.

Fix: track `aliveWorkers` separately and compare to that instead of `workerPool.length`.

data-pipeline/ingester/ingester.js
```diff
@@ -108,204 +111,247 @@ if (cmd.verb === 'hour') {
...
+        while (nextFinalizeIdx < orderedHourIds.length) {
+          const h = orderedHourIds[nextFinalizeIdx];
+          if (!dispatchedAll.has(h)) break;
+          if ((inFlightByHour.get(h) ?? 0) > 0) break;
+          if (writeError) break;
+
+          const stats = statsByHour.get(h) ?? { totalEmitted: 0, droppedNoTimestamp: 0, droppedNoId: 0 };
+          totalEventsEmittedAcrossHours += stats.totalEmitted;
+          logger.info(
+            { hourId: h, totalEmitted: stats.totalEmitted, droppedNoTimestamp: stats.droppedNoTimestamp, droppedNoId: stats.droppedNoId },
```
tomerp20, line 267: **MEDIUM (M1) — Misleading log + future-trap.** In backfill mode, `writer.addBackfillEvent` returns `Promise.resolve()` as soon as the event lands in a partition buffer below the flush threshold. So `inFlightByHour` drops to 0 *before* the events are durably in Cassandra. Logging `'hour complete'` here overstates durability.

This is correct at the *date* level because `flushAllPartitionBuffers` is awaited before `writeBackfillProgress`, but a future maintainer who adds `markFileProcessed` to the backfill branch (mirroring the hourly path) will silently corrupt resume semantics.

Suggested: (a) comment the asymmetry near `inFlightByHour`'s declaration, (b) change the backfill-mode log to `'hour buffered — durability deferred to date flush'`.

data-pipeline/ingester/ingester.js
```diff
@@ -108,204 +111,247 @@ if (cmd.verb === 'hour') {
...
+        if (writeError) return; // drop further events once we've started failing
+        const { hourId, results } = msg;
+        for (const event of results) {
+          inFlightByHour.set(hourId, (inFlightByHour.get(hourId) ?? 0) + 1);
+          totalInFlight++;
+          const p = cmd.mode === 'backfill'
+            ? writer.addBackfillEvent(event)
+            : writer.writeEvent(event);
+          p.then(() => {
+            inFlightByHour.set(hourId, (inFlightByHour.get(hourId) ?? 1) - 1);
```
tomerp20: **LOW (L3) — Dead defensive code.** Every path that creates a promise here is preceded by `inFlightByHour.set(hourId, current + 1)` (line 285), so the key always exists by the time `.then()` or `.catch()` fires. The `?? 1` fallback is unreachable. Either delete it or add a comment explaining what scenario it defends against — otherwise a future reader will assume the map can be empty here.

data-pipeline/ingester/ingester.js
```diff
@@ -108,204 +111,247 @@ if (cmd.verb === 'hour') {
...
-        dispatchBatch(batch);
-        batch = [];
-      }
-      resolve();
+    worker.on('error', (err) => {
+      logger.error({ err: err.message }, 'worker thread errored');
+      recordWriteError(err);
+      // The worker is gone; do not return it to readyWorkers.
+      checkTerminalCondition();
     });
```
tomerp20, line 358: **HIGH (H1, partner site).** The comment correctly notes the worker is gone — but `workerPool.length` (used in `checkTerminalCondition`) doesn't shrink. Either track `aliveWorkers` here or push a sentinel to `readyWorkers` so the terminal check still fires. See H1 in the summary comment.

data-pipeline/ingester/ingester.js
```diff
@@ -315,18 +361,18 @@ function spawnWorkers(count) {
...
 async function gracefulShutdown(signal) {
   logger.info({ signal }, 'shutting down');
-  await writer.shutdown();
+  try {
+    await Promise.all(workerPool.map(w => w.terminate().catch(() => undefined)));
+  } finally {
+    await writer.shutdown();
+  }
   process.exit(0);
 }
```
tomerp20, line 401: **MEDIUM (M2) — `exit(0)` on SIGTERM.** Cron records the run as success even if work was incomplete. Acceptable given durable state (next interval resumes via `backfill_progress` / `processed_files`), but it would be clearer to either exit non-zero on signal or comment that the success exit code is intentional and relies on idempotent resume.

data-pipeline/ingester/worker.js
```diff
@@ -10,53 +13,96 @@ const orgRegex = new RegExp(orgRegexSource);
...
+// Batch extracted events back to main in chunks. 500 keeps postMessage overhead low
+// without ballooning main-thread Cassandra dispatch latency.
+const EVENT_BATCH_SIZE = 500;
+
+parentPort.on('message', async (msg) => {
+  if (msg?.type === 'processFile') {
+    await processFile(msg.hourId, msg.filePath);
+  } else if (msg?.type === 'shutdown') {
+    process.exit(0);
+  }
```
tomerp20, line 26: **LOW (L1) — Dead code.** Ingester always calls `worker.terminate()` directly (`ingester.js:158, 372`); the `'shutdown'` message is never sent. Either delete this branch or wire the ingester to send `shutdown` first and `terminate` after a short grace period.

data-pipeline/ingester/worker.js
```diff
@@ -10,53 +13,96 @@ const orgRegex = new RegExp(orgRegexSource);
...
+    if (batch.length === 0) return;
+    parentPort.postMessage({ type: 'events', hourId, results: batch });
+    batch = [];
+  }
+
+  let fileStream;
+  let gunzip;
+  let rl;
+  try {
+    fileStream = createReadStream(filePath);
```
tomerp20, line 45: **LOW (L2) — Defense-in-depth: worker trusts `filePath` from main.** All current call sites construct `filePath` via `hourIdToPath(GHARCHIVE_DIR, validated_hourId)`, so there is no exploit today. But the worker has no boundary check that `filePath` is under `GHARCHIVE_DIR`. Passing `GHARCHIVE_DIR` via `workerData` and asserting `path.resolve(filePath).startsWith(path.resolve(GHARCHIVE_DIR))` would harden against future call sites that don't go through `hourIdToPath`.

data-pipeline/scripts/run-backfill.sh
```diff
@@ -43,5 +43,9 @@ cd "${REPO_ROOT}"
 # Keep this assignment here (not hoisted) so that ordering invariant holds.
 END_DATE="${END_DATE:-$(date -u -d 'yesterday' +%Y-%m-%d)}"
```
tomerp20, line 45: **LOW (L4) — Style inconsistency.** `BACKFILL_START_DATE` (line 14) is set BEFORE `.env` is sourced, relying on `set -a` to overwrite. `END_DATE` is set AFTER `.env`, relying on `:-` fallback. Both work, but pick one pattern. Hoisting `END_DATE` above `.env` (and trusting `set -a` symmetrically) reads more cleanly.

knowledge/wiki/ingestion-pipeline.md
```diff
@@ -21,7 +21,7 @@ The data-acquisition + ingestion subsystem of the [[cassandra-analytics-pipeline
 ## Pieces
 
 - **Fetcher** — Node.js. Downloads GH Archive `.json.gz` files, atomic-writes to `GHARCHIVE_DIR` on the HDD (`/mnt/hdd/gharchive/YYYY/MM/DD/HH.json.gz`).
-- **Ingester** — Node.js. Main thread reads a file; worker pool parses + filters by `target_companies`; main thread writes to Cassandra. Branches on `--mode hourly | backfill` (see [[adr-0004-microservices-shaped-cli-contract]]).
+- **Ingester** — Node.js. Persistent worker pool (one per `INGEST_WORKERS`, default `min(8, cores-2)`) — each worker opens its own `createReadStream → createGunzip → createInterface` for the dispatched file, applies the `target_companies` substring filter, parses JSON, extracts tags + AI flag, and posts extracted events back to the main thread in 500-event batches. Main thread is a thin coordinator: queues files chronologically, dispatches one file per idle worker, writes events to Cassandra, and writes `processed_files` (hourly) / `backfill_progress` (backfill) in chronological order. Branches on `--mode hourly | backfill` (see [[adr-0004-microservices-shaped-cli-contract]] and [[adr-0008-ingester-worker-side-decompression]]).
```
tomerp20: **NIT (L5) — Missing `max(1, ...)` lower bound.** Text reads `min(8, cores-2)` but `ingester.js:23` is `Math.max(1, Math.min(8, os.cpus().length - 2))`. On a 1-core dev box `cores - 2 = -1`, so the floor matters. Suggest `Math.max(1, Math.min(8, cores-2))` in prose.

tomerp20 (commented):
Inline findings for PR #241 (orchestrator inline review).

tomerp20:
## Phase 3 — Final review status

### Summary table

| Reviewer | Issues found | Fixed | Skipped |
|---|---|---|---|
| Node.js concurrency (inline, orchestrator) | 4 (1 HIGH, 2 MEDIUM, 1 LOW) | 4 | 0 |
| Bash hygiene (inline) | 1 LOW (style consistency) | 1 | 0 |
| Security (inline, OWASP/path/ReDoS) | 1 LOW (defense-in-depth `filePath` check) | 0 | 1 |
| Docs (inline) | 1 NIT | 1 | 0 |
| **Totals** | **7** | **6** | **1** |

Fix commits on this branch:

- `db32b71` — fix(data-pipeline): prevent deadlock when worker thread hard-crashes — **H1** (pre-existing, already pushed before review ran)
- `51059cb` — fix(review): address PR #241 code-review findings — **M1, L1, L3, L4, L5**

### Skipped issues (need human attention)

**L2 — Worker accepts arbitrary `filePath` from main thread without validating containment under `GHARCHIVE_DIR`** &nbsp; `data-pipeline/ingester/worker.js:44`

Not exploitable today — every call site constructs `filePath` via `hourIdToPath(GHARCHIVE_DIR, validated_hourId)`. Hardening would require plumbing `GHARCHIVE_DIR` into `workerData` and asserting `path.resolve(filePath).startsWith(path.resolve(GHARCHIVE_DIR))` inside `processFile`. Skipped because the orchestrator considered it a future-proofing change unrelated to the PR's stated goal, and wanted to avoid expanding scope. Worth a follow-up issue.

**M2 — `gracefulShutdown` exits 0 on SIGTERM** &nbsp; `data-pipeline/ingester/ingester.js:376`

Documented in Phase 1 review comment. Not fixed: changing the exit code is a behavioural change with cron semantics implications (would cause cron to record a "failure" on every clean shutdown). Resume is idempotent via [[adr-0005-processed-files-hourly-only]] and [[adr-0007-processed-files-single-partition]]. Either accept as-is or document explicitly via an ADR amendment.

### Verification

- `node --check data-pipeline/ingester/ingester.js` — passes
- `node --check data-pipeline/ingester/worker.js` — passes
- `bash -n data-pipeline/scripts/run-backfill.sh` — passes
- `bash -n data-pipeline/scripts/run-hourly.sh` — passes (untouched in fix commit)
- ADR/wiki consistency preserved: all references to [[adr-0008-ingester-worker-side-decompression]] still resolve; the `INGEST_WORKERS` description now matches `ingester.js:23`.

### Knowledge-base proposals from this review

Proposed (not applied — orchestrator does not edit `knowledge/learnings.md` or `wiki/` without human confirmation per `knowledge/CLAUDE.md`):

1. **Add a learnings entry on terminal-condition counters in worker-pool code.**

   > **Worker-pool terminal conditions must use a counter of dispatched-but-not-returned workers, not `readyWorkers.length === poolSize`.** A worker that hard-crashes via Node `'error'` is never returned to the ready queue, so checking against the pool's static size deadlocks. The fix in `db32b71` introduced `busyWorkers` (incremented on dispatch, decremented on every return path: `fileDone`, `workerError`, and `'error'`). → When designing any worker-pool termination check, enumerate every code path that removes a worker from "busy" status; a missing decrement is a silent deadlock.

2. **Add a wiki gotcha on durability semantics of `writeEvent` vs `addBackfillEvent`.**

   Suggest extending `knowledge/wiki/ingestion-pipeline.md` "Surprises / gotchas" with:

   > `writer.writeEvent` (hourly mode) returns a promise that resolves on Cassandra ACK. `writer.addBackfillEvent` (backfill mode) returns `Promise.resolve()` as soon as the event lands in a per-partition buffer below the flush threshold — it does **not** signal durability. The per-hour finalize in `processHourBatch` therefore means "events accounted for" in backfill mode, not "events durable". Durability is enforced at the date boundary by `flushAllPartitionBuffers` + `writeBackfillProgress`. Any future per-hour durability marker for backfill MUST first await a flush.

3. **Open question worth tracking.** Should `gracefulShutdown` exit non-zero on signal? Current `exit(0)` makes cron oblivious to interrupted runs. Could be a `q-shutdown-exit-code.md` open-question page or a sentence under the [[ingestion-pipeline]] "Surprises" section.

### Review-workflow note

This review ran fully inline (no `general-purpose` subagent was spawned) because background sessions in this harness cannot invoke custom or general-purpose subagent types. The orchestrator-as-reviewer pattern worked for this PR because all changes are scoped to `data-pipeline/` (Node concurrency + bash + docs) and the reviewer was able to load the full file context (~500 lines of changed code + 4 ADRs + 3 wiki pages) into its own context window. For larger or cross-stack PRs the parallel-subagent path remains preferred.

---

*Phase 1 inline comments: see [review #4352985886](https://github.com/tomerp20/jobflow/pull/241#pullrequestreview-4352985886).*
*Phase 1 summary comment: see [comment #4529044160](https://github.com/tomerp20/jobflow/pull/241#issuecomment-4529044160).*

PR #249, opened 2026-05-24
Company Scout slice 4: HttpShimCompanyRegistry (production write path)

tomerp20:
## Summary

- Adds `HttpShimCompanyRegistry` — the production `CompanyRegistry` implementation that POSTs accepted active Orgs to the analytics-pipeline HTTPS shim (one POST per Company, per ADR 0010)
- Bearer-token auth via `COMPANY_REGISTRY_TOKEN`; 3 s timeout; per-Org `registered`/`already_exists` outcomes logged at `info`; network/4xx/5xx errors logged at `warn` and never propagated
- Composition-root selection: `resolveRegistry()` uses `HttpShimCompanyRegistry` when `COMPANY_REGISTRY_URL` is set; falls back to `LoggingCompanyRegistry` in dev

## Acceptance criteria

- [x] `HttpShimCompanyRegistry` implements the `CompanyRegistry` interface and is used in production
- [x] One POST per Company; request body matches the wire contract in PRD #180 ("CompanyRegistry seam and the HTTPS shim") and ADR 0010
- [x] Bearer token from `COMPANY_REGISTRY_TOKEN` is included in the `Authorization` header
- [x] Per-Org `registered` outcomes are logged with the Company name, Org login, and `last_repo_push` (one line per Org)
- [x] Per-Org `already_exists` outcomes are logged with the Company name and Org login (one line per Org)
- [x] Network error / 4xx / 5xx responses are caught, logged at `warn`, and never delay or fail card creation
- [x] In development with `COMPANY_REGISTRY_URL` unset, behaviour is unchanged from slice #183 (uses `LoggingCompanyRegistry`); in production, `HttpShimCompanyRegistry` is used
- [x] No new database fields, no migration, no Notification, no user-facing change

## Stacked PR — do not merge out of order

This PR is the **top of a 4-PR stack**. All four must be merged in order before this one lands:

```
#244 (slice 1 — CompanyRegistry seam)
  → #247 (slice 2 — OrgScorer + GitHub client)
    → #248 (slice 3 — Org Resolver + Active-Presence Classifier; base of this branch)
      → THIS PR (slice 4 — HttpShimCompanyRegistry)
```

**Do not merge this PR until #244, #247, and #248 are merged first.**

## Shim-blocked — do not merge until the shim is deployed

[...]

Issue #180, opened 2026-05-22
Company Scout: auto-detect a new Company's active GitHub Orgs on first sighting


tomerp20:
## Problem Statement

When I add a new job application to JobFlow, the company on it is just a name on a card. The JobFlow Analytics system (see `docs/CassandraPlan.md`, `knowledge/wiki/cassandra-analytics-pipeline.md`) builds a per-Company technical profile from GitHub Archive data — but it only profiles Companies whose `(Company, Org)` rows exist in `jobflow.companies`, and that table has historically been hand-curated to a fixed list of ~5 target companies that is never updated. Every time I start pursuing a new Company, I have to remember to register it in the analytics system, and I have to figure out its GitHub Org(s) myself — companies frequently own multiple Orgs (Wix → `wix`, `wix-incubator`, …) and a Company's name rarely equals its Org slug. In practice this never happens, so the analytics system stays stuck on the same handful of Companies and never reflects the ones I'm actually interviewing with.

## Solution

JobFlow automatically notices the first time I track a given Company and, in the background, works out which public GitHub Orgs belong to that Company — including its sibling Orgs — and which of those Orgs are actively producing public code. Active Orgs are written into the analytics pipeline's `jobflow.companies` table with `initialized = false`, so the nightly Backfill picks them up and the Hourly Ingest takes over once Backfill flips the flag (the ADR 0003 relay race).

This is introduced as a new automated process, the **Company Scout** (sibling of the existing Email Agent). It runs on the **First Sighting** of a **Company**, enumerates candidate Orgs, scores each candidate against a weighted rubric (the **OrgScorer** module — see ADR 0009), classifies each accepted Org's **Active GitHub Presence** independently, and registers the active accepted Orgs into the analytics system via a small HTTPS shim that runs alongside Cassandra (see ADR 0010). The Scout is entirely invisible in the UI — no notifications, no new screens — and adds no JobFlow-side tables or columns; the only Cassandra writes go through the shim, and the shim is the sole code that sets `initialized = false`.

## User Stories

1. As a job seeker, I want JobFlow to notice automatically the first time I track a Company, so that I never have to manually register Companies anywhere.
2. As a job seeker, I want this to work whether I added the Application myself or the Email Agent auto-created it, so that no Company slips through depending on how it entered the system.
3. As a job seeker, I want a Company to be checked only the first time it is ever seen, so that adding a second or third Application for the same Company does no redundant work.
[...]

PR #250, opened 2026-05-24
Cassandra write shim: HTTPS endpoint on the Linux box for Company Scout writes

### LINKED ISSUE #246: Cassandra write shim: HTTPS endpoint on the Linux box for Company Scout writes
## Context

The Company Scout (PRD #180) writes to the analytics pipeline's `jobflow.companies` Cassandra table via a tiny HTTPS shim that runs alongside Cassandra on the Xubuntu box. This issue covers building and deploying that shim — the **production write path** that JobFlow's `HttpShimCompanyRegistry` (slice #184) targets.

Canonical decision: **ADR 0010 — Company Scout writes to Cassandra via a tiny HTTPS shim, not direct CQL** (`docs/adr/0010-https-shim-write-path.md`).
Related: ADR 0003 (the `initialized = false` invariant the shim enforces), wiki page `knowledge/wiki/company-scout.md`.

## What to build

A small Node service in `data-pipeline/` (~30 lines of business logic plus deploy plumbing), running on the Xubuntu box at `tomer@192.168.10.12`, fronted by HTTPS via Caddy or nginx with a Let's Encrypt cert. One endpoint, bearer-token auth, idempotent writes to Cassandra.

### Wire contract (must match JobFlow's `HttpShimCompanyRegistry` exactly — slice #184)

```
POST /companies
Authorization: Bearer <shared-secret>
Content-Type: application/json

{
  "company": "wix",
  "active_orgs": [
    { "org_name": "wix",           "last_repo_push": "2026-05-20T..." },
    { "org_name": "wix-incubator", "last_repo_push": "2026-04-18T..." }
  ]
}

→ 200 OK
{
  "company": "wix",
  "results": [
    { "org_name": "wix",           "status": "registered" },
    { "org_name": "wix-incubator", "status": "already_exists" }
  ]
}
```

### Cassandra write — idempotent, with the ADR 0003 invariant baked in

Per Org in the request body, the shim issues:

```sql
INSERT INTO jobflow.companies (company, org_name, added_at, active, initialized)
VALUES (?, ?, toTimestamp(now()), true, false)
IF NOT EXISTS;
```

- `INSERT … IF NOT EXISTS` is a Cassandra Lightweight Transaction — atomic check-and-insert, no read-before-write race.
- `[applied] = true` → response status `registered`.
- `[applied] = false` → response status `already_exists` (a row already exists for `(company, org_name)`; the shim does NOT overwrite — that would silently regress `initialized` from `true` back to `false` and break the ADR 0003 relay race).
- **The shim is the sole writer of `initialized = false`.** No other code anywhere should set this column.

## Acceptance criteria

- [ ] HTTPS-only — Caddy or nginx in front with a valid Let's Encrypt cert; Cassandra's native port (9042) stays bound to localhost
- [ ] `Authorization: Bearer <shared-secret>` required on every request; absent/incorrect token → `401`
- [ ] Request body validation: `company` (non-empty string), `active_orgs` (array of `{ org_name, last_repo_push }`); malformed → `400`
- [ ] Per-Org `INSERT … IF NOT EXISTS` — never an unconditional INSERT, never an UPDATE
- [ ] Response is `200 OK` for any well-formed authenticated request, with a per-Org `results` array reporting `registered` or `already_exists`
- [ ] 5xx returned only for Cassandra connectivity / write errors; auth and validation are 4xx
- [ ] Running as a `systemd` service with auto-restart on failure
- [ ] Bound to localhost behind the reverse proxy; not directly internet-exposed
- [ ] One operational doc page (where in `data-pipeline/`?) covering: deploy steps, env vars, how to rotate the shared secret, how to verify it's running, how to roll back
- [ ] End-to-end smoke test from outside the LAN: POST a synthetic Company → confirm row in `jobflow.companies` with `initialized = false`; re-POST the same body → confirm `already_exists` and no row change

## Coordination

- JobFlow's `HttpShimCompanyRegistry` (slice #184, issue #184) targets exactly this wire contract. **If you change the contract here, the JobFlow side must change in lockstep** (slice #184's PR must merge after the shim is deployed and the contract is confirmed honored).
- The shim is needed in production before slice #184's PR can be merged; until then JobFlow falls back to `LoggingCompanyRegistry` automatically (slices #181 and #183 use this).

## Out of scope

- Multi-tenancy, rate limiting, observability beyond basic logs — analytics writes are sparse; v1 doesn't need this.
- Schema migrations on `jobflow.companies` — the existing columns (`company, org_name, added_at, active, initialized, initialized_at`) cover everything.
- Any read endpoints. The shim is write-only; reads happen inside the ingestion pipeline.

## References

- ADR 0010 — `docs/adr/0010-https-shim-write-path.md` (the architectural rationale and rejected alternatives)
- ADR 0003 — `docs/adr/0003-backfill-hourly-relay-race.md` (the `initialized` invariant)
- CassandraPlan §4.4 — `docs/CassandraPlan.md` (the `companies` table schema)
- Wiki — `knowledge/wiki/company-scout.md` (producer-side system context)
- Linux deployment topology — see project memory `reference_linux_deployment`

tomerp20:
## Summary
- Adds `data-pipeline/shim/` — a small Node ESM service that exposes `POST /companies` so JobFlow's `HttpShimCompanyRegistry` can register `(company, org)` pairs into `jobflow.companies` without speaking CQL
- Bearer-token auth, request validation, per-org `INSERT … IF NOT EXISTS` (Cassandra LWT) — shim is the sole writer of `initialized = false`, enforcing the ADR 0003 relay-race invariant
- Full deploy plumbing: systemd unit, Caddy reverse-proxy config (Let's Encrypt), `.env.example`, and a RUNBOOK.md covering deploy, secret rotation, smoke-test, rollback, and log locations

## Acceptance criteria
- [x] HTTPS-only — Caddy in front with automatic Let's Encrypt cert; Cassandra port 9042 stays on localhost
- [x] `Authorization: Bearer <shared-secret>` required; absent/incorrect → `401`
- [x] Body validation: `company` (non-empty string), `active_orgs` (array of `{ org_name, last_repo_push }`); malformed → `400`
- [x] Per-org `INSERT … IF NOT EXISTS` — never unconditional INSERT, never UPDATE
- [x] `200 OK` for any well-formed authenticated request with per-org `registered` / `already_exists` results
- [x] 5xx only for Cassandra connectivity/write errors; auth + validation are 4xx
- [x] Running as a `systemd` service with auto-restart on failure
- [x] Bound to localhost behind the reverse proxy; not directly internet-exposed
- [x] Operational doc (RUNBOOK.md): deploy steps, env vars, secret rotation, how to verify, rollback, log locations
- [x] End-to-end smoke-test script in RUNBOOK.md: POST synthetic Company → confirm `initialized = false` row; re-POST → confirm `already_exists` and no row change

## Deploy note

Actual deployment to the Xubuntu box (`tomer@192.168.10.12`) is a **manual follow-up** — SSH access from CI is not available. Follow `data-pipeline/shim/RUNBOOK.md`.

## Related
Closes #246

### DIFF
--- a/data-pipeline/shim/.env.example
+++ b/data-pipeline/shim/.env.example
@@ -1,10 +1,17 @@
 # Copy to .env and fill in real values before starting the shim.
 # This file is committed; .env is gitignored.
+#
+# After copying, lock down the real file so only the service user can read it:
+#   chmod 600 /opt/jobflow/data-pipeline/shim/.env
+#
+# Never commit the real .env. Never paste real token values into chat,
+# screenshots, or commit messages.
 
 # Port the shim binds on localhost (reverse-proxy forwards to this)
 SHIM_PORT=3100
 
 # Shared secret — must match COMPANY_REGISTRY_TOKEN in JobFlow's Render env vars.
+# Generate a fresh value with: openssl rand -hex 32
 # Rotate by updating this value and restarting the shim, then updating Render.
 SHIM_BEARER_TOKEN=change-me
 
--- a/data-pipeline/shim/.env.example
+++ b/data-pipeline/shim/.env.example
@@ -1,0 +1,18 @@
+# Copy to .env and fill in real values before starting the shim.
+# This file is committed; .env is gitignored.
+
+# Port the shim binds on localhost (reverse-proxy forwards to this)
+SHIM_PORT=3100
+
+# Shared secret — must match COMPANY_REGISTRY_TOKEN in JobFlow's Render env vars.
+# Rotate by updating this value and restarting the shim, then updating Render.
+SHIM_BEARER_TOKEN=change-me
+
+# Cassandra connection — same values as ingester/.env
+CASSANDRA_CONTACT_POINTS=127.0.0.1
+CASSANDRA_LOCAL_DC=datacenter1
+CASSANDRA_KEYSPACE=jobflow
+
+# Optional
+LOG_LEVEL=info
+NODE_ENV=production
--- a/data-pipeline/shim/RUNBOOK.md
+++ b/data-pipeline/shim/RUNBOOK.md
@@ -20,6 +20,11 @@ Render (JobFlow)
 
 Cassandra's native port (9042) is never reachable from outside the box.
 
+**Process model:** Cassandra runs as a Docker container managed by `docker compose`
+(see `data-pipeline/SETUP.md`); the shim runs as a *native* systemd service
+(`jobflow-shim`), not in a container. Both processes are local to this box and
+communicate over `127.0.0.1` — no container/host networking concerns.
+
 ---
 
 ## Deploy steps (first-time)
@@ -45,7 +50,13 @@ npm install --omit=dev
 cp /opt/jobflow/data-pipeline/shim/.env.example /opt/jobflow/data-pipeline/shim/.env
 # Edit .env: set SHIM_BEARER_TOKEN and verify Cassandra vars
 nano /opt/jobflow/data-pipeline/shim/.env
+
+# Lock the env file — it contains SHIM_BEARER_TOKEN.
+sudo chown tomer:tomer /opt/jobflow/data-pipeline/shim/.env
+chmod 600 /opt/jobflow/data-pipeline/shim/.env
 ```
+
+Generate a strong token: `openssl rand -hex 32`.
 
 `SHIM_BEARER_TOKEN` must match `COMPANY_REGISTRY_TOKEN` in JobFlow's Render
 environment variables.
@@ -106,9 +117,12 @@ at reload time). Check: `sudo journalctl -u caddy -n 30`.
 5. Trigger a Render redeploy (env var change auto-triggers it)
 
 Both sides must be updated atomically in terms of effect — there is a brief
-window while Render is restarting where Scout writes will fail (JobFlow returns
-a log line from `LoggingCompanyRegistry`). This is acceptable for the v1
-write volume.
+window while Render is restarting where Scout writes will fail. **The Company
+Scout has no retry**: per `knowledge/wiki/company-scout.md` (Surprises /
+gotchas), any First Sighting that hits the shim during this window is silently
+un-registered until the same Company is sighted again. For v1 the write volume
+is sparse enough that this gap is acceptable; if it becomes a real problem,
+the documented future fix is a manual re-trigger admin endpoint.
 
 ---
 
@@ -124,6 +138,10 @@ sudo journalctl -u jobflow-shim -f
 # Confirm it is bound on localhost only
 ss -tlnp | grep 3100
 # expected: 127.0.0.1:3100 — NOT 0.0.0.0:3100
+
+# Unauthenticated liveness check via Caddy (no bearer required)
+curl -fsS https://YOUR_DOMAIN/health
+# expected: {"status":"ok"}
 ```
 
 ---
@@ -181,6 +199,51 @@ docker exec -it jobflow-cassandra cqlsh -e \
 # ── Step 5: Clean up the test row ────────────────────────────────────────────
 docker exec -it jobflow-cassandra cqlsh -e \
   "DELETE FROM jobflow.companies WHERE company = 'smoke-test-co' AND org_name = 'smoke-test-org';"
+```
+
+### Negative-path smoke checks
+
+The shim should reject malformed or unauthorised requests cleanly. Run these
+once after a fresh deploy to confirm the auth, validation, and method-routing
+paths all behave:
+
+```bash
+# 401 — wrong token
+curl -sS -o /dev/null -w "%{http_code}\n" -X POST "${SHIM_URL}/companies" \
+  -H "Authorization: Bearer wrong-token" \
+  -H "Content-Type: application/json" \
+  -d '{"company":"x","active_orgs":[{"org_name":"y","last_repo_push":"2026-01-01T00:00:00Z"}]}'
+# expected: 401
+
+# 400 — invalid JSON body
+curl -sS -o /dev/null -w "%{http_code}\n" -X POST "${SHIM_URL}/companies" \
+  -H "Authorization: Bearer ${TOKEN}" \
+  -H "Content-Type: application/json" \
+  -d 'not-json'
+# expected: 400
+
+# 400 — missing required field (last_repo_push)
+curl -sS -o /dev/null -w "%{http_code}\n" -X POST "${SHIM_URL}/companies" \
+  -H "Authorization: Bearer ${TOKEN}" \
+  -H "Content-Type: application/json" \
+  -d '{"company":"x","active_orgs":[{"org_name":"y"}]}'
+# expected: 400
+
+# 415 — wrong Content-Type
+curl -sS -o /dev/null -w "%{http_code}\n" -X POST "${SHIM_URL}/companies" \
+  -H "Authorization: Bearer ${TOKEN}" \
+  -H "Content-Type: text/plain" \
+  -d 'whatever'
+# expected: 415
+
+# 405 — wrong method on /companies (Caddy rejects before reaching shim)
+curl -sS -o /dev/null -w "%{http_code}\n" -X GET "${SHIM_URL}/companies" \
+  -H "Authorization: Bearer ${TOKEN}"
+# expected: 405
+
+# 404 — unknown path
+curl -sS -o /dev/null -w "%{http_code}\n" "${SHIM_URL}/nope"
+# expected: 404
 ```
 
 ---
--- a/data-pipeline/shim/RUNBOOK.md
+++ b/data-pipeline/shim/RUNBOOK.md
@@ -1,0 +1,209 @@
+# JobFlow Cassandra Write Shim — Operator Runbook
+
+The shim is a small Node HTTP service running on the Xubuntu box at
+`tomer@192.168.10.12`. It exposes one authenticated endpoint (`POST /companies`)
+that JobFlow's `HttpShimCompanyRegistry` calls to register `(company, org)` pairs
+into the `jobflow.companies` Cassandra table using `INSERT … IF NOT EXISTS`.
+
+---
+
+## Architecture
+
+```
+Render (JobFlow)
+  └─ HttpShimCompanyRegistry
+       └─ POST https://YOUR_DOMAIN/companies
+            └─ Caddy (TLS termination, Let's Encrypt)
+                 └─ 127.0.0.1:3100 ← shim.js (systemd: jobflow-shim)
+                      └─ 127.0.0.1:9042 ← Cassandra (Docker)
+```
+
+Cassandra's native port (9042) is never reachable from outside the box.
+
+---
+
+## Deploy steps (first-time)
+
+### 1. Prerequisites
+
+- Node ≥ 20.6 on PATH
+- Cassandra running via `docker compose` (see `data-pipeline/SETUP.md`)
+- A public hostname with a DNS A record pointing to the box's WAN IP
+  (the Caddy block in `deploy/Caddyfile` must match this hostname)
+- Port 443 open on the router/firewall; port 3100 is NOT forwarded externally
+
+### 2. Install Node dependencies
+
+```bash
+cd /opt/jobflow/data-pipeline/shim
+npm install --omit=dev
+```
+
+### 3. Create the env file
+
+```bash
+cp /opt/jobflow/data-pipeline/shim/.env.example /opt/jobflow/data-pipeline/shim/.env
+# Edit .env: set SHIM_BEARER_TOKEN and verify Cassandra vars
+nano /opt/jobflow/data-pipeline/shim/.env
+```
+
+`SHIM_BEARER_TOKEN` must match `COMPANY_REGISTRY_TOKEN` in JobFlow's Render
+environment variables.
+
+### 4. Install and start the systemd unit
+
+```bash
+sudo cp /opt/jobflow/data-pipeline/shim/deploy/jobflow-shim.service /etc/systemd/system/
+sudo systemctl daemon-reload
+sudo systemctl enable --now jobflow-shim
+sudo systemctl status jobflow-shim   # should show "Active: active (running)"
+```
+
+### 5. Install Caddy (if not already present)
+
+```bash
+sudo apt install -y debian-keyring debian-archive-keyring apt-transport-https
+curl -1sLf 'https://dl.cloudsmith.io/public/caddy/stable/gpg.key' \
+  | sudo gpg --dearmor -o /usr/share/keyrings/caddy-stable-archive-keyring.gpg
+curl -1sLf 'https://dl.cloudsmith.io/public/caddy/stable/debian.deb.txt' \
+  | sudo tee /etc/apt/sources.list.d/caddy-stable.list
+sudo apt update && sudo apt install caddy
+```
+
+### 6. Install the Caddyfile
+
+```bash
+# Replace YOUR_DOMAIN with the real hostname in the file first
+sudo cp /opt/jobflow/data-pipeline/shim/deploy/Caddyfile /etc/caddy/Caddyfile
+sudo systemctl reload caddy
+```
+
+Caddy will provision a Let's Encrypt cert automatically on first request (or
+at reload time). Check: `sudo journalctl -u caddy -n 30`.
+
+---
+
+## Environment variables
+
+| Variable | Required | Default | Description |
+|---|---|---|---|
+| `SHIM_PORT` | no | `3100` | Port shim binds on localhost |
+| `SHIM_BEARER_TOKEN` | **yes** | — | Shared secret; must match `COMPANY_REGISTRY_TOKEN` in Render |
+| `CASSANDRA_CONTACT_POINTS` | **yes** | — | Comma-separated host list (e.g. `127.0.0.1`) |
+| `CASSANDRA_LOCAL_DC` | **yes** | — | Cassandra local datacenter name (e.g. `datacenter1`) |
+| `CASSANDRA_KEYSPACE` | no | `jobflow` | Cassandra keyspace |
+| `LOG_LEVEL` | no | `info` | Pino log level (`trace`, `debug`, `info`, `warn`, `error`, `fatal`) |
+| `NODE_ENV` | no | — | Set to `production` to disable pino-pretty |
+
+---
+
+## Rotating the shared secret
+
+1. Generate a new secret: `openssl rand -hex 32`
+2. Update `SHIM_BEARER_TOKEN` in `/opt/jobflow/data-pipeline/shim/.env`
+3. Restart the shim: `sudo systemctl restart jobflow-shim`
+4. Update `COMPANY_REGISTRY_TOKEN` in JobFlow's Render environment variables
+5. Trigger a Render redeploy (env var change auto-triggers it)
+
+Both sides must be updated atomically in terms of effect — there is a brief
+window while Render is restarting where Scout writes will fail (JobFlow returns
+a log line from `LoggingCompanyRegistry`). This is acceptable for the v1
+write volume.
+
+---
+
+## Verify it is running
+
+```bash
+# Service status
+sudo systemctl status jobflow-shim
+
+# Tail live logs
+sudo journalctl -u jobflow-shim -f
+
+# Confirm it is bound on localhost only
+ss -tlnp | grep 3100
+# expected: 127.0.0.1:3100 — NOT 0.0.0.0:3100
+```
+
+---
+
+## End-to-end smoke test
+
+Run from any machine with internet access (replace values as needed):
+
+```bash
+SHIM_URL="https://YOUR_DOMAIN"
+TOKEN="your-bearer-token"
+
+# ── Step 1: POST a synthetic Company ─────────────────────────────────────────
+curl -s -X POST "${SHIM_URL}/companies" \
+  -H "Authorization: Bearer ${TOKEN}" \
+  -H "Content-Type: application/json" \
+  -d '{
+    "company": "smoke-test-co",
+    "active_orgs": [
+      { "org_name": "smoke-test-org", "last_repo_push": "2026-05-24T00:00:00Z" }
+    ]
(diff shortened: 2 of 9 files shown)


tomerp20 (commented):
## Phase 1 Review — Infrastructure (review-250-infra)

Scope: `data-pipeline/shim/**` (Node ESM service, lib, systemd unit, Caddyfile, `.env.example`, RUNBOOK.md). No `frontend/**` / `backend/**` changes. Reviewed against [[adr-0010-https-shim-write-path]], [[adr-0003-backfill-hourly-relay-race]], [[adr-0004-microservices-shaped-cli-contract]], and [[company-scout]].

### Summary
- **4 HIGH · 7 MEDIUM · 7 LOW · 4 NIT** findings
- Wire contract matches `HttpShimCompanyRegistry` (`backend/src/services/companyScout/companyRegistry.ts`) and ADR 0010 §Wire-contract.
- The `initialized = false` invariant from [[adr-0003-backfill-hourly-relay-race]] is correctly centralised in `lib/cassandra.js` (single hard-coded `INSERT … IF NOT EXISTS … VALUES (?, ?, toTimestamp(now()), true, false)` — no UPDATE path exists), and the LWT `[applied]` mapping is correct per the bracket-notation learning logged in `knowledge/learnings.md`.
- The bearer token is never logged. Caddy correctly fronts a `127.0.0.1`-bound shim — Cassandra native port stays internal.
- Most findings cluster around (a) hardening that the brief explicitly calls out (timing-safe compare, body-size limits, systemd hardening directives, env-file permissions), and (b) one wire-contract drift: the `last_repo_push` field is validated but silently discarded.

---

### HIGH

- **[`data-pipeline/shim/shim.js:53`] Bearer-token comparison is not constant-time.** `auth !== \`Bearer ${SHIM_BEARER_TOKEN}\`` is a naive `!==` against a string built from the secret; that's a textbook timing-attack side channel. Use `crypto.timingSafeEqual` against equal-length `Buffer`s, with an early length check to avoid `RangeError` (and to keep the length check itself constant-time relative to the secret). This is one of the explicit "security pass concerns" called out for this review, and the ADR 0010 threat model (public-internet exposure of a home-LAN box) is exactly the deployment where a timing oracle matters.

- **[`data-pipeline/shim/shim.js:40-47`] No request-body size limit — OOM / slowloris vector.** `readBody` accumulates `chunks` unbounded; a single attacker can POST a multi-gigabyte body and exhaust shim memory before `JSON.parse` ever runs. The review brief explicitly flags "a 10 MB JSON body should be rejected before parsing". Add a max-bytes guard inside the `data` listener that destroys the request once a small limit (e.g. 64 KB — the wire contract is tiny: one Company + a handful of Orgs) is exceeded, and return `413 Payload Too Large`.

- **[`data-pipeline/shim/deploy/Caddyfile:18-27`] No `request_body { max_size … }` and no HTTP→HTTPS redirect or HSTS at the proxy.** Caddy will redirect HTTP to HTTPS by default for a domain block, but the redirect is a 308 that preserves the body — meaning a misconfigured client can leak the bearer token in plaintext on the first hop. Also there is no `request_body { max_size 64KB }` directive (Caddy default is effectively unlimited for `reverse_proxy`), so the body-size finding above is not defended in depth at the proxy either. The brief lists "TLS-only enforcement at the reverse proxy" and "input size limits on the POST body". Recommend:
  - `request_body { max_size 64KB }` inside the `route @companies` block.
  - Explicit `header Strict-Transport-Security "max-age=31536000"`.
  - A `method POST` matcher on the route so `GET /companies` and `PUT /companies` also get `404`'d at the proxy rather than reaching the shim (defense-in-depth even though the shim itself 404s them).

- **[`data-pipeline/shim/lib/cassandra.js:5-8` + `shim.js:76-78`] `last_repo_push` is validated as required but silently discarded by the INSERT — wire-contract drift.** The shim mandates each `active_orgs` entry have a `last_repo_push` string (`shim.js:76-78`) and ADR 0010's wire example documents it as part of the payload, but `INSERT_CQL` binds only `(company, org_name, toTimestamp(now()), true, false)` — the field never reaches Cassandra. This violates the contract surface of [[adr-0010-https-shim-write-path]]: callers reasonably believe the value they sent is now stored, and it is not. Two acceptable resolutions:
  1. **Persist it.** Add a `last_repo_push timestamp` column to `jobflow.companies` (schema migration `006_*.cql`) and bind it. This matches the spirit of the field's name and gives Backfill / future tooling a real upper bound.
  2. **Drop it from the shim and the ADR.** If the field is purely informational for the JobFlow logs (which is what `LoggingCompanyRegistry` already captures), the shim should not require it — relax validation to "optional, accepted-and-ignored" and update ADR 0010's example, or remove it from both sides. Either way, the current "I demand it and then throw it away" stance creates a permanent quiet data loss.

---

### MEDIUM

- **[`data-pipeline/shim/deploy/jobflow-shim.service:36-40`] systemd hardening incomplete vs the security checklist in the brief.** Present: `RestrictAddressFamilies`, `PrivateTmp`, `NoNewPrivileges`. Missing and recommended for a network-exposed Node service holding a bearer secret:
  - `ProtectSystem=strict` (read-only filesystem except explicit paths)
  - `ProtectHome=true` (no `/home/*` access — the shim only needs `/opt/jobflow/...`)
  - `ProtectKernelTunables=true`, `ProtectKernelModules=true`, `ProtectControlGroups=true`
  - `LockPersonality=true`
  - `RestrictRealtime=true`
  - `RestrictNamespaces=true`
  - `SystemCallArchitectures=native`
  - Consider `CapabilityBoundingSet=` (empty — Node needs none for `:3100`)
  None of these change behaviour for the shim's documented operation; all reduce blast radius if the Node process is compromised.

- **[`data-pipeline/shim/deploy/jobflow-shim.service:17-18`] `After=network.target docker.service` + `Wants=docker.service` is the wrong startup ordering for a network-bound, Cassandra-dependent service.** Two issues: (1) `network.target` does not mean the network is up — use `network-online.target` with `Wants=network-online.target`. (2) `Wants=docker.service` only orders against Docker's daemon startup, not against the Cassandra *container* being ready to accept CQL — `await client.connect()` in `shim.js:105` will throw and the process will exit. `Restart=on-failure` + `RestartSec=5` will recover, but adds boot churn. Consider raising `RestartSec=10` and adding `StartLimitIntervalSec=120` / `StartLimitBurst=10` so a real Cassandra outage trips the failure escalation rather than silently respawning forever. (Existing data-pipeline services on the box are the right reference — confirm consistency.)

- **[`data-pipeline/shim/lib/cassandra.js:34`] Unsafe access on LWT result.** `result.rows[0]['[applied]']` will throw `TypeError: Cannot read properties of undefined` if the driver ever returns an empty `rows` array (e.g. transient retry path, a future driver version, or a corner case where the LWT was applied but the result envelope is malformed). Use defensive access and treat unknown as a write failure (5xx) rather than risking misclassifying as `already_exists`:
  ```js
  const applied = result?.rows?.[0]?.['[applied]'];
  if (applied !== true && applied !== false) {
    throw new Error('LWT response missing [applied] column');
  }
  return applied ? 'registered' : 'already_exists';
  ```
  Without this, a malformed response → `undefined` → falsy → returns `'already_exists'` and the client believes the row is registered when it may not be — a silent break of the ADR 0010 idempotency guarantee.

- **[`data-pipeline/shim/lib/cassandra.js:15-19`] LWT consistency not made explicit.** `consistency: localOne` is fine for the single-DC single-node home box today, but: (a) LWTs have *two* consistency levels — `serialConsistency` (defaults to `serial`, which requires DC-quorum and can hang on a multi-DC box) and `consistency` (the commit CL). For a one-node DC this works; if the deployment topology ever expands to multi-DC, `localOne` for the commit CL paired with `serial` for the LWT is internally inconsistent. Set `serialConsistency: types.consistencies.localSerial` explicitly to document the single-DC assumption and to make a future multi-DC migration noisy rather than silent. Matches the spirit of [[adr-0004-microservices-shaped-cli-contract]]'s "env carries operational config" — making the contract explicit at the code site.

- **[`data-pipeline/shim/lib/cassandra.js:6`] CQL keyspace hardcoded as `jobflow.companies`, ignoring `CASSANDRA_KEYSPACE` env var.** The `Client` is configured with `keyspace: CASSANDRA_KEYSPACE` (`shim.js:102`), but the CQL string is fully qualified to `jobflow.companies`. If an operator changes `CASSANDRA_KEYSPACE`, writes still land in `jobflow.companies` — silent. Either drop `jobflow.` from the CQL (rely on the client's bound keyspace) or remove the env var as a knob and hardcode it in `.env.example`.

- **[`data-pipeline/shim/RUNBOOK.md`] Missing `chmod 600` instruction for `/opt/jobflow/data-pipeline/shim/.env`.** The brief explicitly requires this: "RUNBOOK should specify `chmod 600` for the production env file". The file holds the bearer token; current systemd unit reads it as `User=tomer` but the file mode on disk is not enforced. Add a step in §3 "Create the env file" immediately after `cp`/`nano`: `chmod 600 /opt/jobflow/data-pipeline/shim/.env`. Also consider checking `chown tomer:tomer` to match the systemd `User=`.

- **[`data-pipeline/shim/shim.js:108-117`] No request timeout / headers timeout configured on the HTTP server.** Node's `server.requestTimeout` defaults to 5 minutes; `server.headersTimeout` to 60s. For a service whose entire request body is a few hundred bytes, hold-open slowloris connections can exhaust file descriptors. Set `server.requestTimeout = 10_000` and `server.headersTimeout = 5_000` explicitly.

---

### LOW

- **[`data-pipeline/shim/shim.js:108-117`] No `Content-Type` check before `JSON.parse`.** Accepts and parses any body regardless of `Content-Type`. Cosmetic; the parse error path already returns 400, but rejecting non-JSON content-types early gives clearer operator logs.

- **[`data-pipeline/shim/shim.js:83-92`] Per-org LWT is sequential, not batched.** Each `active_orgs` entry triggers one round-trip to Cassandra. v1 batch sizes are tiny (1-5 orgs per Company) so this is fine, but worth noting. Cassandra LWTs cannot be batched across partitions (different `company` keys), but here all rows share the same partition key (`company`) — `BATCH … APPLY BATCH` of conditional updates on the same partition is supported and would be one round-trip instead of N. Defer until measurements warrant.

- **[`data-pipeline/shim/shim.js:135-138`] `main().catch` calls `process.exit(1)` but the `client` may never have been constructed.** Fine in practice; just noting the error path doesn't try to `client.shutdown()` because client is scoped inside `main()`. No leak risk since process exits.

- **[`data-pipeline/shim/RUNBOOK.md` smoke test] Smoke test only covers happy path + idempotency.** Brief calls for 401 / 400 coverage in the smoke test. Three additional one-liners would close the gap:
  ```bash
  # 401 path
  curl -i -X POST "${SHIM_URL}/companies" -H "Authorization: Bearer wrong" -H "Content-Type: application/json" -d '{"company":"x","active_orgs":[{"org_name":"y","last_repo_push":"2026-01-01"}]}'
  # expect HTTP/1.1 401

  # 400 path — missing field
  curl -i -X POST "${SHIM_URL}/companies" -H "Authorization: Bearer ${TOKEN}" -H "Content-Type: application/json" -d '{}'
  # expect HTTP/1.1 400

  # 404 path — wrong route
  curl -i -X POST "${SHIM_URL}/notfound" -H "Authorization: Bearer ${TOKEN}"
  # expect HTTP/1.1 404
  ```

- **[`data-pipeline/shim/RUNBOOK.md` §Verify it is running] No `/health` endpoint.** Operator verification is "check `ss -tlnp`" — fine, but a `GET /health` returning `{ ok: true }` (no auth) would let Caddy and external monitoring probe liveness cheaply. Trivial addition.

- **[`data-pipeline/shim/RUNBOOK.md` §Rotating the shared secret] Restart→Render-redeploy window not bounded.** "Render is restarting where Scout writes will fail" is documented as acceptable, which matches the "no retry" stance in [[company-scout]] §Surprises. Worth explicit cross-reference so the operator knows this is by design.

- **[`data-pipeline/shim/.env.example:9`] `SHIM_BEARER_TOKEN=change-me` is a placeholder — fine that it's committed.** Recommend strengthening the comment to: "Generate with `openssl rand -hex 32`; never commit a real value." Currently the comment only says "fill in real values".

---

### NIT

- **[`data-pipeline/shim/deploy/jobflow-shim.service:21`] `Type=simple`** — could be `Type=notify` if Node ever calls `sd_notify`, but `simple` is the right default. No action.

- **[`data-pipeline/shim/shim.js:6`] `parseInt(process.env.SHIM_PORT ?? '3100', 10)`** — fine, but `Number()` would coerce more strictly. Cosmetic.

- **[`data-pipeline/shim/deploy/Caddyfile:1`] Caddyfile comment block is extensive and helpful** — keep as-is.

- **[`data-pipeline/shim/RUNBOOK.md` §Architecture]** ASCII diagram is clear. Consider adding a one-line caveat that "Cassandra is in Docker, the shim is not" — currently the systemd unit's `Wants=docker.service` reveals this only to readers who open the unit.

---

### Knowledge-base proposals (informational — orchestrator collates in Phase 3)

- **Add a wiki page `[[shim]]`** under `knowledge/wiki/shim.md` capturing: the wire contract (with a link back to ADR 0010), the `initialized = false` invariant centralisation, the deploy topology (Caddy → 127.0.0.1:3100 → Cassandra), and the rotation procedure. `company-scout.md` currently has the only mention and links out to the ADR rather than to a dedicated systems page. Would match the shape of `[[ingestion-pipeline]]` and `[[hourly-ingest]]`.
- **Append a `learnings.md` bullet** capturing: "Validating a payload field you immediately discard creates silent contract drift" (re: the `last_repo_push` finding above). Cheap nudge for future small services.
- **Consider an ADR amendment to 0010** documenting the systemd hardening baseline expected for any future `data-pipeline/` shim — so the next service-on-the-box reaches the same hardening bar without rediscovery.
- **Note for the brief-mandated re-read:** the wiki has very few pages (`company-scout`, `ingestion-pipeline`, plus three ADR mirrors). Most ADR mirror pages listed in `knowledge/wiki/index.md` do not yet exist as files; the `index.md` "wikilink without page" pattern is acknowledged in `learnings.md` 2026-05-23. Not blocking this PR, but it limited the citation surface for this review — I cite the ADR file paths directly where the wiki mirror is absent.

tomerp20:
## Code Review — Phase 3 Status

**Orchestrator:** main session (`/orchestrate-review 250`)
**Phase 1 reviewer:** `review-250-infra` (general-purpose, opus, worktree-isolated) + inline security pass
**Phase 2 fix commit:** `23eb429`

### Phase 1 — findings

| Severity | Count |
|---|---|
| HIGH | 4 |
| MEDIUM | 7 |
| LOW | 7 |
| NIT | 4 |
| **Total** | **22** |

Full Phase 1 review: https://github.com/tomerp20/jobflow/pull/250#pullrequestreview-4353281914

### Phase 2 — fix outcomes

| Severity | Found | Fixed | Skipped / Deferred |
|---|---|---|---|
| HIGH | 4 | **4** | 0 |
| MEDIUM | 7 | **7** | 0 |
| LOW | 7 | **4** | 3 |
| NIT | 4 | **1** | 3 |

**Skipped with rationale (left for the human to act on if desired):**
- **LOW-13** (per-Org LWT → BATCH optimisation) — explicitly marked by the reviewer as a future-improvement, not a v1 blocker. Defer to a follow-up if Cassandra write latency becomes a concern.
- **NIT-19** (`Type=simple` is fine), **NIT-20** (`parseInt` style), **NIT-21** (Caddyfile comments) — the reviewer's recommendation on each was "no change needed" / "keep as-is". No-ops.

### Phase 2 — fix highlights (full body in commit message of `23eb429`)

- **HIGH-1** constant-time bearer comparison via `crypto.timingSafeEqual`
- **HIGH-2** 64 KB request-body cap at both Caddy and the shim; 413 returned on overrun
- **HIGH-3** Caddyfile gained `request_body { max_size }`, explicit `method`+`path` matchers, HSTS / nosniff / Referrer-Policy headers, `405` for wrong method on `/companies`
- **HIGH-4** `last_repo_push` now appears in the structured `org write` log line with a comment explaining the field is logged-not-stored pending a future schema migration — removes the silent-contract-drift the reviewer cited (the validation requirement stays, so the wire contract is unchanged)
- **MEDIUM-5** full systemd hardening pass (`ProtectSystem=strict`, every `Protect*`, `RestrictRealtime`, `LockPersonality`, `MemoryDenyWriteExecute`, `SystemCallFilter`, empty `CapabilityBoundingSet`, …)
- **MEDIUM-6** `After=network-online.target` (was `network.target`); `Wants=network-online.target`; explanatory comment on why `After=docker.service` is not a real container-readiness guarantee
- **MEDIUM-7** defensive `[applied]` access — throws on missing row / non-boolean rather than silently regressing the [[adr-0003-backfill-hourly-relay-race]] invariant
- **MEDIUM-8** explicit `serialConsistency: localSerial` on the LWT
- **MEDIUM-9** unqualified `INSERT INTO companies` so `CASSANDRA_KEYSPACE` env is honored (was hardcoded `jobflow.companies`)
- **MEDIUM-10** explicit `chown` + `chmod 600` step for `.env` in the RUNBOOK
- **MEDIUM-11** `server.requestTimeout=30s`, `server.headersTimeout=10s` against slowloris
- **LOW-12** `415` on non-`application/json` Content-Type
- **LOW-14** `client` hoisted to module scope so `main().catch()` can shut down on startup failure
- **LOW-15** `RUNBOOK.md` smoke test now covers 401 / 400 / 415 / 405 / 404 negative paths alongside the happy path
- **LOW-16** `GET /health` endpoint added in the shim + Caddyfile passthrough + RUNBOOK Verify step
- **LOW-17** rotation downtime in the RUNBOOK now cross-references the Company Scout's documented no-retry stance
- **LOW-18** `.env.example` recommends `openssl rand -hex 32` and warns against committing real values
- **NIT-22** RUNBOOK Architecture section notes the Cassandra-in-Docker / shim-as-native-systemd-service split

### Security pass (orchestrator, inline)

No additional findings beyond Phase 1. Specifically verified after fixes:
- Bearer token never logged under any path (only `org write` entries with `company`, `org_name`, `last_repo_push`, `status`).
- TLS-only by virtue of Caddy's default HTTP→HTTPS redirect; explicit HSTS header added.
- Input validation rejects malformed JSON (400), non-JSON (415), oversize (413).
- Cassandra calls use prepared statements (`prepare: true`); no string concatenation in CQL.
- `sudo` usage in the RUNBOOK is limited to standard ops (`cp`, `systemctl`, `apt install`).
- Service binds to `127.0.0.1` only; reverse-proxy is the sole public surface.

### What's still on the human

1. Merge in order: this PR (the shim) → after deploy & contract-confirmed → PR #249 (`HttpShimCompanyRegistry`).
2. Deploy is manual per the RUNBOOK (`systemd`, Caddy, Cassandra container — all on the Xubuntu box).
3. Run the smoke checks from the RUNBOOK (happy path + the new 401/400/415/405/404 cases) before flipping JobFlow's `COMPANY_REGISTRY_URL`.
4. Decide if you want the deferred LOW-13 (BATCH writes) opened as a follow-up issue.

Do not merge until the deploy + smoke test succeeds. CLAUDE.md prohibits Claude from merging.

tomerp20:

Per the `/orchestrate-review` skill's KB-integration step. Proposals only — no files edited.

### `knowledge/wiki/` — new and amended pages

**New: `knowledge/wiki/cassandra-write-shim.md`** *(type: `system`)*
The shim is now a real, deployed-style subsystem with operational surface (systemd, Caddy, runbook) — large enough to deserve its own wiki page mirroring the shape of [[ingestion-pipeline]]. Should cover:
- Decision flow (POST in → per-Org LWT out)
- Key invariants (sole writer of `initialized=false`; LWT idempotency; never UPDATE)
- Surprises / gotchas (Caddy doesn't enforce body-size by default; `After=docker.service` ≠ container ready; `[applied]` column name is bracketed; `last_repo_push` is in the wire contract but not yet in schema)
- Source pointers (`data-pipeline/shim/`, `RUNBOOK.md`)
- Reciprocate from [[company-scout]] (add to its `related:` list and "How it relates" section).

**Amend: `knowledge/wiki/company-scout.md`**
Add to "Source pointers": this PR's shim files (`data-pipeline/shim/{shim.js,lib/cassandra.js,deploy/*,RUNBOOK.md}`) and the new `[[cassandra-write-shim]]` page.

### `knowledge/learnings.md` — new bullets

- **Validating a wire-contract field you immediately discard is silent contract drift.** PR #250 required `last_repo_push` in the body but never stored or logged it. The fix (this PR) was to surface the field in the structured log alongside an explanatory comment; the real fix is a schema migration. → Whenever the shim or any producer validates a new field, decide at write time whether it gets stored, logged, or dropped — never let "validated but unused" persist past code review.

- **Caddy's body-size limit is opt-in.** `request_body { max_size }` is not a default; without it, every upstream is exposed to whatever the client sends. → Always add an explicit `request_body { max_size }` block for write endpoints when introducing a new Caddy site config.

- **`systemd` `After=docker.service` orders against the daemon, not container readiness.** The original shim unit relied on `After=docker.service` to wait for Cassandra; the Docker daemon being up is not the same thing as the Cassandra container accepting connections. → When a native systemd service depends on a Docker container, either add a healthcheck-aware oneshot ordering helper or rely on the client library's reconnection policy (which this PR now documents).

- **Cassandra LWT response columns use bracket-form keys.** The `[applied]` column from `INSERT … IF NOT EXISTS` is the literal property name `'[applied]'` (with brackets) on the returned row object; `row.applied` is `undefined`. → Already mentioned in `knowledge/learnings.md` (2026-05-23 PR #229 pattern); reinforce in the shim wiki page once created.

### `docs/adr/0010-https-shim-write-path.md` — amendment proposal

Add a short "Hardening baseline" section at the end of the ADR pointing to the now-canonical systemd hardening directives in `data-pipeline/shim/deploy/jobflow-shim.service`. Any future `data-pipeline/` shim should start from the same baseline.

### What I noticed but did NOT propose

- The Phase 1 reviewer noted that most ADR-mirror wiki pages listed in `knowledge/wiki/index.md` are unresolved links (acknowledged in `learnings.md` 2026-05-23). Cleaning that up is a broader hygiene pass, not a finding from this PR.
- The shim has no automated tests. This is consistent with the project-wide "no tests for the Scout feature" decision in PRD #180; no proposal here.

PR #256, opened 2026-05-25
fix(ingester): skip missing-file (ENOENT) errors instead of aborting the backfill run

### LINKED ISSUE #255: fix(ingester): skip missing-file (ENOENT) errors instead of aborting the backfill run
## Background

The nightly backfill orchestrator (\`data-pipeline/orchestrators/backfill.js\`) sets the ingest date range from \`diskBounds.earliest\` → \`diskBounds.latestComplete\`. The orchestrator/fetcher decoupling is **intentional** (per ADR 0003 + the ingestion-pipeline wiki) and stays.

The Ingester then loops day-by-day in \`data-pipeline/ingester/ingester.js\`. Each day, the per-hour worker reads \`/mnt/hdd/gharchive/YYYY/MM/DD/H.json.gz\`.

When a single hour file is missing (or a whole day's directory is absent), the worker throws ENOENT (\`worker.js\` ~line 100), the main thread catches it inside \`processHourBatch\`, and the per-date \`try/catch\` at \`ingester.js:139–146\` runs:

\`\`\`js
} catch (err) {
  logger.error({ date, partition, sampleEventId, err: err.message },
               'failed to process date — no backfill_progress row written');
  exitCode = 1;
  break;       // ← abandons every remaining date in the range
}
\`\`\`

One missing-file error therefore aborts the *entire* remaining date range. The orchestrator then marks \`backfill_runs.status='failed'\`. Because the next nightly run starts fresh (status=failed → fresh path), it computes the same \`diskBounds.earliest\`, hits the same gap, fails the same way.

**Reproduced in production** on backfill run \`8997de70-57b3-11f1-af7e-70e375cc7c33\` (\`2026-05-24 21:01:55 UTC\`) on date \`2025-01-08\` — directory \`/mnt/hdd/gharchive/2025/01/08/\` is absent entirely. 7 days completed (\`2025-01-01 → 2025-01-07\`, 3,876 events) then everything from \`2025-01-08\` → \`2026-05-24\` was skipped. Logs in \`/home/tomer/jobflow-logs/backfill.log\`.

## Goal

A missing-file error (ENOENT) during backfill should be logged with full structured context and the loop should continue — to the next hour within the day, and to the next day at the outer loop. Non-ENOENT errors (corrupt gzip, JSON parse failure, Cassandra write failure, etc.) keep their current fatal-to-the-date behavior, but at the per-date level should \`continue\` (not \`break\`) so one bad date does not wedge the whole nightly run.

## Out of scope

- No orchestrator changes. \`diskBounds.earliest\` stays as-is.
- No schema changes to \`backfill_progress\`.
- No after-run report generation / notification (separate work item, not now).
- No change to hourly-mode behavior; hourly only operates on a tiny batch and exit-on-error is correct there.

### 1. \`data-pipeline/ingester/worker.js\`

Distinguish ENOENT from other worker errors. In the existing outer \`try/catch\` around the file pipeline (\`createReadStream → createGunzip → readline\`), when \`err.code === 'ENOENT'\` post a new message type \`fileMissing\` to the main thread; otherwise post \`workerError\` as today.

\`\`\`js
} catch (err) {
  try { rl?.close(); } catch {}
  try { fileStream?.destroy(); } catch {}
  if (err?.code === 'ENOENT') {
    parentPort.postMessage({
      type: 'fileMissing',
      hourId,
      filePath,
      message: err.message,
    });
  } else {
    parentPort.postMessage({
      type: 'workerError',
      hourId,
      message: err?.message ?? String(err),
    });
  }
}
\`\`\`

### 2. \`data-pipeline/ingester/ingester.js\` — main thread per-hour dispatch

In the worker message handler (around the existing \`workerError\` branch near \`ingester.js\` ~line 273 and surrounding dispatch logic — confirm exact location during impl), add a \`fileMissing\` branch:

- Log \`logger.warn({ hourId, filePath }, 'hour file missing — skipping')\`.
- Mark the hour as 'done' with zero events emitted (same accounting as a normal file completion).
- Do NOT set the dispatch-stopping flag; the batch continues.

### 3. \`data-pipeline/ingester/ingester.js\` — per-date loop (~lines 119–147)

Change \`break\` to \`continue\` in the per-date catch:

\`\`\`js
} catch (err) {
  logger.error(
    { date, partition: err.partitionKey ?? null, sampleEventId: err.sampleEventId ?? null, err: err.message },
    'date errored — skipping, continuing to next date'
  );
  exitCode = 1;   // run is still ultimately marked failed at the orchestrator level
  continue;       // do NOT abandon remaining dates
}
\`\`\`

Keep \`exitCode = 1\` so the orchestrator still marks the \`backfill_runs\` row failed if any date errored — visibility matters. The fix is about *not wedging* the rest of the run, not about hiding failures.

### 4. \`backfill_progress\` row policy for partial / empty days

| Case | Row written? | \`events_written\` | \`status\` |
|---|---|---|---|
| All 24 hours present, all processed | yes | N | \`completed\` (unchanged) |
| Some hours had \`fileMissing\` | yes | N (only what was processed) | \`completed\` (+ warn line listing skipped hours) |
| Every hour missing (e.g. \`2025-01-08\`) | yes | 0 | \`completed\` ← this is what stops the next nightly run from re-trying the gap forever |
| Non-ENOENT error fired | no | — | — (current behavior; \`continue\` moves on, retry possible next night) |

The status column on \`backfill_progress\` doesn't need new values; \`completed\` covers all 'we did everything we could here' cases.

## Test plan

- **Unit-style** (vitest or jest, matching whatever \`data-pipeline/ingester/__tests__/\` already uses): simulate ENOENT on one hour out of 24, assert the rest of the day completes, assert one warn line logged with the missing hourId, assert the \`backfill_progress\` row carries the correct lower \`events_written\`.
- **Unit-style**: simulate ENOENT on every hour of a date, assert a \`backfill_progress\` row with \`events_written=0\` is written, assert the next date in the range is still processed.
- **Live e2e on the deploy box** (\`tomer@192.168.10.12\`): with the existing on-disk gap at \`2025/01/08\`, kick a backfill run manually. Expected: \`backfill_progress\` gains a row for \`2025-01-08\` with \`events_written=0\`, then continues through \`2025-01-09 → latestComplete\`. Confirm via:
  \`\`\`sql
  SELECT date, events_written, status FROM jobflow.backfill_progress WHERE run_id=<new run>;
  \`\`\`
  Total event count in \`company_events\` should jump substantially (currently **846,401**).
- Confirm \`backfill_runs\` row for the new run is still marked \`failed\` (because we kept \`exitCode=1\`) but the per-date table reflects actual progress.

## Acceptance criteria

- [ ] Missing-file error in a worker no longer aborts the entire ingester run.
- [ ] Each missing hour is logged at WARN with structured fields (\`hourId\`, \`filePath\`).
- [ ] Each errored date is logged with structured fields (\`date\`, \`err.message\`).
- [ ] \`backfill_progress\` row is written for every date the loop attempts, including all-empty days, with accurate \`events_written\`.
- [ ] One PR opened against \`main\`. Branch suggestion: \`fix/ingester-skip-missing-files-and-continue\`.
- [ ] PR description references this issue.
- [ ] Live e2e on \`tomer@192.168.10.12\` confirms the previously-wedged date range past \`2025-01-08\` is now processed.
- [ ] Wiki page \`knowledge/wiki/ingestion-pipeline.md\` ('Worker architecture (post-#240)' + 'Failure semantics' sections) updated to reflect the new ENOENT semantic.

## Affected files (predicted)

- \`data-pipeline/ingester/worker.js\`
- \`data-pipeline/ingester/ingester.js\`
- \`data-pipeline/ingester/__tests__/\` (or wherever existing ingester tests live — add coverage)
- \`knowledge/wiki/ingestion-pipeline.md\` (one-line update to Failure semantics)
- No orchestrator files, no schema files.

## Related context

- ADR 0003 (\`docs/adr/0003-backfill-hourly-relay-race.md\`) — backfill/hourly hand-off via \`initialized\` flag.
- \`knowledge/wiki/ingestion-pipeline.md\` → 'Worker architecture (post-#240)' + 'Failure semantics' sections.
- \`docs/InjestionPlan.md\` §4.4 (Fetcher's analogous 404-tolerance pattern — informs the design here).
- Production incident: backfill run \`8997de70-57b3-11f1-af7e-70e375cc7c33\` failed at \`2026-05-24 21:01:55 UTC\` on \`2025-01-08\`; logs in \`/home/tomer/jobflow-logs/backfill.log\`.

tomerp20:
## Summary

- **worker.js**: distinguish ENOENT from other errors by posting a new `fileMissing` message type; other errors still post `workerError`.
- **ingester.js**: handle `fileMissing` in the main thread as a per-hour skip (log warn, mark hour 'done' with 0 events, keep batch going). In the per-date loop, change `break` → `continue` so a non-ENOENT error on one date no longer wedges later dates. Still sets `exitCode=1` for non-ENOENT errors so the orchestrator marks the run failed.
- **`backfill_progress` row policy**: rows are written for every date the loop attempts (including all-empty days with `events_written=0`), so the next nightly run does not retry permanent archive gaps.

## Acceptance criteria

- [x] Missing-file error in a worker no longer aborts the entire ingester run.
- [x] Each missing hour is logged at WARN with structured fields (`hourId`, `filePath`).
- [x] Each errored date is logged with structured fields (`date`, `err.message`).
- [x] `backfill_progress` row is written for every date the loop attempts, including all-empty days, with accurate `events_written`.
- [x] One PR opened against `main`. Branch suggestion: `fix/ingester-skip-missing-files-and-continue` ✓.
- [x] PR description references this issue.
- [ ] **Live e2e on `tomer@192.168.10.12` confirms the previously-wedged date range past `2025-01-08` is now processed.** *(Deferred — operator wants to run the manual backfill jointly with CPU/memory monitoring across fetcher + ingester + cassandra. Will be ticked after that session.)*
- [ ] Wiki page `knowledge/wiki/ingestion-pipeline.md` updated. *(Coming in a follow-up `docs:` commit on this same branch — per ship skill step 8.5.)*

## Notes

- No tests added. `data-pipeline/ingester` has no existing test framework and bootstrapping one would significantly exceed the scope of this surgical fix. The acceptance criteria do not require unit tests; verification is via the live e2e step (deferred per above).
- No orchestrator changes (`diskBounds.earliest` decoupling stays intact per ADR 0003 + ingestion-pipeline wiki).
- No schema changes to `backfill_progress`.
- Pure-ENOENT dates do not trip `exitCode=1`. A run with only ENOENT skips will complete `status=completed`; a run with any non-ENOENT date error stays `failed` for visibility.

## Related

Closes #255


tomerp20:
## Code Review — PR #256 (orchestrate-review)

Inline review (single general-purpose pass; infrastructure-only Node.js diff — neither `react-code-reviewer` nor `backend-code-reviewer` is the right fit per the skill's "Infrastructure-only PRs" branch). Security pass done inline. Knowledge-base sources consulted: [[ingestion-pipeline]], [[adr-0003-backfill-hourly-relay-race]], [[adr-0005-processed-files-hourly-only]], [[adr-0008-ingester-worker-side-decompression]].

**Verdict: APPROVE with minor follow-ups.** The fix is well-targeted, the message-passing accounting is correct, and the wiki + learnings updates are accurate. Findings below are MEDIUM observability / LOW polish — none block merge.

---

### Verified correct

- **`fileMissing` ↔ `fileDone` accounting equivalence.** The new branch in `attachWorker` mirrors the `fileDone` branch precisely: `statsByHour.set(hourId, {0,0,0})`, `dispatchedAll.add(hourId)`, `busyWorkers--`, worker pushed back to `readyWorkers`, `tryDispatch()` + `tryFinalize()` called. Because `inFlightByHour.get(h)` was initialised to 0 in `tryDispatch` and no events were ever emitted for this hour, the chronological finalize loop's `(inFlightByHour.get(h) ?? 0) > 0` guard passes and the hour is finalized in order with `totalEmitted=0`. Honors the contiguous-prefix invariant from [[adr-0005-processed-files-hourly-only]].
- **Pure-ENOENT day → `exitCode=0`** as PR claims. Trace: every hour → `fileMissing` → `tryFinalize` (never calls `recordWriteError`) → `processHourBatch` resolves with `0` → outer loop calls `flushAllPartitionBuffers()` (no-op) + `writeBackfillProgress(runId, date, 0)` → row written, no catch fires. `exitCode` stays 0. The `backfill_progress` row with `events_written=0` is exactly what stops the next nightly run from re-trying the gap. Matches the row-policy matrix in #255.
- **`recordWriteError` skipped on `fileMissing` is safe.** `fileMissing` is a known-tolerable terminal outcome for an hour — invoking the failure path would conflict with the new semantics. No invariant in the file relies on every dispatched hour passing through `recordWriteError` on failure.
- **`workerError` ↔ `fileMissing` interleave is race-free.** If a real `workerError` arrives for hour H1 after a `fileMissing` for hour H2: `writeError` is set, dispatch stops, `tryFinalize` short-circuits on the writeError guard, `checkTerminalCondition` waits for `busyWorkers==0 && totalInFlight==0`, then rejects `done`. The `fileMissing` handler still decrements `busyWorkers` and returns the worker to the pool, so quorum is reached cleanly. No deadlock.
- **`hourId.slice(0, 10)` extracts `YYYY-MM-DD` correctly** given the canonical hourId format `YYYY-MM-DD-H` from `disk-layout` (month/day zero-padded). Same slice pattern is already used in the per-date grouping at ingester.js:132 — the new code is consistent with the existing convention.
- **`break` → `continue` keeps `exitCode=1` sticky** across iterations. Visibility-to-orchestrator preserved per #255 acceptance criteria. The comment "ENOENT errors arrive as fileMissing messages (handled in attachWorker) and never reach this catch — they cannot trip exitCode on their own" is accurate.
- **Wiki + learnings consistency.** `knowledge/wiki/ingestion-pipeline.md` "Failure semantics" bullet and "Surprises / gotchas" entry accurately reflect the code. `updated:` date bump is correct. `knowledge/learnings.md` entry captures the architectural lesson cleanly. No conflicts with [[adr-0003-backfill-hourly-relay-race]] (orchestrator/fetcher decoupling stays — this fix is intentionally layered in the ingester/worker, exactly where [[adr-0003]] says per-row state belongs).

### Security pass (inline)

- No sudo usage. No secrets touched. No file permissions modified.
- **L1 — `filePath` in `fileMissing` log line.** `filePath` is computed as `hourIdToPath(GHARCHIVE_DIR, hourId)` → values like `/mnt/hdd/gharchive/2025/01/08/0.json.gz`. Deterministic, derived from a config env var, no PII, no secrets. Acceptable as-is; flagged only for completeness. If `GHARCHIVE_DIR` were ever repointed under a sensitive parent path, the warn line would surface it — but that risk lives in the env-var policy, not in this change.

---

### Findings

#### M1 — `missingHoursByDate` warn line is silently dropped if `flushAllPartitionBuffers` or `writeBackfillProgress` throws

ingester.js:144–151 — the new "date completed with missing hours" warn line sits AFTER `await writer.writeBackfillProgress(...)`. If the Cassandra-side flush or progress-write throws, control jumps to the catch block and the per-date warn (with the `missingHours` array) is never emitted. Operators investigating a date that had BOTH missing hours AND a Cassandra error lose the missing-hours context. The `missingHoursByDate` Map still holds the entries, but they aren't surfaced anywhere visible.

Suggestion: emit the missing-hours summary BEFORE `flushAllPartitionBuffers` (or in the catch as well), so the warn fires even on a Cassandra failure. Cheap diagnostic, no semantic risk.

#### M2 — Per-partition buffers from a failed date can leak into the next date's `events_written` count

ingester.js:154–166 (the `break` → `continue` change) — when a non-ENOENT error fires on date D, the catch logs and `continue`s to date D+1 WITHOUT calling `writer.flushAllPartitionBuffers()`. The next date's `processHourBatch` runs with leftover buffered events from D still sitting in cassandra-writer's per-partition buffers. The wiki's "Surprises / gotchas" already notes this design ("per-partition batch buffers in cassandra-writer.js now contain events from multiple hours concurrently. Correctness is preserved (Cassandra dedups via the full primary key…)") — so dataset correctness is fine (Cassandra full-key upserts dedup). But `events_written` for date D+1's `backfill_progress` row will include D's leftover events flushed at D+1's boundary, inflating its count.

Suggestion: in the catch (right before `continue`), call `await writer.flushAllPartitionBuffers()` in a defensive `try { … } catch (flushErr) { logger.error({ flushErr }, 'failed to flush after date error'); }` block, so D's buffered events land in Cassandra (idempotent upserts) and D+1's accounting starts clean. Alternative if the writer doesn't support discarding without flushing: document this in the wiki's "Surprises / gotchas" section.

This was the kind of thing [[adr-0005]] and [[adr-0008]] both warn about — the per-partition buffer is shared across hours by design, so any control-flow change that adds a new "skip a date" path needs to consciously handle the buffer.

#### L1 — `missingHoursByDate` is module-scope and never cleared

ingester.js:120–124 — the Map grows for the lifetime of the process. Fine for a single-shot CLI invocation walking a bounded date range (~365 entries for a year of backfill), but the comment "Module-scope because the worker-message handler … and the per-date loop both need access" is slightly misleading: they could equally well share a Map declared in the outer `try` block. Module-scope makes it harder to see the lifetime at a glance.

Suggestion: either move the declaration inside the `try { if (cmd.mode === 'backfill' && cmd.verb === 'range') { … } }` branch and pass it into `processHourBatch` (cleanest), OR add `missingHoursByDate.delete(date)` after the per-date summary log so it doesn't accumulate across dates. Functional impact is negligible at current scale; this is for future-readability.

#### L2 — `fileMissing` branch doesn't guard against `writeError`

ingester.js:365–380 — the `events` branch has `if (writeError) return; // drop further events once we've started failing`. The new `fileMissing` branch keeps logging warns and appending to `missingHoursByDate` even after a fatal `writeError` has fired. The state mutations are harmless (the entries are never read post-failure), but the warn lines add noise during error wind-down.

Suggestion: optionally add `if (writeError) { busyWorkers--; readyWorkers.push(worker); checkTerminalCondition(); return; }` at the top of the `fileMissing` branch — mirrors the `events` branch's defensive shape and keeps the wind-down log clean.

#### N1 — Comment polish: "module-scope because …" rationale

ingester.js:120–124 — the rationale "the handler appends, the loop reads to log a summary line before moving on" is true but doesn't justify *module-scope* specifically (closure-scope inside the outer `try` works identically). If you keep module-scope (per L1 alternative), prefer a one-liner like "module-scope for simplicity; cleared per date below" if you adopt the `delete(date)` pattern.

#### N2 — Wiki / learnings precision

`knowledge/wiki/ingestion-pipeline.md` "Surprises / gotchas" new bullet says "The orchestrator's `diskBounds.earliest` can point at a date whose hour files are not all present (e.g. a permanent gap from a 404 the Fetcher saw historically)". Strictly, the Fetcher's 404-tolerance pattern (`docs/InjestionPlan.md` §4.4) means it logs and continues — there isn't a marker on disk that says "this hour was 404'd." So the gap is invisible to the Ingester until it tries to open the file. The current phrasing is fine for operators but could be slightly more precise: "…a permanent gap from a 404 the Fetcher silently skipped historically (Fetcher logs the 404 but writes no on-disk marker, so the Ingester only discovers the gap when it tries to open the file)."

Optional; the existing phrasing is acceptable.

---

### Knowledge-base proposals from this review

Will be posted as a separate comment after this one (per the orchestrate-review skill's mandatory KB-integration step).

---

### Status

| Phase | Result |
|-------|--------|
| Phase 1 — Reviews + security pass | Done inline (1 single-agent review + security). **0 HIGH · 2 MEDIUM · 2 LOW · 2 NIT.** |
| Phase 2 — Auto-fixes | **Skipped** — operator explicitly requested code-only review (no live e2e on the deploy box). The 2 MEDIUMs are observability/accounting refinements, not correctness bugs blocking merge. Recommend fixing in a follow-up commit before the live verification session so the operator gets accurate `events_written` counts during the manual backfill. |
| Phase 3 — Final status | This comment. |

**Recommendation:** APPROVE. The fix correctly addresses the wedge described in #255. The two MEDIUM findings are worth a follow-up commit before the joint live verification session, but they don't block merge — both relate to observability/accounting around the *failure* path, not the (verified) ENOENT-tolerance success path.

tomerp20:

Per the orchestrate-review skill's mandatory KB-integration step. Proposing edits only — not editing the knowledge base directly. Operator to triage and apply (or reject) in a follow-up `docs:` commit.

### (a) `knowledge/wiki/ingestion-pipeline.md` page edits

**1. "Surprises / gotchas" — add a bullet about the per-date `continue` and shared buffer interaction**

The page already notes that per-partition buffers in `cassandra-writer.js` contain events from multiple hours concurrently and that Cassandra dedup preserves dataset correctness. But after PR #256 changed the per-date catch from `break` to `continue`, there's now a new control-flow path where date D's buffered events can leak into date D+1's `events_written` accounting (the buffer isn't flushed in the catch path before continuing). Suggested addition under "Surprises / gotchas":

> - When a non-ENOENT error fires inside a per-date `processHourBatch` (backfill mode), the per-date catch `continue`s without flushing `cassandra-writer`'s per-partition buffers. Leftover events from the failed date are still buffered and will be flushed at the *next* date's boundary — Cassandra dedup keeps `company_events` correct, but `backfill_progress.events_written` for the next date will include the leftover events from the failed one. If accurate per-date counts matter, add a defensive `flushAllPartitionBuffers()` in the catch path. See PR #256 review M2.

**2. "Worker architecture" → "Failure semantics" — document the `writeError` guard convention**

The dispatch-loop message handlers (`events`, `fileDone`, `workerError`, and now `fileMissing`) have an implicit convention: any handler that mutates state or schedules new work should check `writeError` first to avoid noise during error wind-down. The `events` branch does this explicitly (`if (writeError) return;`); `fileMissing` from PR #256 does not. Suggested clarification at the end of the "Failure semantics" paragraph:

> Convention for adding new worker-message branches in `attachWorker`: check `writeError` at the top before mutating dispatch state or logging — this keeps the error wind-down quiet. The `events` branch is the reference.

### (b) `knowledge/learnings.md` bullets to add

**Bullet 1 — Control-flow changes touching shared per-partition buffers need an explicit flush decision**

> - **2026-05-25 — `break` → `continue` in a date loop sharing a flush boundary (PR #256 review).** When the cassandra-writer's per-partition buffers are explicitly multi-hour by design (see [[ingestion-pipeline]] gotchas), any new control-flow path that skips the normal "end of date" flush boundary needs to either (a) flush in the new path or (b) document and accept that downstream accounting (`events_written` per `backfill_progress` row) will conflate dates. The PR #256 fix added a per-date `continue` without picking either option, so non-ENOENT errors during backfill now silently inflate the next date's count by the failed date's buffered events (Cassandra dedup keeps the dataset correct). → Whenever editing the per-date loop in `ingester.js`, audit every `break`/`continue`/early-`return` against `flushAllPartitionBuffers()` semantics; the buffer is not date-scoped.

**Bullet 2 — Single-shot CLI module-scope state is convenient but obscures lifetime**

> - **2026-05-25 — module-scope state in a one-shot CLI (PR #256 review).** `missingHoursByDate` was declared at module scope so both the worker-message handler (inside `processHourBatch`) and the per-date loop could access it. That works because the ingester is single-shot, but it hides the lifetime: the Map grows for the whole process and there's no `delete(date)` after the per-date summary log. Future control-flow changes (e.g. running multiple date ranges in one invocation, retrying a date in-process) will silently re-use stale entries. → Prefer closure-scoped state in a single outer `try { … }` and pass it into `processHourBatch` as an argument; resort to module-scope only when at least two callers across module boundaries need access.

**Bullet 3 — `writeError` guard convention for worker-message handlers**

> - **2026-05-25 — implicit `writeError` guard convention (PR #256 review).** The `events` branch in `attachWorker` checks `if (writeError) return;` before mutating state. New branches added in PR #256 (`fileMissing`) skipped that guard, so they log + mutate state during error wind-down (harmless but noisy). → When adding a new worker-message branch, mirror the `events` branch's defensive shape: check `writeError` first, decrement `busyWorkers`, return the worker, call `checkTerminalCondition`, and return. Document this as a convention in `knowledge/wiki/ingestion-pipeline.md` "Failure semantics".

---

*Proposed by the orchestrate-review pass for PR #256. No KB edits applied — these are recommendations only.*

Issue #259, opened 2026-06-04
WhatsApp Notifier — outbound WhatsApp Message on every Application creation


tomerp20:
## Problem Statement

As the JobFlow user, every time a new [[application|Application]] enters my pipeline — whether I add it manually or the [[email-agent|Email Agent]] auto-creates it from an [[application-receipt|Application Receipt]] — I only find out by looking at JobFlow. I want a specific person (Netali) to be told in real time, over WhatsApp, the moment a new Application is created, without me opening the app or taking any action.

## Solution

A new automated process — the **[[whatsapp-notifier|WhatsApp Notifier]]** — sends a single ephemeral outbound **WhatsApp Message** from my personal WhatsApp account to one hardcoded recipient (Netali, `+972 54-448-3175`) whenever an [[application|Application]] is created. It is a sibling of the [[email-agent|Email Agent]] and [[company-scout|Company Scout]], triggered by Application creation rather than by email or [[first-sighting|Company novelty]].

It reuses existing infrastructure end to end: a new `POST /notify` route is added to the existing Cassandra write shim on the [[infra-linux-deployment|Linux box]], reached from JobFlow (Render) over the **existing ngrok static domain** — no new container, no proxy, no second tunnel. JobFlow fires the call **detached** so Application creation latency is never affected, and the WhatsApp Message is never persisted in JobFlow (it is explicitly **not** a [[notification|Notification]]).

The message reads:

```
Hey Netali, I found new position, Here are the Details:
<Company> — <Role>
<job-posting URL>
```

The full design is authoritative in [[adr-0011-whatsapp-notifier-on-shim]], [[whatsapp-notifier]], `CONTEXT.md`, and the grill transcript at `knowledge/raw/2026-06-04-grill-whatsapp-notifier.md`.

## User Stories

1. As the JobFlow user, I want a WhatsApp Message sent automatically whenever an Application is created, so that I am notified in real time without opening JobFlow.
2. As the JobFlow user, I want the message sent on **both** the manual create path and the Email Agent auto-create path, so that no new Application is ever missed regardless of how it entered.
3. As the JobFlow user, I want the message delivered to one specific person (Netali), so that the right person is kept informed of every new opportunity.
4. As the JobFlow user, I want the message to contain the company name, the role title, and the job-posting URL, so that Netali has the essential details at a glance.
5. As the JobFlow user, I want the message to still be sent when an Application has no job-posting URL (just dropping the link line), so that I am notified even for Applications lacking a link.
[...]

PR #260, opened 2026-06-04
WhatsApp Notifier — outbound WhatsApp Message on every Application creation (#259)

tomerp20:
## Summary
- Sends a WhatsApp Message from the user's personal account to one hardcoded recipient whenever an **Application** is created (manual + Email Agent paths).
- Implemented as a new `POST /notify` route on the existing Cassandra write shim — **no new component** — reached from JobFlow over the existing ngrok tunnel.
- Implements PRD #259. Design locked in ADR 0011.

## Changes
**Shim (`data-pipeline/shim/`)**
- `lib/message.js` — pure message formatter (`"Hey Netali…"`) + recipient normalizer (`+972525912293 → 972525912293@c.us`); recipient hardcoded shim-side.
- `lib/whatsapp.js` — whatsapp-web.js client wrapper: `LocalAuth` session, QR via `qrcode-terminal`, ready/disconnected state, **throttled serialized send queue** (min 3s gap) so bulk syncs aren't spam-shaped.
- `shim.js` — `POST /notify` (reuses bearer auth; `400/401/415`; `503` when WhatsApp not ready); client initialized **after** `listen()` and isolated so a dead Chromium never affects `/companies`.
- `Dockerfile` — `node:20-alpine → node:20-slim` + system Chromium (puppeteer's bundled Chromium can't run on musl).
- `docker-compose.yml` — bind-mount the `LocalAuth` session dir + `shm_size` for Chromium.
- `RUNBOOK.md` — one-time QR scan, Render config, `/notify` smoke test.

**JobFlow (`backend/`)**
- `whatsappNotifier/whatsappNotifier.ts` — `WhatsAppNotifier` interface + `HttpShimWhatsAppNotifier` + `LoggingWhatsAppNotifier` dev fallback (no-op when `WHATSAPP_NOTIFY_URL` unset), mirroring `CompanyRegistry`.
- `cardService.createCard` — `notifyApplicationCreated` fired **detached** next to `runCompanyCheck` (transaction-boundary edge accepted, ADR 0011).
- `config/env.ts` — `WHATSAPP_NOTIFY_URL` (optional; https + reused token required in prod).

## Test plan
No automated tests (per request). Verification is the **ADR-0006 smoke transcript** — to be captured on the box after deploy:
- [ ] `/companies` still green (no regression from the base-image change)
- [ ] `/notify` happy path **with** URL → recipient receives `Hey Netali…` + link
- [ ] `/notify` happy path **without** URL → message sent, link line dropped
- [ ] Negative paths: `401` wrong token, `400` bad body, `415` wrong content-type, `503` when client not ready
- [ ] First-run QR scanned once; session survives a container restart

## Deploy (operator)
1. Box: `mkdir -p /home/tomer/whatsapp-session && sudo chown 1000:1000 …`, `docker compose up -d --build shim`, watch `docker logs jf-shim -f`, scan the QR once.
2. Render: set `WHATSAPP_NOTIFY_URL` to the same ngrok base as `COMPANY_REGISTRY_URL`.

> ⚠️ whatsapp-web.js is unofficial (against WhatsApp ToS); sender-number ban risk accepted.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -84,6 +84,14 @@ _Avoid_: Alert, Toast, Message
 A user-created action item with a description, priority (low / medium / high / urgent), and status (active / completed). A Task may optionally be linked to an Application; without a link it is standalone. Linked Tasks appear on the Application's detail panel; all Tasks are accessible from the dedicated Tasks page.
 _Avoid_: Todo (internal/DB term), Checklist item
 
+**WhatsApp Notifier**:
+An automated process that sends an outbound **WhatsApp Message** from the user's personal WhatsApp account to a single configured recipient whenever an Application is created — across both the manual create path and the Email Agent's auto-create path. A sibling of the Email Agent and Company Scout (automated processes acting on Applications), but triggered by Application creation rather than by email or Company novelty. It runs as a route on the Cassandra write shim on the Linux box and is reached from JobFlow over the existing ngrok tunnel. It produces **no Notification** and writes no `notifications` row — the WhatsApp Message is ephemeral and never persisted in JobFlow.
+_Avoid_: Notification (reserved for the persistent in-app entity), Alert, WhatsApp bot
+
+**WhatsApp Message**:
+The single ephemeral outbound message the WhatsApp Notifier sends on Application creation. Not stored anywhere in JobFlow; not a Notification.
+_Avoid_: Notification, Alert, Toast
+
 ## Relationships
 
 - A **User** owns a set of **Stages**
@@ -98,6 +106,7 @@ _Avoid_: Todo (internal/DB term), Checklist item
 - The **First Sighting** of a **Company** on a new **Application** triggers the **Company Scout**
 - The **Company Scout** resolves a **Company** to zero or more candidate **Orgs**, scores each candidate, and admits those above the acceptance threshold
 - The **Company Scout** classifies each admitted **Org** as having an **Active GitHub Presence** or not, and writes only the active ones to the JobFlow Analytics system's `companies` table — one row per (Company, Org) pair
+- The **WhatsApp Notifier** sends one outbound **WhatsApp Message** to a single configured recipient whenever an **Application** is created (both the manual and Email Agent auto-create paths); it produces no **Notification** and persists nothing in JobFlow
 
 ## Example dialogue
 
--- a/backend/src/config/env.ts
+++ b/backend/src/config/env.ts
@@ -25,6 +25,11 @@ const schema = z.object({
   GITHUB_TOKEN:            z.string().optional(),
   COMPANY_REGISTRY_URL:    z.string().url().optional(),
   COMPANY_REGISTRY_TOKEN:  z.string().optional(),
+
+  // WhatsApp Notifier (ADR 0011). Optional — feature is a no-op when unset.
+  // Base URL of the shim (same host as COMPANY_REGISTRY_URL); the notifier
+  // appends /notify and reuses COMPANY_REGISTRY_TOKEN as the bearer token.
+  WHATSAPP_NOTIFY_URL:     z.string().url().optional(),
 }).superRefine((v, ctx) => {
   if (v.NODE_ENV !== 'production') return;
 
@@ -42,6 +47,15 @@ const schema = z.object({
 
   if (v.COMPANY_REGISTRY_URL && !v.COMPANY_REGISTRY_URL.startsWith('https://')) {
     ctx.addIssue({ code: 'custom', path: ['COMPANY_REGISTRY_URL'], message: 'COMPANY_REGISTRY_URL must use https:// in production (bearer token must not be sent in plaintext)' });
+  }
+
+  // WhatsApp Notifier reuses COMPANY_REGISTRY_TOKEN as its bearer token, so it
+  // must be present when the notify URL is set; and it must be https in prod.
+  if (v.WHATSAPP_NOTIFY_URL && !v.COMPANY_REGISTRY_TOKEN) {
+    ctx.addIssue({ code: 'custom', path: ['COMPANY_REGISTRY_TOKEN'], message: 'COMPANY_REGISTRY_TOKEN is required when WHATSAPP_NOTIFY_URL is set (reused as the shim bearer token)' });
+  }
+  if (v.WHATSAPP_NOTIFY_URL && !v.WHATSAPP_NOTIFY_URL.startsWith('https://')) {
+    ctx.addIssue({ code: 'custom', path: ['WHATSAPP_NOTIFY_URL'], message: 'WHATSAPP_NOTIFY_URL must use https:// in production (bearer token must not be sent in plaintext)' });
   }
 
   for (const [k, defaultVal] of [
--- a/backend/src/services/cardService.ts
+++ b/backend/src/services/cardService.ts
@@ -3,6 +3,7 @@ import db from '../config/database';
 import { AppError } from '../middleware/errorHandler';
 import { shiftUp, shiftDown, withTransaction } from '../util/positions';
 import { runCompanyCheck } from './companyScout/companyScout';
+import { notifyApplicationCreated } from './whatsappNotifier/whatsappNotifier';
 
 export interface CardFilters {
   stage?: string;
@@ -328,6 +329,17 @@ export const cardService = {
         careersUrl: data.careers_url,
       }).catch(() => { /* intentionally suppressed — Scout errors must not fail card creation */ });
     }
+
+    // WhatsApp Notifier (ADR 0011) — fire detached on EVERY Application
+    // creation (manual + Email Agent paths), never awaited so card-creation
+    // latency is unaffected. NOTE: createCard may run inside gmailSync's
+    // per-email transaction, so on a rare rollback this can notify for an
+    // Application that was never saved (accepted edge — see ADR 0011).
+    notifyApplicationCreated({
+      company: data.company_name,
+      role: data.role_title,
+      url: data.application_url,
+    });
 
     // Return card with stage name
     const stage = await runner('stages').where({ id: card.stage_id }).first();
--- a/backend/src/services/whatsappNotifier/whatsappNotifier.ts
+++ b/backend/src/services/whatsappNotifier/whatsappNotifier.ts
@@ -1,0 +1,130 @@
+import logger from '../../config/logger';
+import { env } from '../../config/env';
+
+/**
+ * WhatsApp Notifier (ADR 0011). Sends an outbound WhatsApp Message on every
+ * Application creation by POSTing { company, role, url } to the shim's /notify
+ * route. The recipient and wording live shim-side — JobFlow only supplies the
+ * three data fields.
+ *
+ * Mirrors the CompanyRegistry pattern: an interface with a production HTTP
+ * implementation and a logging dev fallback, fire-and-forget so card-creation
+ * latency is never affected.
+ */
+export interface ApplicationDetails {
+  company: string;
+  role: string;
+  url?: string | null;
+}
+
+export interface WhatsAppNotifier {
+  notify(details: ApplicationDetails): Promise<void>;
+}
+
+/**
+ * Dev fallback used when WHATSAPP_NOTIFY_URL is not configured. Logs a
+ * structured line describing what would have been sent — so local dev and
+ * tests never message anyone.
+ */
+export class LoggingWhatsAppNotifier implements WhatsAppNotifier {
+  notify(details: ApplicationDetails): Promise<void> {
+    logger.info('whatsapp_notifier.would_notify', {
+      service: 'whatsapp-notifier',
+      company: details.company,
+      role: details.role,
+      url: details.url ?? null,
+    });
+    return Promise.resolve();
+  }
+}
+
+/**
+ * Production implementation — POSTs to the shim's /notify route running
+ * alongside Cassandra on the analytics-pipeline host. See ADR 0011.
+ *
+ * Fire-and-forget: network errors, 4xx, and 5xx are caught and logged at warn;
+ * they never propagate so card creation latency is unaffected.
+ */
+export class HttpShimWhatsAppNotifier implements WhatsAppNotifier {
+  private readonly url: string;
+  private readonly token: string;
+
+  constructor(url: string, token: string) {
+    this.url = url;
+    this.token = token;
+  }
+
+  async notify(details: ApplicationDetails): Promise<void> {
+    const controller = new AbortController();
+    // Mirror the 3 s timeout used by the Company Scout registry.
+    const timer = setTimeout(() => controller.abort(), 3000);
+
+    try {
+      const res = await fetch(`${this.url}/notify`, {
+        method: 'POST',
+        headers: {
+          'Authorization': `Bearer ${this.token}`,
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify({
+          company: details.company,
+          role: details.role,
+          url: details.url ?? undefined,
+        }),
+        signal: controller.signal,
+      });
+
+      clearTimeout(timer);
+
+      if (!res.ok) {
+        logger.warn('whatsapp_notifier.http_error', {
+          service: 'whatsapp-notifier',
+          company: details.company,
+          status: res.status,
+        });
+        return;
+      }
+
+      logger.info('whatsapp_notifier.queued', {
+        service: 'whatsapp-notifier',
+        company: details.company,
+        role: details.role,
+      });
+    } catch (err: unknown) {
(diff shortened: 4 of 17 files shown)


tomerp20:
## Infrastructure Review (general-purpose)

Scope: container/deploy surface for the WhatsApp Notifier on the shim — `Dockerfile`, `docker-compose.yml`, `RUNBOOK.md`, `package.json` build/deploy implications. Reviewed against [[adr-0011-whatsapp-notifier-on-shim]] (`docs/adr/0011`), [[adr-0010-https-shim-write-path]] (`docs/adr/0010`), and ADR 0006 (smoke-test transcripts).

Overall this is solid and well-reasoned. The headless-Chromium path is mostly correct: `node:20-slim` is Debian **bookworm** so `libasound2` is the right package name (the `libasound2t64` rename is trixie/Debian 13 only); `PUPPETEER_SKIP_DOWNLOAD=true` is set *before* `npm install` so it takes effect; `/usr/bin/chromium` is the correct Debian path; the `node` uid-1000 / host `chown 1000:1000` pairing is consistent; and `whatsapp.init().catch(...)` in `shim.js:258` means a Chromium launch failure is caught and logged with the server still listening — so `restart: unless-stopped` will **not** crash-loop. The `/companies` Cassandra path is unaffected (cassandra-driver + pino are pure JS on Debian; ADR 0011 isolation guard holds). Findings below are mostly hardening, not blockers.

---

**[MEDIUM]** `data-pipeline/shim/package.json:16-17` (+ no `package-lock.json`, Dockerfile:51 uses `npm install`)
`whatsapp-web.js ^1.34.0` is unpinned and there is no lockfile, so each rebuild can resolve a different puppeteer driver version. The image launches the *system* apt Chromium (pinned at build time), but the puppeteer DevTools-protocol version floats independently — a future `whatsapp-web.js`/puppeteer bump can drift out of sync with the apt Chromium and silently break launch (→ perpetual 503 on `/notify`). Since this rebuilds a working production container (ADR 0011 "base-image change touches a working production container"), reproducibility matters.
*Fix:* commit a `package-lock.json` and switch the Dockerfile to `npm ci --omit=dev`.

**[MEDIUM]** `data-pipeline/shim/RUNBOOK.md:249-252` — root-owned host dir failure mode is undocumented
The chown step is correct, but the symptom of getting it wrong is not called out. If `/home/tomer/whatsapp-session` is created/owned by root (e.g. the operator runs `mkdir` as root, or skips the `chown`), the uid-1000 `node` user cannot write the `LocalAuth` session, so `whatsapp.authenticated`/`whatsapp.ready` never persist and every restart re-shows the QR (perpetual QR loop) — easy to misread as a code bug.
*Fix:* add a one-line note: "If you see a fresh QR on every restart, the host session dir is not writable by uid 1000 — re-run the `chown 1000:1000`. Verify with `ls -ld /home/tomer/whatsapp-session` (owner must be `1000`)."

**[LOW]** `data-pipeline/shim/RUNBOOK.md:276-303` — `/notify` smoke section omits the `/companies` no-regression step
ADR 0006 / ADR 0011 line 50 require the PR transcript to show `/companies` *still green* **and** a `/notify` send. The `/companies` smoke test exists earlier in the doc (lines 139-211), but an operator following the new WhatsApp section in isolation could skip it after a rebuild that changed the base image.
*Fix:* add a sentence cross-linking the `/companies` smoke test as a mandatory post-rebuild regression check, or duplicate the single happy-path `/companies` curl into the WhatsApp section.

**[LOW]** `data-pipeline/shim/Dockerfile:22-47` — `--no-install-recommends` + explicit lib list is redundant and slightly fragile
The apt `chromium` package's hard `Depends` already pull libnss3/libgbm1/libasound2/gtk/etc., so the explicit list is belt-and-suspenders, not the thing that makes launch work. That is fine, but the list omits a couple Chromium sometimes wants (`libxrender1`, `libxtst6`) — only harmless *because* `chromium`'s Depends cover them. The risk is the inverse: the hand-maintained list can drift and give a false sense that it is authoritative.
*Fix:* either trust the `chromium` Depends and trim the list to the known headless extras (`fonts-liberation`, `ca-certificates`), or add a build-time comment that the list is supplementary to chromium's own Depends. No action strictly required.

**[NIT]** `data-pipeline/docker-compose.yml:80-82` — `shm_size: "512mb"` is largely moot alongside `--disable-dev-shm-usage`
`lib/whatsapp.js:40` passes `--disable-dev-shm-usage`, which makes Chromium back shared memory with `/tmp` instead of `/dev/shm`, so the raised `shm_size` is mostly inert. It is harmless belt-and-suspenders and a reasonable safety margin; flagging only so a future reader doesn't think one depends on the other. No change needed.

**[NIT]** `data-pipeline/shim/Dockerfile:17` — `PUPPETEER_SKIP_DOWNLOAD` name
Correct for puppeteer ≥ v19 (which `whatsapp-web.js` 1.34 uses), and ordering is right. If the puppeteer dep ever resolves older than v19 the legacy name was `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD`. With a pinned lockfile (see MEDIUM above) this is a non-issue.

---

**Verdict:** Approve with minor changes. No launch-blocking or blast-radius defects — the `/companies` path is protected and the init-failure-can't-thrash design is correct. Address the two MEDIUMs (lockfile + npm ci; document the root-owned-dir QR-loop symptom) before merge; LOW/NIT are optional hardening.

**Severity counts:** HIGH 0 · MEDIUM 2 · LOW 2 · NIT 2

tomerp20:
## Backend Code Review (backend-code-reviewer)

**Classification:** STANDARD (fire-and-forget personal-notification feature; not a critical write path). Reviewed only the backend + Node service logic per scope — `whatsappNotifier.ts`, the `cardService` wiring, `env.ts`, and the shim's `/notify` route + `lib/whatsapp.js` / `lib/message.js`. Docs/ADR/compose/Dockerfile excluded.

**ESLint/tsc baseline:** Ran `eslint` + `tsc --noEmit` against the three changed backend files (symlinked node_modules, removed after). `whatsappNotifier.ts` and `env.ts` are **clean** (0 errors). `cardService.ts` reports 58 `@typescript-eslint/no-unsafe-*` errors but **all are at lines 398–521 — pre-existing Knex-`any` noise in untouched code, none in the changed region (323–342)**. New code adds no lint debt.

Overall this is a faithful mirror of the `CompanyRegistry` / ADR 0010 pattern and the ADR 0011 isolation contract holds (a dead Chromium 503s `/notify` and never touches `/companies`). Findings are minor.

### Findings

**[MEDIUM] `data-pipeline/shim/lib/message.js:25` — newline/control-char injection into the outbound WhatsApp Message.**
`company` and `role` are user-controlled card fields, interpolated raw into `` `${company} — ${role}` ``. A value containing a newline injects extra lines into the rendered message — e.g. a `role` of `"Eng\nhttps://evil.example"` forges a link line below the real content, or fakes the "Hey Netali…" framing. Outbound formatting-injection (not classic XSS), but it lets card content spoof message structure.
*Fix:* sanitize before formatting — strip control chars / collapse newlines:
```js
const clean = (s) => String(s).replace(/[\r\n\t]+/g, ' ').trim();
// ...
`${clean(company)} — ${clean(role)}`
```
Apply the same `clean()` to `url` (a stray newline in `url` also breaks line structure).

**[LOW] `data-pipeline/shim/lib/whatsapp.js:64-76` — `drain()` does not re-check `ready`; a mid-drain `disconnected`/`auth_failure` silently drops every queued message.**
`/notify` gates on `isReady()` at enqueue time, but once items are queued `drain()` loops calling `client.sendMessage` regardless of state. If the session drops mid-drain, each send throws, is logged `whatsapp.send_failed`, and the message is lost after a 200/"queued" was already returned to JobFlow. This is *consistent with* the ADR 0011 "Silent death" accepted edge ([[whatsapp-notifier]] → Surprises), so not a blocker — but the drain loop is slightly broader than the documented "dead session at request time" case.
*Fix (v1-optional):* re-check `ready` at the top of each loop iteration (re-queue or drop-with-loud-log when it flips false) so behaviour matches the 503 gate; or at minimum add a code comment tying this to the ADR's accepted silent-drop edge.

**[LOW] `backend/src/services/whatsappNotifier/whatsappNotifier.ts:112-120` — `resolveWhatsAppNotifier()` memoizes via a module-level `cached`, diverging from `resolveRegistry()` which constructs fresh each call.**
Harmless in production (env read once at module load), but inconsistent with the sibling pattern this file explicitly mirrors, and a memoized singleton is a test-isolation footgun (a test that stubs `env.WHATSAPP_NOTIFY_URL` after first resolve gets the stale instance). No current tests reference it.
*Fix:* drop the memoization to match `resolveRegistry()`, or keep it and add an exported reset for tests. Prefer matching `resolveRegistry()`.

**[NIT] `data-pipeline/shim/shim.js:179` / `lib/message.js` — `url` accepted as any string, never URL-validated, sent verbatim.**
`body.url` only has its `typeof` checked; a garbage / `javascript:` string is rendered into the message as-is. Low impact (JobFlow sends a trusted `application_url`; WhatsApp does its own link preview), but a cheap guard adds defense-in-depth.
*Fix:* optionally `new URL(body.url)` in try/catch and drop the line if it doesn't parse.

**[NIT] `data-pipeline/shim/lib/whatsapp.js:18` — queue unbounded in principle.**
In practice the `isReady()` 503 gate short-circuits `enqueue` while WhatsApp is down, so the queue only grows as fast as `/notify` is hit *while ready* — bounded by the 3s throttle and real traffic (a few/day). Not a real leak; a one-line defensive cap (drop + warn above e.g. 1000) would document the intent.

### Testing gaps
- No unit test for `message.js#formatMessage` (URL-present vs URL-absent, plus the sanitization if the MEDIUM is fixed). Pure function, trivially testable, the one piece with branching logic.
- No test asserting `notifyApplicationCreated` never rejects / never throws into `createCard`. The fire-and-forget contract is the whole safety story — one test pinning it is cheap insurance.
- Per ADR 0006, this data-pipeline PR needs the smoke-test transcript (`/companies` still green + a `/notify` send) attached before merge — flagging, not blocking.

### Verdict
**CHANGES REQUESTED** — 1 MEDIUM, 2 LOW, 2 NIT. Only the MEDIUM (message-injection sanitization) is worth blocking on; the rest is optional polish or v1-accepted per ADR 0011.

tomerp20:
## Security Review (orchestrator, inline)

OWASP / secrets / auth / injection / CORS pass on the WhatsApp Notifier diff.

**Clean — no action:**
- **Auth:** `/notify` reuses the existing constant-time `checkAuth` (timingSafeEqual) bearer check — same protection as `/companies`, no new bypass surface.
- **Token in transit:** `WHATSAPP_NOTIFY_URL` is `https://`-enforced in production via `env.ts` superRefine (bearer reused from `COMPANY_REGISTRY_TOKEN` is never sent in plaintext) — matches the PR #249 learning.
- **DoS:** `/notify` reuses `readBody`'s `MAX_BODY_BYTES` (64 KB) cap; the `isReady()`→503 gate bounds the send queue in practice.
- **No SSRF** (fetch target is operator-configured env, not user input), **no CORS surface** (server-to-server), **no SQLi** (no new queries; Knex stays parameterized), no stack-trace leakage in error responses.

**Findings:**
- **[MEDIUM] Message-content injection — `data-pipeline/shim/lib/message.js:25`.** `company`/`role` are user-controlled card fields interpolated raw; an embedded `\n` forges extra lines in the outbound WhatsApp Message (e.g. a spoofed link line). Same issue the backend reviewer raised. Fix in Phase 2: strip CR/LF/control chars from `company`, `role`, `url`.
- **[LOW] Supply-chain integrity — `data-pipeline/shim/package.json` + no `package-lock.json`.** `whatsapp-web.js`/puppeteer are unpinned with no lockfile, so a rebuild can resolve a different (potentially compromised) transitive tree. Fix in Phase 2: commit a lockfile + `npm ci`.
- **[LOW] PII in VCS — `data-pipeline/shim/lib/message.js:11`.** Netali's real phone number (`+972525912293`) is committed to source and now lives permanently in git history (sensitive if this repo is/ever becomes public). This was an explicit product decision ("hardcode it in the shim"), so **left as-is — flagged for human awareness only.** If undesired, move it to a shim-side env var (out of VCS) without changing the design.

**Verdict:** APPROVE with the MEDIUM addressed. Counts: HIGH 0 · MEDIUM 1 · LOW 2.

tomerp20:
## ✅ Code Review Complete — Phase 3 Status

**PR #260 — WhatsApp Notifier.** Background session, so the orchestrator ran inline with two `general-purpose` (opus) reviewers pointed at the agent specs + an inline security pass (the documented bg-session pattern).

| Reviewer | HIGH | MEDIUM | LOW | NIT | Verdict |
|---|---|---|---|---|---|
| Backend (`backend-code-reviewer` spec) | 0 | 1 | 2 | 2 | CHANGES REQUESTED |
| Infrastructure (`general-purpose`) | 0 | 2 | 2 | 2 | Approve w/ minor |
| Security (orchestrator, inline) | 0 | 1 | 2 | 0 | Approve w/ MEDIUM |

Deduped across reviewers and resolved in fix commit `3b7c87e`:

| # | Sev | Issue | Disposition |
|---|---|---|---|
| 1 | MEDIUM | Newline injection — `company`/`role` could forge message lines (`message.js`) | **Fixed** — collapse whitespace + http(s)-validate `url` |
| 2 | MEDIUM | No lockfile → unpinned `whatsapp-web.js`/puppeteer rebuilds | **Fixed** — committed `package-lock.json` (pins 1.34.7) + `npm ci` |
| 3 | MEDIUM | Root-owned session dir failure mode undocumented (perpetual QR loop) | **Fixed** — RUNBOOK note + `chown`/verify steps |
| 4 | LOW | `drain()` didn't re-check `ready` → silent mid-drain drops | **Fixed** — re-check + logs dropped count |
| 5 | LOW | `resolveWhatsAppNotifier()` module-level memoization (test leak) | **Fixed** — construct fresh, matches `resolveRegistry()` |
| 6 | LOW | `/notify` smoke section missing `/companies` no-regression (ADR 0006) | **Fixed** — cross-linked |
| 7 | LOW | Dockerfile lib list looks authoritative but is redundant | **Fixed** — clarifying comment |
| 8 | NIT | `url` never validated | **Fixed** — folded into `sanitizeUrl` |
| 9 | LOW | Netali's phone number in VCS (PII) | **Deferred** — explicit product decision; flagged for human |
| 10–11 | NIT | `shm_size` inert w/ `--disable-dev-shm-usage`; `PUPPETEER_SKIP_DOWNLOAD` name | **No change** — moot |

**Verification:** backend `tsc --noEmit` exit 0; all shim JS pass `node --check`; backend ESLint baseline unchanged (the 58 `cardService.ts` errors are pre-existing Knex-`any`, none in the changed region).

**Needs human action before merge:**
- ⚠️ Capture the **ADR-0006 smoke transcript** on the box (`/companies` no-regression + `/notify` happy/negative paths) — cannot be produced without the device QR scan.
- Decide whether the hardcoded recipient number (#9) is acceptable in VCS.

Per workflow: **not merged** — awaiting your review and approval.

tomerp20:

Patterns surfaced here that the wiki/learnings don't yet capture. **Proposals only — not applied.**

**(a) `knowledge/wiki/whatsapp-notifier.md` — add to "Surprises / gotchas":**
- *Root-owned session dir → perpetual QR loop.* If the host `LocalAuth` bind-mount isn't owned by the container's uid 1000, the session can't persist and re-authenticates every restart — easily misread as a code bug. `chown 1000:1000` the host dir.
- *Outbound-message field sanitization.* `company`/`role` are user-controlled card fields; they're whitespace-collapsed before formatting so an embedded newline can't forge extra message lines. The `url` is http(s)-validated and dropped otherwise.
- *Reproducible Chromium image.* The shim image now bundles Chromium; puppeteer's protocol version floats independently of the apt-pinned Chromium, so the build uses a committed lockfile + `npm ci` to keep rebuilds in sync.

**(b) `knowledge/learnings.md` — add bullets:**
- When a shim route interpolates user-controlled fields into an outbound message (WhatsApp/SMS/email), collapse whitespace (`\s+`→space) to neutralize line-forging injection — the same class as header injection, minus the transport.
- For any container that drives `whatsapp-web.js`/puppeteer, commit a lockfile and use `npm ci`: the puppeteer↔Chromium protocol handshake breaks silently (perpetual `/notify` 503) if the JS side drifts from the build-time-pinned distro Chromium.
- A host bind-mount consumed by a non-root container user must be `chown`'d to that uid up front; "works on first run, fresh QR/state on every restart" is the signature of a uid-mismatched volume.
- (Reinforces the PR #249 learning) The `z.string().url()` + explicit `startsWith('https://')` superRefine pattern for bearer-carrying URLs was reused here for `WHATSAPP_NOTIFY_URL` — now applied to two env vars, worth promoting to a named convention.

